PR #18, opened 2017-02-03
SlidePanel

### LINKED ISSUE #2: SlidePanel component
@bitpshr commented on [Wed Jan 04 2017](https://github.com/dojo/widget-core/issues/201)

**Enhancement**

Add a SlidePanel component

Related to #85

### LINKED ISSUE #12: Update components with new widget API
**Enhancement**

Pending the next release of [dojo/widget-core](http://github.com/dojo/widget-core), existing components should be updated to use `render` instead of `getChildrenNodes` and `nodeAttributes`.

### LINKED ISSUE #17: Custom Element Wrapper - Dialog
**Enhancement** 

Dialog is the only widget that landed prior to the custom element support, so need to retrospectively create the custom element wrapper.

bitpshr:
**Type:** feature

**Description:**

This PR introduces a SlidePanel component, updates the existing Dialog component to the new widget API, and adds custom element wrappers for each. It also adds accessibility features to the Dialog based on suggestions from @smhigley. Focus accessibility was omitted until we work through a focus manager.

Initial SlidePanel API:

```ts
align?: string;
open?: boolean;
underlay?: boolean;
width?: number;
onOpen?(): void;
onRequestClose?(): void;
```

Resolves #2, Resolves #12, Resolves #17


src/components/slidePanel/createSlidePanel.ts
```diff
@@ -0,0 +1,175 @@
...
+	onSwipeStart?(event: MouseEvent & TouchEvent): void;
+	onSwipeMove?(event: MouseEvent & TouchEvent): void;
+	onSwipeEnd?(event: MouseEvent & TouchEvent): void;
+	afterCreate?(element: HTMLElement): void;
+	onTransitionEnd?(event: TransitionEvent): void;
+};
+
+export interface SlidePanelFactory extends WidgetFactory<SlidePanel, SlidePanelProperties> { };
+
+let content: HTMLElement;
```
agubler: Do these not have to be stored in a weakmap keyed by the instance? Otherwise having multiple instances will mess with the values.

bitpshr: You're right. Do you have an example of how that would look?

agubler:
Can we add the a custom element wrapper by default for all our components? There are instructions on the readme.

src/components/slidePanel/createSlidePanel.ts
agubler: Something like this I'd imagine.

```ts
interface InternalState {
    content: any;
}

const internalStateMap = new Weakmap<SlidePanel, InternalState>();

// and something like this in the initialise
initialize(instance) {
    internalStateMap.set(instance, initialValues):
}
```

package.json
```diff
@@ -17,12 +17,12 @@
     "test": "grunt test"
   },
   "dependencies": {
-    "@dojo/widget-core": "2.0.0-alpha.20",
     "@dojo/compose": "2.0.0-beta.21",
     "@dojo/core": "2.0.0-alpha.20",
     "@dojo/has": "2.0.0-alpha.7",
     "@dojo/i18n": "2.0.0-alpha.5",
     "@dojo/shim": "2.0.0-beta.8",
+    "@dojo/widget-core": "^2.0.0-alpha.23",
```
agubler: we are trying to work with pinned dependencies at the moment.

src/components/slidePanel/createSlidePanel.ts
```diff
@@ -0,0 +1,205 @@
+import { DNode, Widget, WidgetProperties, WidgetFactory } from '@dojo/widget-core/interfaces';
+import createWidgetBase from '@dojo/widget-core/createWidgetBase';
+import { v } from '@dojo/widget-core/d';
+import WeakMap from '@dojo/shim/WeakMap';
+
+import * as css from './styles/slidePanel.css';
+import * as animations from '../../styles/animations.css';
+import themeable, { ThemeableMixin } from '@dojo/widget-core/mixins/themeable';
+
+export interface SlidePanelProperties extends WidgetProperties {
```
agubler: let's add some jsdoc for the properties (and other exported APIs)

src/components/slidePanel/createSlidePanel.ts
```diff
@@ -0,0 +1,205 @@
...
+				return;
+			}
+
+			const {
+				width = 256,
+				align = 'left'
+			} = this.properties;
+
+			// Current pointer position
+			let currentX = event.type === 'touchmove' ? event.changedTouches[0].screenX : event.pageX;
```
agubler: presume these can both be `const`?

src/components/slidePanel/createSlidePanel.ts
```diff
@@ -0,0 +1,205 @@
...
+
+			// If the panel was swiped far enough to close
+			if (delta > width / 2) {
+				// Cache the transform to apply on next render
+				state.transform = 100 * delta / width;
+				internalStateMap.set(this, state);
+				onRequestClose && onRequestClose();
+			}
+			// If the underlay was clicked
+			else if (delta > -5 && delta < 5 && (<HTMLElement> event.target).classList.contains(css.underlay)) {
```
agubler: I think we might have a type somewhere for this (I'll see if I can find it)

src/components/slidePanel/createSlidePanel.ts
```diff
@@ -0,0 +1,205 @@
...
+				open = false,
+				align = 'left',
+				underlay = false,
+				onOpen
+			} = this.properties;
+
+			const classes: {[key: string]: any} = {};
+			const styles: {[key: string]: any} = {};
+
+			classes[css.content] = true;
```
agubler: I think that we should be using the `this.class(` function as it effectively manages the creation of any required `false` classes, so for this example you could just have an array of classNames that you want to be set and then spread them `this.classes(...classesToSet).get()`?

```ts
const classes = [css.content];

if (open && !state.wasOpen) {
    classes.push(css.slideIn);
}
else if (!open && state.wasOpen) {
    classes.push(css.slideOut);
    if (state.transform !== 0) {
        styles['transform'] = `translateX(${ align === 'left' ? '-' : '' }${ state.transform }%)`;
    }
}

const content = v('div', { 
    key: 'content', 
    classes:  this.classes(...classes).get(), 
    styles, 
    afterCreate: this.afterCreate 
}, this.children);
```

bitpshr:
@agubler: Ready for another review when you get a chance. Want to make sure my JSDoc looks ok, I was a little confused what to document for returned factory functions.

src/components/dialog/createDialog.ts
```diff
@@ -1,97 +1,128 @@
...
-export type Dialog = Widget<DialogProperties> & Themeable<typeof baseTheme> & {
+/**
+ * @type Dialog
+ *
+ * A Dialog component
+ *
+ * @property	{Function?}		onCloseClick		Event handler for when the close button is clicked
+ * @property	{Function?}		onUnderlayClick		Event handler for when a click occurs outside the dialog
+ */
+export type Dialog = Widget<DialogProperties> & ThemeableMixin & {
```
agubler: these are mandatory really? as you implement them as the author?

src/components/slidePanel/createSlidePanel.ts
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * A SlidePanel component
+ *
+ * @property	{Function?}		onSwipeStart		Event handler for when a touch or mouse press starts on the panel or underlay
+ * @property	{Function?}		onSwipeMove			Event handler for when the panel or underlay is dragged or swiped
+ * @property	{Function?}		onSwipeEnd			Event handler for when a touch or mouse press ends on the panel or underlay
+ * @property	{Function?}		afterCreate			Called after the panel is rendered as DOM
+ * @property	{Function?}		onTransitionEnd		Event handler for when the panel finishes any animation
+ */
+export type SlidePanel = Widget<SlidePanelProperties> & ThemeableMixin & {
```
agubler: pretty much same comment, they are all implemented below so can remove the `?`?

src/components/slidePanel/createSlidePanel.ts
```diff
@@ -0,0 +1,205 @@
...
+	align?: string;
+	open?: boolean;
+	underlay?: boolean;
+	width?: number;
+	onOpen?(): void;
+	onRequestClose?(): void;
+};
+
+export type SlidePanel = Widget<SlidePanelProperties> & ThemeableMixin & {
+	onSwipeStart?(event: MouseEvent & TouchEvent): void;
```
bitpshr: Fixed.

src/components/slidePanel/createSlidePanel.ts
bitpshr: Fixed.

src/components/dialog/createDialog.ts
bitpshr: Fixed.

bitpshr:
Ready for another look.

src/components/dialog/createDialogElement.ts
```diff
@@ -0,0 +1,47 @@
...
+		attributes: [
+			{
+				attributeName: 'enterAnimation'
+			},
+			{
+				attributeName: 'exitAnimation'
+			},
+			{
+				attributeName: 'modal',
+				value: value => Boolean(value)
```
rorticus: These Boolean conversions aren't working quite right. `Boolean('false')` and `Boolean('0')` return `true`.

bitpshr:
Party on.

Issue #14, opened 2017-02-01
AccordionPane


bitpshr:
**Enhancement**

Add an AccordionPanel component

rishson:
So we need to list out the initial features.

Things to consider:
- Can multiple sections be open at the same time, or only 1 section at a time?
- Does the accordion's height stay constant, irrespective of the open section(s) content, or does it resize?
- Can you disable sections?

For the initial implementation, my suggestion would be:

1 section open at once
height stays constant
no disabled sections.

rishson:
Also, I'd like to standardise on naming things that may have been called 'panel' in the past.
I've been using semantic ui for too long, but i've got used to the brevity of their widget naming, so we could have:

accordion
tab
sideBar
split

Thoughts?

msssk:
I am also a fan of names being as long and descriptive as necessary, but no more. I think the "panel" suffix is often unnecessary.

For now, we should try to agree on the accordion name, and my vote goes to "accordion".

The overall naming scheme might be best tracked in a separate issue, but some quick thoughts:

* tab: we probably want both `tabs` (or `tabContainer`?) and `tab`, yes?
* split: this is a little lacking in descriptiveness for me... the "panel" suffix might actually be warranted

rishson:
`tab` is a contentious one. I've always viewed `tab` as both the `tabStrip` and the `tabContents` (for want of better names. 
`tabPane` and `splitPane`? 
(I just dislike 'panel' as its too retro).

rishson:
Which would give us:

- tabPane
- splitPane
- slidePane
- accordion
?

bitpshr:
While I don't really care what suffix we use (e.g. "Pane" vs. "Panel" vs. anything else), I do think that every container component should use the same suffix. I think having one component that doesn't use it, like "Accordion", is confusing. This suffix sort of serves as an indicator that a component is a container-style component.

rishson:
@bitpshr I've started to wonder if we need to consider all the widgets where the primary function is layout, such as : 

- menu
- header
- footer

To me, it seems clunky to have `pane` suffixes to these widgets. If the list was something like:

- tabPane
- splitPane
- slidePane
- accordion
- menu
- header
- footer

then I think its ok to drop the suffix at times. [shrug_emoji]

bitpshr:
@rishson sounds good to me.

msssk:
Some questions on what we want from an accordion pane:

[...]

PR #26, opened 2017-02-08
TitlePane

msssk:
**Type:** bug / feature

**Description:**

This PR introduces a TitlePane component very similar to [Dijit's TitlePane](https://dojotoolkit.org/reference-guide/1.10/dijit/TitlePane.html).

A title pane should be a suitable child for an [Accordion widget](https://github.com/dojo/widgets/issues/14).


src/components/titlePane/createTitlePane.ts
```diff
@@ -0,0 +1,88 @@
...
+	enterAnimation?: string;
+	exitAnimation?: string;
+	title: string;
+
+	onRequestCollapse?(): void;
+	onRequestExpand?(): void;
+};
+
+export type TitlePane = Widget<TitlePaneProperties> & ThemeableMixin & {
+	onClickTitle?(): void;
```
bitpshr: Event handlers shouldn't be optional since you're the widget author and you know they'll always be present.

src/components/titlePane/createTitlePane.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import * as css from './styles/titlePane.css';
+import * as animations from '../../styles/animations.css';
+
+export interface TitlePaneProperties extends WidgetProperties {
+	collapsed?: boolean;
+	collapsible?: boolean;
+	enterAnimation?: string;
+	exitAnimation?: string;
+	title: string;
```
bitpshr: `title` should probably be an optional property like it is on the Dialog. It looks like you're already correctly defaulting it to an empty string in `getChildrenNodes`.

src/components/titlePane/createTitlePane.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export interface TitlePaneFactory extends WidgetFactory<TitlePane, TitlePaneProperties> { };
+
+const createTitlePane: TitlePaneFactory = createWidgetBase
+.mixin(themeable)
+.mixin({
+	mixin: {
+		baseClasses: css,
+
+		onClickTitle: function (this: TitlePane) {
```
bitpshr: Should this be named `onTitleClick` instead? Just preference, but it would be good to stay consistent with other components. Also, all of these functions can be written without the `function` keyword:

```ts
...

onTitleClick(this: TitlePane) {
    // ...
}

...
```

src/components/titlePane/createTitlePane.ts
```diff
@@ -0,0 +1,88 @@
...
+			const {
+				properties: {
+					collapsed = false
+				}
+			} = this;
```
bitpshr: The following may be cleaner, negating the need for `this.properties` four other times in this function:

```ts
const {
    collapsed = false,
    onRequestExpand,
    onRequestCollapse
} = this.properties;
```

src/components/titlePane/createTitlePane.ts
```diff
@@ -0,0 +1,88 @@
...
+
+			if (collapsed) {
+				this.properties.onRequestExpand && this.properties.onRequestExpand();
+			}
+			else {
+				this.properties.onRequestCollapse && this.properties.onRequestCollapse();
+			}
+		},
+
+		getChildrenNodes: function (this: TitlePane): DNode[] {
```
bitpshr: If you rebase against master, widgets will be using the most recent release of widget-core. Now, `getChildrenNodes` is replaced by a `render` function. This function should return a single top-level `DNode`, the children of which should be what you're currently returning here. See the Dialog or SlidePanel for reference.

src/components/titlePane/createTitlePane.ts
```diff
@@ -0,0 +1,88 @@
...
+				titleProperties = {
+					classes: this.classes(css.title).get()
+				};
+			}
+
+			const children: DNode[] = [
+				v('div', titleProperties, [ title ])
+			];
+
+			if (!collapsed) {
```
bitpshr: This should probably be `if (collapsible && !collapsed)` so a non-collapsible TitlePane can't be collapsed programmatically.

src/components/titlePane/createTitlePane.ts
```diff
@@ -0,0 +1,88 @@
...
+					exitAnimation
+				}, this.children));
+			}
+
+			return children;
+		}
+	}
+});
+
+export default createTitlePane;
```
bitpshr: I don't care what suffix we use ("Panel" vs. "Pane" vs. "BestDamnLayoutWidget"), but we should stay consistent with other components.

src/components/titlePane/styles/titlePane.css
```diff
@@ -0,0 +1,26 @@
...
+
+.main {
+	overflow: hidden;
+}
+
+.title {
+	background-color: var(--title-bar-bg);
+	border-color: var(--title-bar-bg);
+	border-style: solid;
+	border-width: 1px;
```
bitpshr: This should be a CSS variable, e.g. `var(--title-pane-border-width)`. Don't worry about getting specific.

src/components/titlePane/styles/titlePane.css
```diff
@@ -0,0 +1,26 @@
...
+	color: var(--title-bar-color);
+	font-size: var(--title-font-size);
+	padding: 5px;
```
bitpshr: This should be a variable.

src/components/titlePane/styles/titlePane.css
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.collapsible {
+	cursor: pointer;
+}
+
+.content {
+	border-color: var(--title-bar-bg);
+	border-style: solid;
+	border-width: 1px;
```
bitpshr: This should be the same variable used above for the `.title` border width.

src/components/titlePane/createTitlePane.ts
msssk: Should it? I feel like if you haven't specified title, you've chosen the wrong widget.

bitpshr: Fair. In that case, there's no need to default it when you destructure in `getChildrenNodes`.

src/components/titlePane/createTitlePane.ts
msssk: Thanks, fixed!

src/components/titlePane/createTitlePane.ts
msssk: I gave this some thought as I was writing it and I think "on&lt;Action>&lt;Qualifier>" makes the most sense. I feel like I also found existing usage of this form, but I can't recall where.

src/components/titlePane/createTitlePane.ts
msssk: I'm not a fan of local variables without good justification:

* local default: `collapsed` gets a local default of `false`, the others are used as-is
* frequent repeated access: name lookup time is reduced with a local variable
* excessive verbosity: more subjective, but for me `this.properties.someThing` does not cross the threshold of being too verbose

src/components/titlePane/createTitlePane.ts
msssk: Thanks, good catch!

msssk:
Something that still needs to be figured out is how to animate collapse/expand.

`translateY` looks very nice, but does not animate the position of the element below - that element just snaps up and down.

`max-height` is what I am using, but it doesn't work well if you don't use a good value. Right now it's 300px, which of course will be bad for content taller than that. But if I set it to say, 2000px, then for 100px tall content the first 95% of the animation has no effect and all the visible animation happens in 5% of the time.

src/components/titlePane/createTitlePane.ts
msssk: Thanks - I tried just now but `v()` doesn't accept `undefined` as a child, so we either have to cast to string when passing it as the child to `v('div', ..., [ String(title) ])` or default it... I'm a little preferring "default it".

src/components/titlePane/createTitlePane.ts
bitpshr: Whatever you prefer, just try to make sure the existing components are updated as well to stay consistent.

src/components/titlePane/createTitlePane.ts
bitpshr: I'd argue that having to use `this.properties.someThing` four additional times, well as a lengthier destructuring, is excessively verbose...

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,90 @@
...
+export interface TitlePanelFactory extends WidgetFactory<TitlePanel, TitlePanelProperties> { };
+
+const createTitlePanel: TitlePanelFactory = createWidgetBase
+	.mixin(themeable)
+	.mixin({
+		mixin: {
+			baseClasses: css,
+
+			onClickTitle(this: TitlePanel) {
+				const {
```
tomdye: I'd be tempted to drop the nested destructure in favour of `const { collapsed } = this.properties`.
Could also destructure out `onRequestExpand` and `onRequestCollapse`. You could even consider defaulting them both to a `noop` such that they can both just be called without the checks.

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export interface TitlePanelFactory extends WidgetFactory<TitlePanel, TitlePanelProperties> { };
+
+const createTitlePanel: TitlePanelFactory = createWidgetBase
+	.mixin(themeable)
+	.mixin({
+		mixin: {
+			baseClasses: css,
+
+			onClickTitle(this: TitlePanel) {
```
tomdye: I think the naming convention for this should be `onTitleClick`, but I could be wrong.

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,90 @@
...
+import createWidgetBase from '@dojo/widget-core/createWidgetBase';
+import { v } from '@dojo/widget-core/d';
+import { DNode, Widget, WidgetFactory, WidgetProperties } from '@dojo/widget-core/interfaces';
+import themeable, { ThemeableMixin } from '@dojo/widget-core/mixins/themeable';
+
+import * as css from './styles/titlePanel.css';
+import * as animations from '../../styles/animations.css';
+
+export interface TitlePanelProperties extends WidgetProperties {
+	collapsed?: boolean;
```
tomdye: I'd recommend swapping `collapsed` for `open` as we've used `open` elsewhere and it seems that you're needing to negate `collapsed` for most use cases.

src/components/titlePanel/styles/titlePanel.css
```diff
@@ -0,0 +1,31 @@
+@import '../../../styles/variables.css';
+
+:root {
+	--title-panel-border-width: 1px;
```
tomdye: I think these are likely standard component variables / styling and thus should go to `variables.css` (with different names)

src/components/titlePanel/createTitlePanel.ts
tomdye: I'd recommend swapping `collapsed` for `open` as we've used `open` elsewhere and it seems that you're needing to negate `collapsed` for most use cases.

src/styles/animations.css
```diff
@@ -29,3 +29,33 @@
...
+@keyframes slideInDown {
+	from {
+		max-height: 0;
+	}
+	to {
+		max-height: 300px;
+	}
+}
+
+.slideInDown {
```
tomdye: I think we need to change this name, in `animate-js` this is the class / naming for sliding something *in* and *down* from off the page. I think this should be `slideDown` or `expandDown` We likely need to be cleverer than this though as we want to make sure the text doesn't move about as we expand and it may be better to fade that in at the same time? Worth playing with a little. Also, where did `300px` come from for a max height? Seems restrictive, what are other frameworks / solutions doing?

src/components/titlePanel/createTitlePanel.ts
tomdye: I'd recommend swapping `collapsed` for `open` as we've used `open` elsewhere and it seems that you're needing to negate `collapsed` for most use cases.

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,90 @@
...
+					titleProperties = {
+						classes: this.classes(css.title).get()
+					};
+				}
+
+				const children: DNode[] = [
+					v('div', titleProperties, [ title ])
+				];
+
+				if (collapsible && !collapsed) {
```
tomdye: I don't think you need to have `collapsible` here, would be better if `!collapsed` was replaced with `open` and defaulted to true.

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,90 @@
...
+					children.push(v('div', {
+						classes: this.classes(css.content).get(),
+						enterAnimation,
+						exitAnimation
+					}, this.children));
+				}
+
+				return v('div', {
```
tomdye: No need for the `children` array you created, you can do this:

``` typescript
return v('div', {
  classes: this.classes(css.main).get()
}, [
  v('div', titleProperties, [ title ]),
  !collapsed ? v('div', {
    classes: this.classes(css.content).get(),
    enterAnimation,
    exitAnimation
  }, this.children): null
]);
```

src/components/titlePanel/createTitlePanel.ts
bitpshr: `collapsible` is necessary so a non-collapsible TitlePane can't be closed programmatically.

src/components/titlePanel/createTitlePanel.ts
bitpshr: +1, and maybe `closeable` instead of `collapsible`.

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,90 @@
+import { VNodeProperties } from '@dojo/interfaces/vdom';
```
rishson: Before we land this, can we also add `a11y` support please?

bitpshr: @smhigley will have good suggestions for what's needed. You can omit focus / tabindex stuff for now until we get a focus manager.

bitpshr:
@msssk, looks like we decided to go with Pane, not Panel. Just a heads up. master will be updated shortly.

src/components/titlePanel/createTitlePanel.ts
msssk: I don't like ternary usage except for very simple and short (in character count) cases. Perhaps it's a matter of preference between a functional/declarative approach versus procedural. I'd rather read more lines of code that are each simple, grokkable at a glance, and sequentially and clearly build up to a solution than fewer lines of code that have to be stared at for quite some time to decipher what all is being accomplished.

src/components/titlePanel/createTitlePanel.ts
smhigley: I'll add specific comments, but I'm working off the a11y specs in https://www.w3.org/TR/wai-aria-practices-1.1/#accordion

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,90 @@
...
+					title = ''
+				} = this.properties;
+
+				let titleProperties: VNodeProperties;
+
+				if (collapsible) {
+					titleProperties = {
+						classes: this.classes(css.title, css.collapsible).get(),
+						onclick: this.onClickTitle
+					};
```
smhigley: The title should have a `role="heading"` and ideally an `aria-level="#"`. The heading level may need to be added to properties, if we use it. The title div should also wrap a `button` element with the following properties:

- `aria-controls="id of content div"`
- `aria-expanded: boolean` set to `collapsed`
- `aria-disabled: boolean` set to `!collapsible`
- `id` attribute for the content div to reference

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,90 @@
...
+				const children: DNode[] = [
+					v('div', titleProperties, [ title ])
+				];
+
+				if (collapsible && !collapsed) {
+					children.push(v('div', {
+						classes: this.classes(css.content).get(),
+						enterAnimation,
+						exitAnimation
+					}, this.children));
```
smhigley: The content div needs an `id` for the title to reference, as well as `aria-labelledby="id of title button"`

src/components/titlePanel/createTitlePanel.ts
msssk: We should figure out and document a convention - I think `on<Action><Qualifier>` is preferable myself.

src/components/titlePanel/createTitlePanel.ts
msssk: The change from `collapsible` to `closeable` destroyed this logic, which is fine with me - as long as we enforce UI consistency, I feel OK with allowing programmatic actions that are disabled in the UI.

src/styles/animations.css
msssk: Seems like they're all doing `$(node).collapsible`, but I don't think we want to go that way? 😉 

I know this solution is [not good](https://github.com/dojo/widgets/pull/26#issuecomment-278499474) and would like some more discussion on what we want to do. Are we OK with programmatic animations? Not sure there's a good way to do this in CSS alone.

msssk: These were originally sliding animations, which personally I think is a nice effect, but collapse/expand seems more common for accordions, and in any case, what we currently have is collapse/expand. Renamed to `expandDown` and `collapseUp`.

src/components/titlePanel/createTitlePanel.ts
bitpshr: The logic should still work:

```ts
if (!closeable || (closeable && open)) {
    ...
}
```

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -68,13 +79,30 @@ const createTitlePanel: TitlePanelFactory = createWidgetBase
...
 
+				titleProperties.role = 'heading';
+
+				if (ariaHeadingLevel) {
+					titleProperties['aria-level'] = String(ariaHeadingLevel);
+				}
+
+				if (open) {
+					// FIXME - id of content node
+					titleButtonProperties['aria-controls'] = 'content';
```
smhigley: The button should actually have `aria-controls="id string"` set all the time, both open and closed. (Incidentally, I believe Matt G recommended `uuid` to generate unique id's, if it helps)

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -68,13 +79,30 @@ const createTitlePanel: TitlePanelFactory = createWidgetBase
...
+				}
+
 				const children: DNode[] = [
-					v('div', titleProperties, [ title ])
+					v('div', titleProperties, [
+						v('div', titleButtonProperties, [ title ])
```
smhigley: Nitpicking, sorry! I personally think a `button` element is preferable to a `div`, but if you prefer the div, it'll also need `tabindex="0"` set (a focus manager wouldn't set initial `tabindex` values). You'd also need to make sure keyboard events (enter and space) trigger the content opening -- I'm not completely sure they work by default on a fake button.

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,118 @@
...
+					title = ''
+				} = this.properties;
+
+				// VNodeProperties, but its typing is too restrictive
+				let titleProperties: any;
+				// VNodeProperties, but its typing is too restrictive
+				let titleButtonProperties: any = {
+					'aria-disabled': String(!closeable),
+					'aria-expanded': String(open),
+					// FIXME - set unique id
```
bitpshr: You can use `@dojo/core/uuid` for this, see the Dialog for an example.

src/components/titlePanel/createTitlePanel.ts
bitpshr: +1

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,118 @@
...
+					titleButtonProperties['aria-controls'] = 'content';
+				}
+
+				const children: DNode[] = [
+					v('div', titleProperties, [
+						v('div', titleButtonProperties, [ title ])
+					])
+				];
+
+				if (open) {
```
bitpshr: Thinking about it more, a non-closeable Dialog can be closed programmatically, so this is fine.

src/components/titlePanel/createTitlePanel.ts
bitpshr: Consensus in the Dojo 2 room was for `on<Qualifier><Action>`, so `onTitleClick`.

src/components/titlePanel/createTitlePanel.ts
tomdye: That being so, the example given above is consistent with how widgets and examples of `v` are shown. The `null` is filtered out by `v`.

tomdye:
Can you please drop the camelCase naming on the folder name. ie. `tabPanel` -> `tabpanel`.

src/components/titlePanel/createTitlePanel.ts
msssk: Allocating extra function objects just to avoid null checks doesn't strike me as a good approach.

src/components/titlePanel/createTitlePanel.ts
msssk: So how about we improve widget and `v` examples?

src/components/titlePanel/createTitlePanel.ts
bitpshr: Mangala, you don't have to use noop defaults if you don't want to (I didn't in SlidePane), but I'd still destructure fully to avoid the four extra uses of `this.properties`.

src/components/titlePanel/createTitlePanel.ts
msssk: This seems very awkward to me since with vdom, in the closed state there is no content pane existing in the DOM. So we set `aria-controls="<id of non-existent element>"`?

src/components/titlePanel/createTitlePanel.ts
msssk: I left it as a div since the w3 example doesn't use a button, and I wanted to avoid having to override button styles. But now I realize their example uses `<a>`, which gets much of the behavior that button has. I'd rather override button styles than reimplement button behavior, so I'll use `<button>` here.

src/components/titlePanel/createTitlePanel.ts
```diff
@@ -0,0 +1,118 @@
...
+				};
+
+				if (closeable) {
+					titleProperties = {
+						classes: this.classes(css.title, css.closeable).get(),
+						onclick: this.onClickTitle
+					};
+				}
+				else {
+					titleProperties = {
```
matt-gadd: does this not break the second rule of maquette? (and elsewhere for these properties?) http://maquettejs.org/docs/rules.html

src/components/titlePanel/createTitlePanel.ts
msssk: I prefer the clarity of knowing precisely where the information is coming from at the time it's being used, and don't find this case to be excessively verbose, but I can change it.

`this.properties.onRequestClose` - at a glance I know exactly where `onRequestClose` came from
`onRequestClose` - at a glance I don't know where this came from, I have to look at other code

src/components/titlePanel/createTitlePanel.ts
msssk: Thanks!

matt-gadd: this is a good reason in itself not to procedurally build up properties btw 😄

src/examples/titlePanel/index.html
```diff
@@ -0,0 +1,39 @@
...
+		.main-app > div {
+			margin-bottom: 15px;
+		}
+	</style>
+</head>
+
+<body>
+	<script src="../../../../node_modules/@dojo/loader/loader.min.js"></script>
+	<!-- TODO: should come from project dependency -->
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/maquette/2.4.2/css-transitions.min.js"></script>
```
matt-gadd: I've seen this elsewhere in the codebase but i'll ask anyway, is there any reason this cannot be loaded from `node_modules/maquette/dist/css-transitions.min.js`?

src/components/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,81 @@
...
+	exitAnimation?: string;
+	open?: boolean;
+	title: string;
+
+	onRequestClose?(): void;
+	onRequestOpen?(): void;
+};
+
+@theme(css)
+export default class TitlePane extends ThemeableMixin(WidgetBase)<TitlePaneProperties> {
```
agubler: We need to set the result of `ThemeableMixin(WidgetBase)` to a `const` and use the result to extend from.

```ts
const TitlePaneBase = ThemeableMixin(WidgetBase);

export default class TitlePane extends TitlePaneBase<TitlePaneProperties> {
```

src/components/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,81 @@
...
+	onClickTitle(event: MouseEvent & TouchEvent) {
```
agubler: Do we need the `event` param?

msssk:
Test coverage is up to 100%.

Outstanding issues to resolve:

* What do we do with content when the TitlePane is collapsed? @smhigley?
    * Current behavior: don't render the content, and don't set the `aria-controls` attribute on the title button
    * If we are going to set `aria-controls` then we should render the content. How do we hide it in an a11y-friendly manner?
* How do we handle animation?
    * There are only a few CSS properties that animate smoothly, and the currently used `max-height` is not one of them. Also `max-height` is a poor solution because we don't know what height the content will be.
    * An animation function would give us precise control and handle arbitrary heights, but we don't have support for that

smhigley:
> What do we do with content when the TitlePane is collapsed?

Either the content or an element containing the content needs to be always present in the DOM, and its ID should be set in `aria-controls`. It (or its children) can be hidden with `display: none` or `visibility: hidden`, or the content can be added/removed from the DOM, so long as the containing element is present.

> How do we handle animation?

I agree, I've never liked the max-height trick, but all the other options involve calculating height. What I've done in the past is clone the content node and position it offscreen with the correct width and box styles, and use it to calculate height.

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,92 @@
...
+		return v('div', {
+			classes: this.classes(css.main).get()
+		}, [
+			v('div', {
+				'aria-level': ariaHeadingLevel ? String(ariaHeadingLevel) : '',
+				classes: this.classes(css.title, closeable ? css.closeable : null).get(),
+				onclick: closeable ? this.onClickTitle : undefined,
+				role: 'heading'
+			}, [
+				v('div', {
```
smhigley: All the changes look good! One very, very last thing and then I'm satisfied a11y-wise. The `button` just needs to either be an element that takes focus by default (`<button>` or `<a>`), or have `tabindex` set to 0.

msssk:
Thanks @smhigley that makes sense - DOM structure is now consistent, it's only the content nodes that are omitted when collapsed.

Now the remaining issues are:

* expand/collapse animations
* shared CSS: right now there is a `hidden` class in `TitlePane.css` - do we have a structure for sharing common classes like this?

smhigley:
@msssk the changes look great! I think the `hidden` class would be a good candidate for a util css file along with `visuallyHidden`.

Maybe for animations, `max-height` is as good as we'll get for now, if you think it's better to not mess around with DOM height calculations. Here's an example where it's done well: https://codepen.io/chriswrightdesign/pen/cmanI

src/titlepane/TitlePane.ts
msssk: Thanks, done! (`tabIndex` is 0, but if `closeable` is false then `tabIndex` is -1)

msssk:
@smhigley The `max-height` solution is terrible (besides being visually janky, it doesn't support arbitrary content height) and should never see the light of day, so we definitely need a better solution.

I'm happy to do height calculations, but afaik we still don't have support for [programmatic animations](http://maquettejs.org/docs/animations.html#programmatic-animations) in `dojo/widget-core`.

tomdye:
@msssk I don't think programatic animations will be required here as you should be able to set the `style: height` attribute on the element and allow it to transition via css.

src/common/styles/animations.css
```diff
@@ -29,3 +29,33 @@
 	animation-fill-mode: both;
 	animation-name: fadeOut;
 }
+
+@keyframes expandDown {
```
bitpshr: These can all go away now.

src/titlepane/example/index.html
```diff
@@ -0,0 +1,38 @@
+<!doctype html>
```
bitpshr: If you rebase, this page can be deleted. You can hit your test page using the URL `http://your_local_server/_build/src/common/example/?module=TitlePane`.

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,99 @@
...
+		const contentId = uuid();
+		const titleId = uuid();
+
+		return v('div', {
+			classes: this.classes(css.main)
+		}, [
+			v('div', {
+				'aria-level': ariaHeadingLevel ? String(ariaHeadingLevel) : '',
+				classes: this.classes(css.title, closeable ? css.closeable : null),
+				onclick: closeable ? this._onTitleClick : undefined,
```
smhigley: This also needs to open/close with the keyboard (specifically space and enter)

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * @type TitlePaneProperties
+ *
+ * Properties that can be set on a TitlePane component
+ *
+ * @property closeable          If false the pane will not collapse in response to clicking the title
+ * @property headingLevel       'aria-heading-level' for the title's DOM node
+ * @property open               If true the pane is opened and content is visible
+ * @property title              Title to display above the content
+ * @property onRequestClos      Called when the title of an open pane is clicked
```
bitpshr: `onRequestClos` -> `onRequestClose`

src/titlepane/createTitlePaneElement.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Configures a TitlePane web component
+ */
+export default function createTitlePaneElement(): CustomElementDescriptor {
+	return {
+		tagName: 'dojo-title-pane',
+		widgetConstructor: TitlePane,
+		attributes: [
+			{
+				attributeName: 'headingLevel',
```
bitpshr: Attribute names should be lowercase (there's a PR right now to update other component descriptors), so this should be:

```js
attributeName: 'headinglevel',
propertyName: 'headingLevel'
```

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,118 @@
...
+
+import * as css from './styles/titlePane.m.css';
+
+/**
+ * @type TitlePaneProperties
+ *
+ * Properties that can be set on a TitlePane component
+ *
+ * @property closeable          If false the pane will not collapse in response to clicking the title
+ * @property headingLevel       'aria-heading-level' for the title's DOM node
```
smhigley: It's a nitpick, but if you quote the aria attribute, it should be 'aria-level'.

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,118 @@
...
+export default class TitlePane extends TitlePaneBase<TitlePaneProperties> {
+	private _afterRender(element: HTMLElement) {
+		// Conditionally adjust top margin. Done manually instead of through Maquette
+		// so the underlying DOM is accessible, as we need to know the content height.
+		// Put in a timeout to push this operation to the next tick, otherwise
+		// element.offsetHeight below can be incorrect (e.g. before styling is applied)
+		requestAnimationFrame(() => {
+			const { open = true } = this.properties;
+			const height = element.offsetHeight;
+			element.style.marginTop = open ? '0px' : `-${ height }px`;
```
smhigley: Is there a reason to animate `margin-top` instead of `height`? I had thought `height` would be more performant, but I haven't tested those two directly against each other.

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,118 @@
...
+		});
+	}
+
+	private _onTitleClick() {
+		this._toggle();
+	}
+
+	private _onTitleKeyUp(event: KeyboardEvent) {
+		if (event.keyCode === /* enter */ 13 ||
+			event.keyCode === /* space */ 32) {
```
smhigley: Should import `Keys` from `common/util` and use those now

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,118 @@
...
+	render(): DNode {
+		const {
+			headingLevel,
+			closeable = true,
+			open = true,
+			title
+		} = this.properties;
+
+		const contentId = uuid();
+		const titleId = uuid();
```
smhigley: I've been storing these as private properties so they don't change and need updating on every render.

src/titlepane/example/index.ts
```diff
@@ -0,0 +1,94 @@
...
+				})
+			]),
+			v('div', {
+				styles: {
+					height: '15px'
+				}
+			}),
+			w(TitlePane, {
+				key: 'titlePanel3',
+				open: <boolean> (this.state['t3open'] === undefined ? false : this.state['t3open']),
```
smhigley: Since this is an example showing different initial states, would it make sense to use destructuring? E.g.
```javascript
const {
   t2open = true,
   t3open = false
} = this.state;
```

src/titlepane/styles/titlePane.css
```diff
@@ -0,0 +1,34 @@
...
+	overflow: hidden;
+	box-sizing: border-box;
+}
+
+.main * {
+	box-sizing: border-box;
+}
+
+.title {
+	background-color: var(--title-bar-bg);
```
smhigley: Probably need to change some of these now that @tomdye has added an initial theme. Also, why does this file exist alongside `titlePane.m.css`?

agubler:
@msssk @mwistrand @bitpshr Needs to be rebased to for the latest `beta1` release of `widget-core` and `test-extras`.

src/titlepane/TitlePane.ts
mwistrand: Good catch.

src/titlepane/TitlePane.ts
mwistrand: @bitpshr can confirm, but I *think* the animation for `height` is slightly more involved, and also animating `margin-top` gives the impression of the container being pulled out like a drawer, which is nice touch for this particular widget.

tomdye: ^ disregard my comment. We will leave current animations in place until such time as our animations api is created

src/titlepane/tests/unit/TitlePane.ts
```diff
@@ -0,0 +1,208 @@
...
+			classes: titlePane.classes(css.root)
+		}, [
+			v('div', {
+				'aria-level': '',
+				classes: titlePane.classes(css.title, css.closeable),
+				onclick: titlePane.listener,
+				onkeyup: titlePane.listener,
+				role: 'heading'
+			}, [
+				v('div', <any> {
```
tomdye: are these `<any>` casts required?

mwistrand: Yes. `VirtualDomProperties` expects a `string | undefined` for the `id` property, but for the tests we're providing a custom diff function.

smhigley: I got the same comment on Calendar tests, and I think it may be better to individually cast the diff functions instead of the entire properties object.

agubler: I would cast the individual property perhaps, I know @smhigley originally cast the entire properties object but then moved to just the property that a custom diff function was being used for.

agubler: Damn you beat me @smhigley 😄

Issue #5, opened 2017-01-24
Add Form Widgets


dylans:
@smhigley commented on [Fri Dec 09 2016](https://github.com/dojo/widget-core/issues/174)

**Enhancement** 

Related to #2. Update existing textinput and button components to be accessible, and add the following basic form widgets based on #85:

- [x] Email/number/password/etc to existing textInput
- [x] Checkbox
- [x] Radio buttons
- [x] Select box
- [x] Textarea

rishson:
@smhigley we OK to close this now?

smhigley:
@rishson the select widget is still open, it's ending up being a little more trouble than I thought. That's the last one, though.

PR #30, opened 2017-02-10
Button Widget

smhigley:
**Type:** feature

**Description:**
Related to #5
Adds a button widget, with optional icon and styles. Includes tests and example usages.


src/button/createButton.ts
```diff
@@ -0,0 +1,96 @@
+import { DNode, Widget, WidgetProperties, WidgetFactory } from '@dojo/widget-core/interfaces';
+import createWidgetBase from '@dojo/widget-core/createWidgetBase';
+import { v } from '@dojo/widget-core/d';
+import themeable, { ThemeableMixin } from '@dojo/widget-core/mixins/themeable';
+import * as css from './styles/button.css';
+
+export interface ButtonProperties extends WidgetProperties {
```
bitpshr: This should probably use JSDoc similar to the existing components.

src/button/createButton.ts
```diff
@@ -0,0 +1,96 @@
...
+
+	/**
+	 * ID of element with descriptive text
+	 */
+	describedBy?: string;
+
+	/**
+	 * Indicates status of a toggle button
+	 */
+	pressed?: boolean;
```
bitpshr: How do I use this button as a toggle button? Would a parent widget pass in a function for `onClick`, then set `pressed` accordingly?

src/button/example/index.ts
```diff
@@ -0,0 +1,46 @@
...
+					disabled: true,
+					type: 'submit'
+				}),
+				v('p', {
+					innerHTML: 'Icon Button'
+				}),
+				w(createButton, {
+					content: 'Favorite',
+					icon: ''
+				})
```
bitpshr: Maybe add a toggle example?

src/button/styles/button.css
```diff
@@ -0,0 +1,47 @@
+@import '../../styles/variables.css';
+
+.button {
+	background-color: var(--button-bg);
+	border: 1px solid var(--button-bg);
+	box-shadow: inset 0 0 0 1px rgba(255, 255, 255, 0.25);
```
bitpshr: This and most other numerical values below (`border-radius`, `padding`, `transition-duration`, `font-size`, etc.) should be CSS variables.

src/button/styles/button.css
```diff
@@ -0,0 +1,47 @@
...
+}
+
+.button[disabled],
+.button[disabled]:hover {
+	background-color: var(--button-disabled-bg);
+	border-color: var(--button-disabled-bg);
+	cursor: default;
+}
+
+.button[data-dojo-icon]:before {
```
tomdye: can you use a class for this instead please? It feels to me that the icon should be passed as a property on the button widget itself rather than via a data-attr. Also worth noting that these rules around font size, weight etc... should be overridable by the icon-font used which may be font-awesome, material-icons etc...

src/button/styles/button.css
```diff
@@ -0,0 +1,47 @@
...
+	speak: none;
+	font-size: 1.4em;
+	font-style: normal;
+	font-weight: normal;
+	font-variant: normal;
+	text-transform: none;
+	line-height: 1;
+	margin: -0.5em 0.3em -0.5em -0.3em;
+	vertical-align: sub;
+	-webkit-font-smoothing: antialiased;
```
tomdye: drop the vendor prefixes, `autoprefixer` does this for us.

src/styles/variables.css
```diff
@@ -14,4 +15,10 @@
 	--title-bar-color: var(--dojo-white);
 	--underlay-bg: var(--dojo-shade-gray);
 	--slide-panel-width: 256px;
+
+	--button-bg: var(--dojo-blue);
```
tomdye: I think we need to be more generic with these naming conventions. `--button` may be too specific when you consider that we will likely use these variables across multiple widgets.

src/button/createButton.ts
```diff
@@ -0,0 +1,96 @@
...
+	hasPopup?: boolean;
+}
+
+export type Button = Widget<ButtonProperties> & ThemeableMixin & {
+	onClick(event?: MouseEvent): void;
+};
+
+export interface ButtonFactory extends WidgetFactory<Button, ButtonProperties> { }
+
+const createButton: ButtonFactory = createWidgetBase.mixin(themeable).mixin({
```
tomdye: I believe this will need converting to use the new `Classes` approach

src/button/createButton.ts
```diff
@@ -0,0 +1,96 @@
...
+				describedBy,
+				hasPopup
+			} = this.properties;
+			const buttonNodeAttributes: any = {
+				innerHTML: content,
+				classes: this.classes(css.button).get(),
+				onclick: this.onClick
+			};
+
+			type ? buttonNodeAttributes.type = type : false;
```
tomdye: I think that most of this can be done via object destructuring and default value assignment.

ie `const { type = false ) = this.properties;`

smhigley: I was using a weird shorthand for `if (type) buttonNodeAttributes.type = type;`, not defaulting `type` to false. I don't want to add those attributes to the `VNodeProperties` object unless they actually exist in properties. Is there another way you'd prefer?

src/styles/variables.css
smhigley: Is there another name you'd prefer? I was going by the `specific usage name = color name` pattern, but I'm open to other naming patterns. I'm not sure what a more generic variable would be in this case, though. It wouldn't necessarily be the same as a link; what else is there?

src/button/createButton.ts
msssk: Even though I've read Maquette's rules several times I did something similar and @matt-gadd had to remind me of this:
http://maquettejs.org/docs/rules.html

src/button/styles/button.css
smhigley: I'm happy to switch the attribute selector for a class. How would you pass the icon content without using a data-attr, though? Or would you prefer creating a child element just for the icon? It doesn't make a difference a11y-wise, but I've always preferred pseudo-elements for stuff like this.

For the font-weight, variant, etc. I added the same styles I use for every icon font I've ever come across. I have a hard time imagining an icon font that uses `font-weight: bold`, for instance :). I could definitely move the `font-size` to a variable, though.

src/button/createButton.ts
smhigley: Yup! I imagine a parent widget would need to be stateful, have `pressed: this.state['buttonPressed']` in the button properties, and use `this.setState` in an onClick handler. I'll add an example for it, as you suggested below :)

src/button/styles/button.css
tomdye: I was thinking more that whatever icon-font a user utilised, it would come with those kind of font size / weight etc... settings. Personally, I'd have expected a user to create an icon button by passing the icon through as part of the button text? Thus allowing any popular icon font to be used?
 ie. 
```
content: 'my icon button <i class="fa fa-save"></i>'
```

tomdye:
@smhigley I believe this will need to be converted to use `Classes` rather than `compose`

src/button/styles/button.css
smhigley: Unfortunately in this case, the font smoothing used differs between -webkit and -moz

src/button/styles/button.css
smhigley: I'm not entirely sure what you mean by icon fonts coming with font size/weight etc. If you think it best, we could just drop the icon property and let users deal with it themselves, as you suggested. I think Torrey mentioned having an icon button though, so it might be nice to have an explicit property for it. One also wouldn't rule out the other -- a user could always ignore `icon` and add extra html to `content` if they wish.

src/button/Button.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @property disabled			Whether the button is disabled or clickable
+ * @property hasPopup			Whether the button triggers a popup
+ * @property icon					Creates an icon button with the specified icon glyph
+ * @property name					The button's name attribute
+ * @property pressed			Indicates status of a toggle button
+ * @property type					Button type can be "submit", "reset", "button", or "menu"
+ * @property onClick			Called when the button is clicked
+ */
+export interface ButtonProperties extends ThemeableProperties {
+	content?: string;
```
bitpshr: Nitpicking, but would `label` be better?

tomdye: Consistency is the key, I believe we are using `label` elsewhere for this.

smhigley: I've been trying to get rid of label as a keyword for text content in favor of label as a keyword for labels

src/button/Button.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+@theme(css)
+export default class Button extends ThemeableMixin(WidgetBase)<ButtonProperties> {
+	onClick(event: MouseEvent) {
+		this.properties.onClick && this.properties.onClick(event);
+	}
+
+	render() {
+		const {
```
agubler: curiously, do you need them do default explicitly to `null`?

src/button/Button.ts
```diff
@@ -0,0 +1,66 @@
...
+			describedBy = null,
+			disabled = null,
+			hasPopup = null,
+			icon = null,
+			name = null,
+			pressed = null,
+			type = null
+		} = this.properties;
+
+		const buttonNodeAttributes: any = {
```
agubler: could this be `VirtualDomProperties` from `widget-core`

src/button/Button.ts
```diff
@@ -0,0 +1,112 @@
...
+	onKeyDown (event: KeyboardEvent) { this.properties.onKeyDown && this.properties.onKeyDown(event); }
+	onKeyPress (event: KeyboardEvent) { this.properties.onKeyPress && this.properties.onKeyPress(event); }
+	onKeyUp (event: KeyboardEvent) { this.properties.onKeyUp && this.properties.onKeyUp(event); }
+	onMouseDown (event: MouseEvent) { this.properties.onMouseDown && this.properties.onMouseDown(event); }
+	onMouseUp (event: MouseEvent) { this.properties.onMouseUp && this.properties.onMouseUp(event); }
+	onTouchStart (event: TouchEvent) { this.properties.onTouchStart && this.properties.onTouchStart(event); }
+	onTouchEnd (event: TouchEvent) { this.properties.onTouchEnd && this.properties.onTouchEnd(event); }
+	onTouchCancel (event: TouchEvent) { this.properties.onTouchCancel && this.properties.onTouchCancel(event); }
+
+	render() {
```
bitpshr: This needs to return a `DNode` otherwise a production build will fail (TS should complain.)

src/button/example/index.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
```
bitpshr: If you rebase, this page can be deleted. You can hit your test page using the URL `http://your_local_server/_build/src/common/example/?module=Button`.

src/button/example/index.ts
```diff
@@ -0,0 +1,52 @@
...
+			}),
+			v('p', {
+				innerHTML: 'Disabled menu button:'
+			}),
+			w(Button, {
+				key: 'b2',
+				content: 'Open',
+				disabled: true,
+				hasPopup: true,
+				type: <ButtonType> 'menu'
```
bitpshr: Is this cast needed?

smhigley: I get TS errors if I don't :(. Is there a way to avoid it?
`Type 'string' is not assignable to type '"button" | "menu" | "reset" | "submit" | undefined'.`

agubler: It feels like there certainly should be a way to avoid this, don't let me block the PR but I will get back to you.

mwistrand: If `"menu"` is a valid `ButtonType`, then casting shouldn't be necessary. It's very odd that it's not registering it as a `ButtonType`.

smhigley: @mwistrand I have `export type ButtonType = 'submit' | 'reset' | 'button' | 'menu';`, `type?: ButtonType;`, and `type: 'menu'` ¯\_(ツ)_/¯

bitpshr: @smhigley this is working in Dialog: https://github.com/bitpshr/widgets/blob/tab-pane/src/dialog/Dialog.ts#L13

smhigley: @bitpshr try setting `role` in `example/index.ts`. I get an TS error in Dialog too.

src/button/styles/button.css
```diff
@@ -0,0 +1,55 @@
...
+	content: '';
+	display: inline-block;
+	height: 1em;
+	margin: -1em 0.5em 0 0;
+	transform: rotate(45deg);
+	width: 0.5em;
+}
+
+/* TODO: this could be a down arrow icon */
+.button.popup::after {
```
tomdye: Does this work if you drop `.button` from this and other similar chained selectors?

tomdye:
Can you remove the `TODO`s before this is merged please?

Issue #22, opened 2017-02-08
Uniform validation approach for form components


rishson:
We need to consider if we plan to add mandatory and status indicators to our form fields.

By 'mandatory' I mean the usual '*' denotation for 'this form element must have a value'.
By 'status' I mean an indicator in/by the field to show that the current value is valid, e.g. [here](http://jamiewilson.io/corpus/#forms).

If we include these indicators, then some things to bear in mind:
- These indicators should not adversely impact `a11y` support.
- We need to consider how we would implement if we also allow for [addons](https://react-bootstrap.github.io/components.html#forms-input-addons) - see [also](https://react-bootstrap.github.io/components.html#forms-validation) 
- We need to consider a 'wait state' for status (if validation of the current value is async).

rishson:
@smhigley keen to get your thoughts on this.

dylans:
We should probably get this done before beta? @rishson and @smhigley

smhigley:
@dylans it's in the latest PRs, but mostly just using color. I'd like to go back and add icons, now that we've talked about how to add them, but I wonder if those and other more complex styles should go in themes.

bitpshr:
I updated the title of this ticket so we can address a general approach to validation here beyond just styling. We should make sure all form components follow the same approach. My current thinking:

A parent component should handle validation of child form components whenever it wants to by using appropriate hooks such as `onChange`. The parent component can use any logic to determine if a child form component value is valid, and if it's not valid, the parent can set `invalid` to `true` on the respective child. I also think that an optional `invalidMessage` could be useful for per-component error messaging. We can use tooltips once we have a proper component to show inline messaging.

smhigley:
@bitpshr +1 for the `invalidMessage`. I wonder if form widgets should actually just allow the parent to pass in a DNode for the error message to allow tooltips and whatever else.

The parent component is probably going to be a form widget, so it might be good to add specs for that, and how we want consumers to specify things like validation functions and error messages.

bitpshr:
Agreed. For now, I think the main thing to keep consistent is programmatic and styling support for `required` and `invalid` for all form components.

rishson:
Agreed. Just for reference thought, potentially add `hintMessage` for when helper text is too large to be used as placeholder text (often seen in material design directly below a field)?

```
email:  [yourname@yourcompany.com     ]*
        (we promise not to spam you).
```

bitpshr:
[...]

PR #32, opened 2017-02-20
Textinput 05

### LINKED ISSUE #41: Text Input Widget
**Enhancement** 
Related to #5 

Features:
- Includes styles for disabled, required, valid, and invalid states
- accepts callback functions for `onInput`, `onChange`, `onFocus`, and `onBlur`

Future changes:
- Could use better styles for certain states, a "wait" state, and perhaps icons

smhigley:
**Type:** feature

**Description:**

Resolves #41, adds text input widget


src/styles/variables.css
```diff
@@ -10,7 +11,13 @@
...
 
+	--block-spacing: 25px;
+	--line-spacing: 5px;
+
 	--title-bar-bg: var(--dojo-blue);
 	--title-bar-color: var(--dojo-white);
 	--underlay-bg: var(--dojo-shade-gray);
+
+	--focus-highlight: 0 0 4px 1px var(--dojo-light-blue);
+	--input-padding: 0.5em 1em;
```
tomdye: can we switch this to `px` to patch other paddings elsewhere please?

src/textinput/styles/textinput.css
```diff
@@ -0,0 +1,21 @@
+@import '../../styles/variables.css';
+
+.textinput {
+	font: var(--base-font);
+}
+
+.textinput input {
```
tomdye: can you please drop the usage of tag names and give a class name to both the `input` and the `span` then you can drop the `.textinput` nesting as it is not required.

src/styles/variables.css
```diff
@@ -10,7 +11,13 @@
...
 	--short-animation-duration: .2s;
 
+	--block-spacing: 25px;
+	--line-spacing: 5px;
+
 	--title-bar-bg: var(--dojo-blue);
 	--title-bar-color: var(--dojo-white);
 	--underlay-bg: var(--dojo-shade-gray);
+
+	--focus-highlight: 0 0 4px 1px var(--dojo-light-blue);
```
tomdye: this looks like it's being used as a `shadow` and we already have a `subtle-shadow` variable so can we align these please and call this `--input-focus-shadow` ?

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,37 @@
...
+		this.properties.onInput && this.properties.onInput(event);
+	}
+
+	render() {
+		const {
+			type = 'text'
+		} = this.properties;
+
+		return v('input', {
+			classes: this.classes(css.textinput).get(),
```
tomdye: I've been thinking about how we should standardise outer class naming and was thinking along the lines of `root` as it's the root node of the widget. Thus this would become `this.classes(css.root).get()`. What do you think? Would make it clearer and more consistent when using an `overrideClass` to position an element as you would override `root: position-class`.

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,37 @@
...
+	onInput?(event: Event): void;
+}
+
+@theme(css)
+export default class TextInput extends ThemeableMixin(FormLabelMixin(WidgetBase))<TextInputProperties> {
+	onInput(event: Event) {
+		this.properties.onInput && this.properties.onInput(event);
+	}
+
+	render() {
```
tomdye: Can you add an `initialValue` to this as well please.

src/styles/variables.css
rishson: Whilst consistency across our codebase is valid, we need to have the age old `px vs em vs rem` debate - for which I've added: #35

src/textinput/TextInput.ts
bitpshr: I think we can use `value` + `onInput` instead of `initialValue`. I've been playing with this for the ComboBox. In the parent component that's using either a TextInput or a ComboBox, bind the `value` property on the input to some state in the parent component; then, update the `value` in the `onInput` callback:

**Parent component**
```ts
...
w(TextInput, {
	value: this.inputValue,
	onInput: (value: string) => this.inputValue = value
})
...
```

**Input**
```ts
...
render() {
	const {
		type = 'text',
		value = ''
	} = this.properties;

	return v('input', {
		classes: this.classes(css.textinput).get(),
		type: type,
		value: value,
		oninput: this.onInput
	});
}
...
```

smhigley: I was thinking something along the lines of what @bitpshr suggested for form widgets. This is already possible with the widget as it is now -- `value` is passed to the widget in the `FormLabel` mixin. I can add an example to demonstrate, but I've been unsure of how much to demo `FormLabel` uses within individual widgets. Do you guys have any opinion on it? Should I just add examples of all `FormLabel` properties to each widget that extends it, or choose one widget (TextInput, perhaps)?

bitpshr: Agreed. Maybe add a default value for one of the examples to demonstrate the flow of default value -> user changes form widget -> `onInput` called -> parent widget updates state. It's not a super obvious flow, but it's how most React components work as well.

smhigley: @bitpshr do you think I should add a default value example to each widget, so someone looking at a specific widget (e.g. Textarea or Select) would immediately see what to do? It seems repetitive, but if it enhances user experience I'm in favor. I could also add either examples of the other generic form widget properties -- `placeholder`, `readOnly`, `describedBy`, etc. or a text list of all the allowed attributes to each example page.

tomdye: It really should be `initialValue` guys, as I believe @matt-gadd discussed with @bitpshr for his combobox issues previously. This will be the *standard* way that initial values are passed into our widgets across the board.

bitpshr: @tomdye: The discussion I had with @matt-gadd was based on the assumption that there was no easy way to invert control to the parent for form controls and that we'd have to break our unidirectional flow approach, but that's not the case. Using `initialValue` vs. `value` would work the same way, but the former wouldn't allow our form controls to invert control to the parent like all other components. We should be using `value` + `onInput` + updating parent state like all other components currently do.

tomdye: @bitpshr the problem with the approach you are suggesting is that it is assuming that every widget wants to update it's value `onInput`. It feels like you're trying to make the data flow bi-directional such that the parent knows the value of the input at all times which goes against out uni-directional reactive design principles. Even without triggering a render, there will be an overhead involved with firing an `onInput` function on every key stroke. The `value` could be accessed in a number of ways depending on the use case; `onFocus`, `onBlur`, `onSubmit` to name a few.  

Also worth considering that external state can be `async`  which would introduce even more problems as you cannot guarantee the order of the callbacks thus the value could easily be corrupted. Using the `initialValue` approach keeps the concerns separate and leaves widget authors / app creators able to get the `value` via whichever callback / event they desire. The uncontrolled approach keeps the single source of truth in the dom.

bitpshr: @tomdye In my example, `onInput` doesn't have to be used. `onChange` could be used for less intensive re-rendering. For what it's worth, [form controls in React](https://facebook.github.io/react/docs/forms.html#controlled-components) are controlled components and work how I described above. This behavior is inherently unidirectional, not bidirectional, and doesn't break any of our design principles. The single source of truth in a virtual DOM environment should be the parent component, not the DOM.

tomdye: Thanks for the great examples @bitpshr. Having read all the facts and doing some more research I agree with your approach. I guess we would suggesting using `onchange` / `oninput` depending on the use case. I would imagine that this will be crucial for how we implement forms as the form inputs will need to be controlled as such.

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Properties that can be set on a TextInput component
+ *
+ * @property type			Input type, e.g. text, email, tel, etc.
+ * @property onInput	Called when the node's 'input' event is fired
+ * @property onChange	Called when the node's 'change' event is fired
+ * @property onFocus	Called when the input is focused
+ * @property onBlur		Called when the input loses focus
+ */
+export interface TextInputProperties extends ThemeableProperties, FormLabelMixinProperties {
+	type?: string;
```
tomdye: `type` should be strictly typed. Will also aid with intelisense

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,64 @@
...
+	onInput?(event: Event): void;
```
tomdye: Is this an extensive list of all of the events that can be attached? dojo1 textbox dijit has `keyPress`, `keyDown` etc..

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,64 @@
...
+	render() {
+		const {
+			type = 'text',
+			invalid
+		} = this.properties;
+
+		const classes = [
+			css.root,
+			typeof invalid === 'boolean' ? invalid ? css.invalid : css.valid : null
+		];
```
tomdye: what about required?

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,64 @@
...
+
+	render() {
+		const {
+			type = 'text',
+			invalid
+		} = this.properties;
+
+		const classes = [
+			css.root,
+			typeof invalid === 'boolean' ? invalid ? css.invalid : css.valid : null
```
tomdye: i'd argue that we should have classes for `disabled`, `readonly` etc such that they can work with our theming

src/textinput/styles/textinput.css
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/* TODO: better required styles */
+.root input:required {
+	border-left-width: 8px;
+}
+
+/* TODO: this needs non-color design as well */
+.invalid input {
+	border-color: var(--dojo-red);
```
tomdye: we should avoid calling out colours directly in our `css` files. better to have `--dojo-error-color` and `--dojo-valid-color` as red and green don't always mean these things.

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @property onChange	Called when the node's 'change' event is fired
+ * @property onFocus	Called when the input is focused
+ * @property onBlur		Called when the input loses focus
+ */
+export interface TextInputProperties extends ThemeableProperties, FormLabelMixinProperties {
+	type?: string;
+	onInput?(event: Event): void;
+	onChange?(event: Event): void;
+	onFocus?(event: Event): void;
+	onBlur?(event: Event): void;
```
tomdye: Should `value` not be here at all for an initial value?

smhigley: It's still using the `FormLabel` mixin for now. When I test out a separate label widget, all the form widget attributes could either go into each widget, or get separated out into a different mixin.

src/textinput/TextInput.ts
smhigley: Required is added in `FormLabel`

src/textinput/TextInput.ts
smhigley: Should we just pass through all events that can be added to an input? I think those would be:
- `onblur`
- `onblur`

src/textinput/TextInput.ts
smhigley: I personally prefer pseudo-classes, where possible, for things like that. Is there a reason `:disabled` is worse than `:hover` or `:focus` for theming, given it'll still be using the root localized classname?

This might all end up being moot anyway, if the final disabled/readonly/etc design is something that would require a classname to accomplish.

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,154 @@
...
+	disabled?: boolean;
+	formId?: string;
+	invalid?: boolean;
+	label?: string | LabelOptions;
+	maxLength?: number | string;
+	minLength?: number | string;
+	name?: string;
+	placeholder?: string;
+	readOnly?: boolean;
+	required?: boolean;
```
bitpshr: Do we need the `required` attribute? See #22. If a parent component determines if a child form component is valid, when would this ever be used? A parent component would be responsible for saying "this value should exist, since it doesn't, set `invalid` to `true`." This is the same reason something like `pattern` probably isn't necessary. `maxLength` is different since it actually modifies behavior.

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,154 @@
...
+	onKeyPress?(event: KeyboardEvent): void;
+	onKeyUp?(event: KeyboardEvent): void;
+	onMouseDown?(event: MouseEvent): void;
+	onMouseUp?(event: MouseEvent): void;
+	onTouchStart?(event: TouchEvent): void;
+	onTouchEnd?(event: TouchEvent): void;
+	onTouchCancel?(event: TouchEvent): void;
+}
+
+const TextInputBase = ThemeableMixin(WidgetBase);
```
bitpshr: I think you have to export this; try running `grunt dist`, it should fail since this is private.

src/textinput/TextInput.ts
rishson: Wouldn't you need to indicate those fields that should have a value but currently don't, i.e. the field is not invalid but is required?

bitpshr: I guess I thought that would be part of the label text, e.g. including an asterisk in the label itself. If `required` will be used for that purpose and we'll have specific styling, that's fine, we just need to make sure it's used the same way for all form components.

rishson: That closely couples the 'required' with the textual content, which makes it hard to do text validation. Usually the '*' indicator is outside of the actual text content.
I agree that we should have a consistent way of saying 'this is optional/required', 'this is valid/invalid'.

smhigley: I'm more with Andy on this one. Also, I usually don't want to indicate valid/invalid before the user has at least interacted with a form field, but I would want to show required/optional fields. There's also the separate issue that screen readers use and read the required attribute and invalid attribute separately.

rishson: In addition, something like material design puts the mandatory indicator on the field label rather than the input so this might effect our approach.

rishson: screen shot 2017-03-06 at 4 04 50 pm

dylans: FWIW, the dojo 1 approach was tristate... valid/invalid/incomplete... that was you could decide what to show someone before a field was focused. In parallel there is required which influences the tristate but is not part of the state, just a validation rule.

smhigley: @dylans do you think we need anything more than valid/invalid for validation? I'm not sure it's necessary to differentiate incomplete from invalid. I'm also wondering if I should change the property name to `valid`, just because it sounds better to me.

dylans: @smhigley I'm not sure what the right answer is. I think the point I was making is that there are possibly multiple ways to think of it. In Dojo 1 we had 3, incomplete, valid, and invalid. This was only with regards to what message or styling to display. Independently we had required vs optional. I've also seen systems that want to show something the first time you interact with the field and then when it is active (e.g. showing if a password matches the constraints). I don't know the right answer, just that while valid vs not valid is sufficient for submitting the form, it's probably not enough for determining what to display?

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,161 @@
...
+ * @property onMouseUp		Called on the input's mouseup event
+ * @property onTouchStart	Called on the input's touchstart event
+ * @property onTouchEnd		Called on the input's touchend event
+ * @property onTouchCancel	Called on the input's touchcancel event
+ */
+export interface TextInputProperties extends ThemeableProperties {
+	describedBy?: string;
+	disabled?: boolean;
+	formId?: string;
+	invalid?: boolean;
```
bitpshr: We need to be consistent. If radio / checkbox support a valid state (with visuals), so should every other form component, including this.

src/textinput/TextInput.ts
tomdye: IMO the `required` attribute is value due the requirement to show a `required` indicator.

src/textinput/TextInput.ts
smhigley: It does support it, the `invalid` boolean controls `css.valid` and `css.invalid`. It works the same way on radio / checkbox (I even copied the same code over)

rishson: So: 
- `invalid === false` show as valid
- `invalid === true` show as invalid
- `invalid === undefined` show as 'normal', so if user forgets to pass a value, they don't get the nice ❌ ✅ indicators?

smhigley: Yup, the validation is entirely controlled by the parent through `properties.invalid`. Do you think the form widget itself should handle valid/invalid? Or base it on a combination of something like `invalid` and `required`?

rishson: I think you need to be able to set both `invalid` and `required` on the widget directly, and possibly have the `form` be able to do it as well. The widget should be dumb enough to just apply classes based on props. Its up to a higher level widget to do stuff and set the right props IMO.

smhigley: @rishson that's what it's doing right now

bitpshr: Yeah I think this is OK for now.

rishson: Yep - I was agreeing 😃 
I do agree that the props `valid vs invalid` should be consistent across all form elements, so we need to pick one.

smhigley: Right now they are consistent, and use `invalid`. I'd be happy to change them all to `valid`, which does sound better to me if you guys agree.

bitpshr: I vote for `invalid` since it aligns with the DOM attribute.

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,161 @@
...
+				'aria-describedby': describedBy,
+				disabled,
+				'aria-invalid': invalid,
+				maxlength: maxLength ? maxLength + '' : null,
+				minlength: minLength ? minLength + '' : null,
+				name,
+				placeholder,
+				readOnly,
+				'aria-readonly': readOnly ? 'true' : null,
+				required,
```
smhigley: Yup, it has an extra border, but I should add in the label `*` too. Forgot about that :)

bitpshr:
Looks good now. Not sure what's up with the CI.

src/common/tests/unit/all.ts
```diff
@@ -2,3 +2,4 @@ import './main';
 import '../../../dialog/tests/unit/Dialog';
 import '../../../slidepane/tests/unit/SlidePane';
 import '../../../label/tests/unit/Label';
+import '../../../textInput/tests/unit/TextInput';
```
bitpshr: I think this should be `../../../textinput/tests/unit/TextInput` (lowercase directory name)

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,162 @@
...
+
+		if (label) {
+			textinputWidget = w(Label, {
+				classes: this.classes(css.label, ...stateClasses),
+				formId,
+				label
+			}, [ textinput ]);
+		}
+		else {
+			textinputWidget = v('div', {
```
tomdye: As raised by @mwistrand on another pr, do you think this should be a `span` rather than a `div` to reflect the inline nature of the equivalent dom element?

src/textinput/TextInput.ts
```diff
@@ -0,0 +1,162 @@
...
+
+		const stateClasses = [
+			disabled ? css.disabled : null,
+			invalid ? css.invalid : null,
+			invalid === false ? css.valid : null,
+			readOnly ? css.readonly : null,
+			required ? css.required : null
+		];
+
+		const textinput = v('div', { classes: this.classes(css.inputWrapper) }, [
```
tomdye: why do we have a wrapper around the input?

src/textinput/TextInput.ts
smhigley: I think the difference is here `div` isn't replacing the native element, it's wrapping it. We still have an actual `input` inside.

smhigley: I think the difference is here `div` isn't replacing the native element, it's just wrapping it. There's still an `input` element inside. I don't see much point in replacing it with `span`, but I suppose there isn't much drawback either.

src/textinput/TextInput.ts
bitpshr: I _think_ this is because either the Label or a `<div>` will wrap the input, so the same state classes can be applied to each. Otherwise, state classes would either be applied to the Label (a parent of the input) or the input itself, which is odd.

Will defer to @smhigley though.

smhigley: Yup, @bitpshr got it. Otherwise in the css we'd have to have `.invalid .inputWrapper, .invalid.inputWrapper { ... }`

src/textinput/TextInput.ts
tomdye: just that now, if you wanted to make it an inline element, you'd have to set the `div` so `display: inline`, which feels wrong to me.

src/textinput/TextInput.ts
tomdye: this will result in a double `div` wrapper for the case without a `label`

tomdye: I think this one is wrong as there's already the second `div` (which i think should be a `span`) added further down.

src/textinput/TextInput.ts
smhigley: Making it a span puts more restrictions on future additions, though. For example, if we addd error text as a tooltip, the tooltip would also need to be an inline element.

src/textinput/example/index.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Text-based Input Example</title>
+</head>
+<body>
+	<script src="../../../../node_modules/@dojo/loader/loader.min.js"></script>
+	<link rel="stylesheet" href="../../common/styles/widgets.css">
```
tomdye: should this not just be the `textinput.css` file imported?

src/textinput/TextInput.ts
smhigley: @tomdye I tried various other configurations, but this is the only way that allows to add things like an error text tooltip and icons in the future.

tomdye: but then we end up with 
``` html
<div>
   <div>
      <input>
   </div>
</div>
```

src/textinput/example/index.html
smhigley: The other widgets have been using `widgets.css` in their example files. If we change it, we should probably change all of them

src/textinput/TextInput.ts
smhigley: It's the same issue with future additions mentioned above. We need a wrapper directly around the input (unlike `label`, which also wraps the label text) in order to position things like icons and errors. If we make this a (possibly inline) `span`, those elements will also need to be inline.

src/textinput/example/index.html
tomdye: fine, disregard

src/textinput/TextInput.ts
smhigley: Yup, but it's the issue where if we leave off the top-level div, the state classes (valid, invalid, required, etc) don't have a reliable relation to the input wrapper, so there will be confusing extra css selectors.

PR #33, opened 2017-02-20
textarea widget

### LINKED ISSUE #44: Textarea widget
**Enhancement** 
Related to #5 

Features:
- Includes styles for disabled, required, valid, and invalid states
- accepts callback functions for:
  - `onBlur`
  - `onChange`
  - `onClick`
  - `onFocus`
  - `onInput`
  - `onKeyDown`
  - `onKeyPress`
  - `onKeyUp`
  - `onMouseDown`
  - `onMouseUp`
  - `onTouchStart`
  - `onTouchEnd`
  - `onTouchCancel`

smhigley:
**Type:** feature

**Description:**

Resolves #44, adds textarea widget


src/textarea/Textarea.ts
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	render() {
+		const {
+			columns = null,
+			rows = null,
+			wrapText = null
+		} = this.properties;
+
+		return v('textarea', {
```
bitpshr: How would a user set a default value on this?

smhigley: `dojo/widget-core/mixins/FormLabel` is where most of the common form node attributes, including `value`, are added.

src/textarea/Textarea.ts
```diff
@@ -0,0 +1,96 @@
...
+	onKeyUp?(event: KeyboardEvent): void;
+	onMouseDown?(event: MouseEvent): void;
+	onMouseUp?(event: MouseEvent): void;
+	onTouchStart?(event: TouchEvent): void;
+	onTouchEnd?(event: TouchEvent): void;
+	onTouchCancel?(event: TouchEvent): void;
+}
+
+@theme(css)
+export default class Textarea extends ThemeableMixin(FormLabelMixin(WidgetBase))<TextareaProperties> {
```
agubler: Can you convert this to use the "base" pattern? Create the mixin and assign to a `const` and use that to extend. If you run `grunt dist` this should actually fail as it is.

src/textarea/Textarea.ts
```diff
@@ -0,0 +1,96 @@
...
+			typeof invalid === 'boolean' ? invalid ? css.invalid : css.valid : null
+		];
+
+		return v('textarea', {
+			classes: this.classes(...classes).get(),
+			cols: columns ? columns + '' : null,
+			rows: rows ? rows + '' : null,
+			wrap: wrapText,
+			onblur: onBlur,
+			onchange: onChange,
```
agubler: Does this work? Don't we need to define these in the widget class and call `this.onChange` for example? If the consumer changed the function passed to the widget then I believe Maquette would throw an error (you're not allowed to change the function for handlers).

src/textarea/example/index.ts
```diff
@@ -0,0 +1,52 @@
+import { WidgetBase } from '@dojo/widget-core/WidgetBase';
+import { WidgetProperties } from '@dojo/widget-core/interfaces';
+import { ProjectorMixin } from '@dojo/widget-core/mixins/Projector';
+import { StatefulMixin } from '@dojo/widget-core/mixins/Stateful';
+import { v, w } from '@dojo/widget-core/d';
+import Textarea from '../../textarea/Textarea';
+
+export class App extends StatefulMixin(WidgetBase)<WidgetProperties> {
```
agubler: Same applies for the base pattern.

src/textarea/Textarea.ts
```diff
@@ -0,0 +1,158 @@
...
+	onKeyUp?(event: KeyboardEvent): void;
+	onMouseDown?(event: MouseEvent): void;
+	onMouseUp?(event: MouseEvent): void;
+	onTouchStart?(event: TouchEvent): void;
+	onTouchEnd?(event: TouchEvent): void;
+	onTouchCancel?(event: TouchEvent): void;
+}
+
+@theme(css)
+export default class Textarea extends ThemeableMixin(WidgetBase)<TextareaProperties> {
```
bitpshr:
```ts
export const TextareaBase = ThemeableMixin(WidgetBase);

export default class Textarea extends TextareaBase<TextareaProperties> {
```

PR #34, opened 2017-02-20
Radio checkbox 05

smhigley:
**Type:** feature

**Description:**

Addresses #5, adds radio and checkbox widgets


src/checkbox/Checkbox.ts
```diff
@@ -0,0 +1,133 @@
...
+	onClick?(event: MouseEvent): void;
+	onFocus?(event: FocusEvent): void;
+	onMouseDown?(event: MouseEvent): void;
+	onMouseUp?(event: MouseEvent): void;
+	onTouchStart?(event: TouchEvent): void;
+	onTouchEnd?(event: TouchEvent): void;
+	onTouchCancel?(event: TouchEvent): void;
+}
+
+const CheckboxBase = ThemeableMixin(WidgetBase);
```
bitpshr: I think this needs to be exported, otherwise `grunt dist` will complain.

src/checkbox/Checkbox.ts
```diff
@@ -0,0 +1,133 @@
...
+			name,
+			readOnly,
+			required,
+			value
+		} = this.properties;
+
+		const stateClasses = [
+			checked ? css.checked : null,
+			disabled ? css.disabled : null,
+			typeof invalid === 'boolean' ? invalid ? css.invalid : css.valid : null,
```
bitpshr: A double ternary can be confusing to read.

src/checkbox/example/index.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const AppBase = StatefulMixin(WidgetBase);
+export class App extends AppBase<WidgetProperties> {
+	onChange(event: Event) {
+		const {
+			checkboxValues = defaultCheckboxValues
+		} = this.state;
+		const value = (<HTMLInputElement> event.target).value;
+		checkboxValues[value] = (<HTMLInputElement> event.target).checked;
+		this.setState({ checkboxValues: checkboxValues });
```
bitpshr: Not super important, but you could store each checkbox value as a separate state item. That way they don't all re-render every time one changes.

src/checkbox/styles/checkbox.css
```diff
@@ -0,0 +1,78 @@
...
+.inputWrapper::before,
+.inputWrapper::after {
+	box-sizing: border-box;
+	content: '';
+	display: block;
+}
+
+.inputWrapper::before {
+	border: 2px solid var(--form-highlight-color);
+	height: 1.5em;
```
bitpshr: Are we sure that this and other rules using `em` will scale up proportionally?

src/checkbox/styles/checkbox.css
```diff
@@ -0,0 +1,78 @@
...
+/* invalid style */
+.invalid .inputWrapper::before {
+	border-color: var(--dojo-error-color);
+}
+.invalid .inputWrapper::after {
+	background-color: var(--dojo-error-color);
+}
+
+/* valid style */
+.valid .inputWrapper::before {
```
bitpshr: This is odd to me. Are all input widgets going to explicitly support a valid state as opposed to only an invalid state? It's also weird to me that a checkbox that isn't required isn't styled as valid by default.

src/checkbox/Checkbox.ts
```diff
@@ -0,0 +1,133 @@
...
+			readOnly ? css.readonly : null,
+			required ? css.required : null
+		];
+
+		const checkbox = v('div', { classes: this.classes(css.inputWrapper) }, [
+			v('input', {
+				bind: this,
+				classes: this.classes(css.root).get(),
```
bitpshr: Can be removed throughout.

src/checkbox/styles/checkbox.css
smhigley: I was just thinking that some consumers might like separate untouched/invalid/valid styles. React has a "no validation" state as well as valid/invalid/warning: https://react-bootstrap.github.io/components.html#forms-input-addons. Good question, though. I'm fine changing it to whatever the consensus is.

src/checkbox/Checkbox.ts
tomdye: he's referring to the `.get()` part

src/checkbox/styles/checkbox.css
tomdye: surely there's a vanilla / clean style without valid / invalid applied that would show through before the input is focused and in the event that it is not required?

src/checkbox/styles/checkbox.css
```diff
@@ -0,0 +1,78 @@
...
+
+.inputWrapper::before {
+	border: 2px solid var(--form-highlight-color);
+	height: 1.5em;
+	width: 1.5em;
+}
+
+.inputWrapper::after {
+	background-color: var(--form-highlight-color);
+	height: calc(1.5em - 6px);
```
tomdye: what is the `6px` here?

src/checkbox/styles/checkbox.css
smhigley: @tomdye yes there is. That's why I think `.valid` is a good state to have separate from default.

src/checkbox/styles/checkbox.css
smhigley: Two borders, 2px + 2px, plus a small gap of 1px * 2 on the top and bottom.

src/checkbox/styles/checkbox.css
bitpshr: I don't care much either way, but it's worth mentioning that in our unidirectional flow paradigm, a parent component would have to explicitly set `valid` just like it would have to set `invalid`. Do we really think users will do this?

Either way, all form components should either support this or not to stay consistent.

src/checkbox/styles/checkbox.css
smhigley: Yup! They do, you can try it in-browser in `example/` if you'd like. Just change `font-size` on the label.

src/checkbox/styles/checkbox.css
smhigley: It think the verdict is `properties.invalid` controls `.valid` and `.invalid` (or neither if undefined)

src/checkbox/styles/checkbox.css
tomdye: Ok, was just thinking that it should be clearer such that someone else doesn't come in and change it without understanding.

smhigley: Yeah, I think ideally this will get split out to a theme and probably changed. I've just been writing more than basic styles to show that you can with this widget structure.

PR #46, opened 2017-03-01
Form label widget

smhigley:
**Type:** feature

Addresses comments on form widget PRs about the form label mixin. The label widget eliminates the `afterRender` decorator and allows more flexibility in node structure and properties.

It includes a `classes` property, since form field widgets using it should be able to set state classes (e.g. invalid) on the label node.

An example of a form widget using the label widget instead of the FormLabel mixin is here: https://github.com/smhigley/widgets/blob/form-label/src/textinput/TextInput.ts


src/label/Label.ts
```diff
@@ -0,0 +1,77 @@
...
+			labelProps = assign({}, labelDefaults, label);
+		}
+
+		// add label text node to children
+		const labelText = v('span', {
+			innerHTML: labelProps.content,
+			classes: this.classes().fixed(labelProps.hidden ? css.visuallyHidden : null).get()
+		});
+		if (labelProps.before) {
+			this.children.unshift(labelText);
```
agubler: You can drop the `.get` from the classes calls

src/label/Label.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @type LabelWidgetProperties
+ *
+ * Properties that can be set on a Label component
+ *
+ * @property classes	Optional classes to be set on the label node
+ * @property formId		ID of a form element associated with the form field
+ * @property label		Label settings for form label text, position, and visibility
+ */
+export interface LabelProperties extends ThemeableProperties {
+	classes?: { [key: string]: boolean };
```
agubler: Perhaps a problem with this is we are requiring the consumer to ensure that they negate a class if it needs to be removed on a subsequent render? Or are we expecting consumers to use `this.classes` internally to pass to the label?

smhigley: I was expecting users to pass `this.classes(...)` in to `label.properties.classes`, yes. Is there a different or stricter way you'd prefer?

agubler: Okay. I hadn't thought in any detail, but at least that way themable still does all the work :)

src/label/styles/label.css
```diff
@@ -0,0 +1,11 @@
+/* TODO: this should go in a util.css */
```
rishson: Should this go in `variables.css` in widgets/styles?
I think most things in the `variables.css` should be in a theme, and classes like this *do* belong in a x-theme variables file.

src/label/Label.ts
rishson: Could you type to `ClassNames` from `Themeable` though? Should really reference the type used in the theme system.

src/label/styles/label.css
smhigley: It's not exactly a variable though, and I'm not sure it should be in a theme. I think it and `.hidden` should go in something like `util.css`, and should probably not be themeable. It's hard to think of a case where a widget consumer should be able to override it.

PR #60, opened 2017-03-08
Select 05

### LINKED ISSUE #59: Select Widget
**Enhancement** 
Related to #5 

**Features:**
- Includes styles for disabled/readonly, required, valid, and invalid states
- accepts callback functions for:
  - `onBlur`
  - `onChange`
  - `onClick`
  - `onFocus`
  - `onKeyDown`
  - `onKeyPress`
  - `onKeyUp`
  - `onMouseDown`
  - `onMouseUp`
  - `onTouchStart`
  - `onTouchEnd`
  - `onTouchCancel`
- Options are added in an object with format `{ [key: string]: value: string }`

smhigley:
**Type:** feature

**Description:**

Resolves #59


src/select/Select.ts
```diff
@@ -0,0 +1,155 @@
...
+ * @property onTouchCancel	Called on the input's touchcancel event
+ */
+export interface SelectProperties extends ThemeableProperties {
+	describedBy?: string;
+	disabled?: boolean;
+	formId?: string;
+	invalid?: boolean;
+	label?: string | LabelOptions;
+	name?: string;
+	options?: { [key: string]: string };
```
bitpshr: How can I disable an individual option? Also, if multiple selection support is added, how can I default multiple options as `selected`?

src/select/Select.ts
```diff
@@ -0,0 +1,155 @@
...
+ * @property onKeyDown		Called on the input's keydown event
+ * @property onKeyPress		Called on the input's keypress event
+ * @property onKeyUp			Called on the input's keyup event
+ * @property onMouseDown	Called on the input's mousedown event
+ * @property onMouseUp		Called on the input's mouseup event
+ * @property onTouchStart	Called on the input's touchstart event
+ * @property onTouchEnd		Called on the input's touchend event
+ * @property onTouchCancel	Called on the input's touchcancel event
+ */
+export interface SelectProperties extends ThemeableProperties {
```
bitpshr: We should probably support multiple selection via a property for `multiple`.

src/select/Select.ts
```diff
@@ -0,0 +1,155 @@
...
+			invalid ? css.invalid : null,
+			invalid === false ? css.valid : null,
+			readOnly ? css.readonly : null,
+			required ? css.required : null
+		];
+
+		/* create option nodes */
+		const optionNodes = [];
+		for (let key in options) {
+			optionNodes.push(v('option', {
```
bitpshr: Need to support `selected` and `disabled` somehow.

src/select/Select.ts
```diff
@@ -0,0 +1,155 @@
...
+				value: key,
+				innerHTML: options[key]
+			}));
+		}
+
+		const select = v('div', { classes: this.classes(css.inputWrapper) }, [
+			v('select', {
```
bitpshr: Need to support `multiple`.

src/select/styles/select.css
```diff
@@ -0,0 +1,81 @@
+@import '../../common/styles/variables.css';
+
+.input {
+	appearance: none;
+	background-color: transparent;
+	border: 3px solid var(--input-border-color);
```
bitpshr: Can any of these values be pulled out into variables?

src/select/Select.ts
smhigley: So I've actually read that multiple selects have poor usability on assistive tech -- essentially the way you select multiple items is confusing and unintuitive, and checkboxes are recommended instead. I was thinking of making a separate multi-select widget that uses checkboxes under the hood to accomplish the same thing. How high priority do you think a multi select would be? I thought they were a dying UI pattern, but I might be a bit biased because I've never actually used one in the wild.

bitpshr: I still think we should support `multiple` despite its poor usability since it's thing in browsers and people use it. We shouldn't make it so that its impossible to do, we should just warn against it.

Also, for what it's worth, I do think we'll either have a multi-select component or add multiple selection support to ComboBox.

smhigley: I'm still not entirely convinced. We're already cutting down on browser functionality by not supporting `<opt-group>`, and I'm fine being a little opinionated about good and bad widget patterns, as lot as it's possible to achieve the functionality users might want. If we add a separate multi-select component with good usability, what's the added benefit of supporting multi-select here with poor usability? Additionally, supporting multi-select here should probably start a discussion on whether we want to support non-native styling. I added some dom structure to allow better custom styles, but it may make sense to remove it if we're also doing multi-select.

smhigley:
Closed to add non-native option with multi-select and the ability to style individual options

PR #67, opened 2017-03-10
Range slider

### LINKED ISSUE #4: Range slider widget
@smhigley commented on [Fri Dec 23 2016](https://github.com/dojo/widget-core/issues/196)

**Enhancement**
Add a range slider with custom track and output based on this mock: [http://codepen.io/smhigley/pen/ObWbdy](http://codepen.io/smhigley/pen/ObWbdy). 

Related to #85.

smhigley:
**Type:** feature

**Description:**

Resolves #4


src/slider/styles/slider.css
```diff
@@ -0,0 +1,193 @@
...
+	content: "";
+	left: 0;
+	position: absolute;
+	top: 0;
+}
+
+.thumb {
+	background-color: var(--input-highlight-color);
+	border-radius: 50%;
+	box-shadow: inset 0 0 3px 0 var(--dojo-shade-gray);
```
bitpshr: Can any of these move into variables, or are the too specific to the Slider?

src/slider/styles/slider.css
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/* required style, only applies to inputs with a label */
+.required span::after {
+	content: '*';
+	color: var(--dojo-error-color);
+}
+
+/* invalid style */
+.invalid .inputWrapper::before {
```
bitpshr: Could you add an example of a validated and / or required Slider?

src/slider/Slider.ts
```diff
@@ -0,0 +1,199 @@
...
+	onKeyDown (event: KeyboardEvent) { this.properties.onKeyDown && this.properties.onKeyDown(event); }
+	onKeyPress (event: KeyboardEvent) { this.properties.onKeyPress && this.properties.onKeyPress(event); }
+	onKeyUp (event: KeyboardEvent) { this.properties.onKeyUp && this.properties.onKeyUp(event); }
+	onMouseDown (event: MouseEvent) { this.properties.onMouseDown && this.properties.onMouseDown(event); }
+	onMouseUp (event: MouseEvent) { this.properties.onMouseUp && this.properties.onMouseUp(event); }
+	onTouchStart (event: TouchEvent) { this.properties.onTouchStart && this.properties.onTouchStart(event); }
+	onTouchEnd (event: TouchEvent) { this.properties.onTouchEnd && this.properties.onTouchEnd(event); }
+	onTouchCancel (event: TouchEvent) { this.properties.onTouchCancel && this.properties.onTouchCancel(event); }
+
+	render() {
```
bitpshr: This needs to return a `DNode` otherwise a production build will fail (TS should complain.)

src/slider/styles/slider.css
smhigley: I considered it, but we already have two box shadows in variables already :P. I can't use either of those b/c the thumb is too small for them. I think it might actually be cleaner overall to keep this one specific to Slider, at least until we get official styles for all this.

Issue #37, opened 2017-02-21
Specifications for widgets


rishson:
Hi all.
Going forward, I think we need much clearer intentions when creating widgets. What I'd like to propose is:

1) each widget 'spec' is a separate issue (and PR)
2) spec lists features needed for initial version, and things excluded for now

What I'd like to do, is create initial issues for all in-flight widgets and have a standard set of features that either do or do not need to be completed, e.g.

[Template spec](https://github.com/dojo/widgets/wiki/Widget-specification-template) is now on the wiki for this repo.

rishson:
@smhigley @tomdye @bitpshr comments appreciated so we can resolve asap.
Once we have agreement, I'll create widget spec issues for everything in-flight.

bitpshr:
This sounds fine to me. I have a few verbiage nitpicks (components don't emit events, they call a function property) but this conveys the right ideas.

rishson:
@bitpshr ta - updated.

rishson:
- what responsive design features are needed

rishson:
We now have a basic specification that we agree on.
I'll leave this open so we have an issue that you can copy the spec details from, until we have a suitable other location for this information.

rishson:
Closing as the widget spec is now on the `widget` repo wiki.

PR #68, opened 2017-03-12
TabPane

### LINKED ISSUE #61: TabPane
**Enhancement**

Create a TabPane component.

Note that features such as showing tabs that don't fit within available real estate in a menu can be added at a later time.

### Specification

**Callback props**:
```ts
activeIndex?: number;
loadingIndex?: number;
tabs?: TabConfig[];
onRequestTabChange?(index: number): void;
onRequestTabClose?(tabs: TabConfig[]): void;
```

**Interaction**
- Clicking on a tab triggers `onRequestTabChange`
- Clicking on a tab close button triggers `onRequestTabClose`

**States**
_None._

**Value add**
A TabPane would solve the common UX issue of needing to show multiple sets of information as designated by the currently selected tab.

**Controlled or uncontrolled**
Controlled.

**Icons**
- Close Icon

**Responsive design**
None, until dictated by initial theme mockups from @itorrey.

bitpshr:
**Type:** feature

**Description:**

This PR introduces a TabPane component. Tabs can be closeable, disabled, and can load asynchronously.

Initial API:
```ts
activeIndex: number;
alignButtons?: Align;
onRequestTabChange?(index: number, key: string): void;
onRequestTabClose?(index: number, key: string): void;
```

Example uses:

_Basic_
```ts
w(TabPane, {
	activeIndex: this.state.activeIndex,
	onRequestTabChange: index => this.setState({ activeIndex: index })
}, [
	w(Tab, {
		key: 'foo',
		label: 'Foo'
	}, [ 'Foo' ]),
	w(Tab, {
		disabled: true,
		key: 'bar',
		label: 'Bar'
	}, [ 'Bar' ])
])
```

_Closeable tabs_
```ts
w(TabPane, {
	activeIndex: this.state.activeIndex,
	onRequestTabChange: index => this.setState({ activeIndex: index }),
	onRequestTabClose: (index, key) => this.setState({ closedKeys: [...this.state.closedKeys, key] })
}, [
	w(Tab, {
		key: 'foo',
		label: 'Foo'
	}, [ 'Foo' ]),
	!includes(this.state.closedKeys, 'bar') ? w(Tab, {
		closeable: true,
		key: 'bar',
		label: 'Bar'
	}, [ 'Bar' ]) : null
])
```

_Async tabs_
```ts
w(TabPane, {
	activeIndex: this.state.activeIndex,
	onRequestTabChange: index => {
		this.setState({
			activeIndex: index,
			loading: true
		});
		asyncOperation.then(data => {
			this.setState({
				loading: false,
				someData: data
			})
		});
	}
}, [
	w(Tab, {
		key: 'foo',
		label: 'Foo'
	}, [ loading ? 'Loading...' : this.state.someData.foo ]),
	w(Tab, {
		key: 'bar',
		label: 'Bar'
	}, [ loading ? 'Loading...' : this.state.someData.bar ])
])
```

Resolves #61


agubler:
I have a question regarding widget functionality. Do we detail specific acceptance criteria of the widget delivery or is it just the high level requirements / API? It might not directly related to this widget but generally do we think this would be a good practice to get into? It would help facilitate a form of TDD also.

bitpshr:
@agubler: The [component specification tickets](https://github.com/dojo/widgets/issues/37) are pretty [high level](https://github.com/dojo/widgets/issues/61). I think we could get more detailed, like agreeing on a properties interface and a public API.

agubler:
I think we should be detailing specific behaviour that is the acceptance criteria for the requirements of the widgets personally. Certainly worth a discussion about what level of detail would be most appropriate :) maybe not on this pull request though!

src/tabpane/TabButton.ts
```diff
@@ -0,0 +1,79 @@
...
+
+		const children = closeable ? this.children.concat([
+			v('button', {
+				classes: this.classes(css.close),
+				innerHTML: 'close tab',
+				onclick: this._onCloseClick
+			})
+		]) : this.children;
+
+		return v('div', {
```
smhigley: This needs some a11y stuff:
- `tabIndex: 0` and focus style
- `role: 'tab'`
- `aria-selected` of true or false, depending on `active`
- an `id` for the tab container to point to
- `aria-controls` pointing to the id of the tab container

src/tabpane/TabButton.ts
```diff
@@ -0,0 +1,79 @@
...
+			classes: this.classes(
+				css.tabButton,
+				active ? css.activeTabButton : null,
+				disabled ? css.disabledTabButton : null
+			),
+			onclick: this._onClick
```
smhigley: This also needs keyboard interaction -- W3 specs give two options: either arrow keys can automatically change tabs, or arrow keys can change focus and space/enter can open the tab content. I'm partial to the former because it'll be simpler to implement, and it seems to be the recommended method as well. There might be some issue with async tabs though; check out the first note in the green box: https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel

The delete key should close closable tabs as well. Home/End should go to the first/last tab, and arrow keys should wrap last -> first and vice versa.

(commenting here, but it might make sense to put the listeners on the parent)

src/tabpane/Tab.ts
```diff
@@ -0,0 +1,35 @@
...
+};
+
+export const TabBase = ThemeableMixin(WidgetBase);
+
+@theme(css)
+export default class Tab extends TabBase<TabProperties> {
+	render(): DNode {
+		const { loading } = this.properties;
+
+		return v('div', {
```
smhigley: Tab content also needs `role: tabpanel`, `aria-labelledby` pointing to the id of the correct tab button, and an `id` for the tab's `aria-controls` attribute.

src/tabpane/TabPane.ts
```diff
@@ -0,0 +1,181 @@
...
+			case Align.bottom:
+				alignClass = css.alignBottom;
+				children.reverse();
+				break;
+			case Align.left:
+				alignClass = css.alignLeft;
+				break;
+		}
+
+		return v('div', {
```
smhigley: This one needs `role: tablist`, and if alignButtons is left or right, it'll need `aria-orientation` set to vertical (`horizontal` is the default, but it wouldn't hurt to set it explicitly either)

tomdye:
@bitpshr you've got combo box changes in this PR too

bitpshr:
@tomdye I just alphabetized imports and fixed a few documentation typos in ComboBox and SlidePane. Will be better about separating work in different PRs from now on.

src/tabpane/TabButton.ts
```diff
@@ -0,0 +1,152 @@
...
+			controls,
+			disabled,
+			id
+		} = this.properties;
+
+		const children = closeable ? this.children.concat([
+			v('button', {
+				classes: this.classes(css.close),
+				innerHTML: 'close tab',
+				onclick: this._onCloseClick
```
smhigley: This should probably also have `tabindex: -1` unless its parent tab is the active one, and a focus style. Right now it's in the tab order all the time, and focus is hidden, so it's a little confusing to land on.

src/tabpane/TabButton.ts
```diff
@@ -0,0 +1,152 @@
...
+				break;
+			// Home
+			case 36:
+				onHomePress && onHomePress();
+				break;
+			// End
+			case 35:
+				onEndPress && onEndPress();
+				break;
+		}
```
smhigley: The vertically oriented tabs should also listen for up/down key presses to change tabs. They probably shouldn't do anything when it's horizontally oriented to minimize hijacking their normal scrolling functionality. Left/right arrows can keep their current behavior all the time.

bitpshr:
@smhigley: Ready for another review when you have time.

src/tabpane/TabButton.ts
```diff
@@ -0,0 +1,167 @@
...
+		} = this.properties;
+
+		if (disabled) {
+			return;
+		}
+
+		// Accessibility
+		switch (event.keyCode) {
+			// Escape
+			case 27:
```
tomdye: Do you think this would be more readable with a `const ESCAPE_KEY = 27;` instead? If the code needs comments to be understood then it's likely that it could be re-written better.

src/tabpane/TabButton.ts
```diff
@@ -0,0 +1,167 @@
...
+			classes: this.classes(
+				css.tabButton,
+				active ? css.activeTabButton : null,
+				disabled ? css.disabledTabButton : null
+			),
+			id,
+			key: 'tab-button',
+			role: 'tab',
+			tabIndex: active ? 0 : -1,
+			onclick: this._onClick,
```
tomdye: I don't `believe` that we've used underscores anywhere else for private functions, is this the new approach?

src/tabpane/TabButton.ts
tomdye: Creating a common key-code module would be even more useful as this pattern is repeated a lot.

src/tabpane/TabPane.ts
```diff
@@ -0,0 +1,238 @@
...
+
+		const validIndex = this._validateIndex(index, backwards);
+
+		if (validIndex !== null && validIndex !== activeIndex) {
+			const key = (<TabProperties> (<WNode> this._tabs[validIndex]).properties).key;
+			onRequestTabChange && onRequestTabChange(validIndex, key);
+		}
+	}
+
+	protected selectLastIndex() {
```
tomdye: should all the selection not be handled via prop changes and the render function?

bitpshr: This is handled by a prop change and a `render`: `selectLastIndex` calls `selectIndex` which eventually calls `onRequestTabChange`.

src/tabpane/TabButton.ts
bitpshr: All of widgets uses underscores for any private member. I started doing this after some past PR comments saying we were going to be consistent about doing this.

PR #71, opened 2017-03-13
Remove individual example pages

### LINKED ISSUE #27: Consider dropping the index.html per example
Nearly all the example index.html's are identical apart from the `require` they make to the actual example module. I get the feeling this may be a maintenance burden in the future.

Perhaps we should just have one index.html somewhere near the root that takes the module id of the example as a query string.

bitpshr:
**Type:** bug

**Description:**

This PR adds a common example HTML page at `common/example/index.html` that looks for a `module` URL parameter with the name of the component example to load:

```
http://localhost:8080/_build/src/common/example/?module=SlidePane
```

I also included a few minor fixes for production builds (adding a return type to `render` functions.)

Resolves #27


src/common/example/index.html
```diff
@@ -1,20 +1,13 @@
 <!DOCTYPE html>
 <html>
 <head>
-	<title>SlidePane Example</title>
+	<title>Example</title>
 	<meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1">
```
smhigley: Just being slightly nitpicky, but everything else looks so good! Could we do `<meta name="viewport" content="width=device-width, initial-scale=1">` and drop the min/max scale? They tend to be bad for usability, especially for people who zoom in manually to read.

Issue #52, opened 2017-03-06
Menu widget


rishson:
Create a `menu` widget.

Initial thoughts: menu as in 'full width page header that contains links to other pages | dropdowns | logos | searchbox etc.

Spec:

- list of callback funcs that can be passed in props
some kind of 'onHide' or 'onPageScroll' that indicates when the menu would be scrolled off screen (so controlling component can reduce hight etc)

- mouse/keyboard interactions (if any)
as per individual controls passed to menu

- mandatory/valid/empty/wait states (if any)
as per individual controls passed to menu

- value-add of the widget (e.g. why use rather than just use VDom directly)
ability to easily style to make sticky, align to top, bottom, left, right..
responsive on mobile
ability to vertically align items of different heights (for a horizontal menu)

- is the widget controlled/uncontrolled
I am imagining that you would pass the menu child widgets, therefore it is controlled.

- list of any icons needed
none

- is the widget responsive
yes, on mobile, this widget should shrink to a hamburger menu

bitpshr:
Here's what I was thinking:

A NavBar would go across the top of a page and can contain a logo or header text. It can also contain a list of MenuItems, Menus that contains MenuItems, or both (think Bootstrap NavBars with links and dropdowns to other links). The parent NavBar should collapse its child MenuItems and Menus to use a SlidePanel to show them when screen width goes below a designated breakpoint.

smhigley:
Do we want to build the static/fixed position into this, though? I think that might be good for a separate container widget, since it's pretty common to want all sorts of random things in a scrolling header (or footer).

bitpshr:
I think a MenuItem should be able to be whatever you want (such as a search input). I also think the NavBar should be fixed (and thus collapse its children into a SlidePane) based on a property. If that property is false, the menu isn't fixed and doesn't collapse.

mwistrand:
Rather than limiting a `NavBar` to just a logo/header text and menu items, it would be better to allow any number of children as required, with any menu items being contained in a separate `Menu` widget.

smhigley:
@bitpshr I'm not sure the fixed positioning should be tied to collapsing into a SlidePane. What if someone wants a desktop-width uncollapsed fixed nav bar? I like allowing MenuItems to be anything, that solves the flexible scrolling header issue.

bitpshr:
Maybe we could use separate properties to determine menu collapsing and fixed positioning?

mwistrand:
[...]

PR #79, opened 2017-03-21
Menu widget

mwistrand:
**Type:** feature

**Description:**

Provides the base `Menu` widget for #52. Once dojo/widget-core#415 lands, the `DropDown` widget that extends `Menu`, and the `NavBar` widget that relies on both can also be included to completely fulfill #52.


src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	protected _toggleDisplay(isHidden?: boolean) {
+		const {
+			onRequestHide,
+			onRequestShow
+		} = this.properties;
+
+		if (typeof isHidden === 'undefined') {
+			const { hidden = this._getDefaultHidden() } = this.properties;
```
agubler: Also what is the difference between `isHidden` and `this.properties.hidden`?

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+			href: url || undefined,
+			onclick: this.onClick,
+			onkeypress: this.onKeypress,
+			role: 'menuitem',
+			tabIndex : disabled ? -1 : tabIndex,
+			target: url && external ? '_blank' : undefined
+		}), label ? [ label ] : undefined);
+
+		if (this.children.length) {
+			const children = this.children as HNode[];
```
agubler: Why do you need the cast?

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+
+	/**
+	 * Applies only when a URL is provided. If `true`, the URL will be opened in a new window.
+	 */
+	external?: boolean;
+
+	/**
+	 * Returns an object of aria properties to apply to the widget's DOM element.
+	 */
+	getAriaProperties?: () => { [key: string]: string; };
```
agubler: This doesn't feel like an external property?

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+	tabIndex?: number;
+
+	/**
+	 * A URL to navigate to on click.
+	 */
+	url?: string;
+}
+
+@theme(css)
+export class MenuItem extends ThemeableMixin(WidgetBase)<MenuItemProperties> {
```
agubler: Same comment about creating a base `const` for the expression result.

src/menu/example/index.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
```
agubler: Don't need the specific widget example index.html anymore, this is centralised in `common`

mwistrand: Thanks, I just remembered I hadn't moved this yet last night.

src/menu/MenuItem.ts
mwistrand: I might not anymore. I'll double check, but I did need it at one point.

src/menu/MenuItem.ts
mwistrand: It doesn't, but this will be used in a few different ways requiring different ARIA properties. For example, use in `Menu` requires one set of properties, whereas use with a drop down would require other properties. This seemed like the easiest way to provide any flexibility needed with hard-coding a bunch of scenarios directly within `MenuItem`.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+			}, properties));
+		}
+	}
+
+	protected _afterCreate(element: HTMLElement) {
+		const { animate = true } = this.properties;
+		this._initialRender = false;
+
+		if (animate) {
+			const hidden = element.classList.contains(css.hidden);
```
mwistrand: It can, but I got tired of `const { hidden = this._getDefaultHidden() }` everywhere.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+			nested,
+			role = 'menu'
+		} = this.properties;
+
+		const label = this.renderLabel(id);
+		const menu = v('nav', {
+			id,
+			role,
+			classes: this.classes.apply(this, this._getMenuClasses(hidden, Boolean(label))),
+			afterCreate: this._afterCreate,
```
mwistrand: Great. I'll update those.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+}
+
+function getMenuHeight(menuElement: HTMLElement): number {
+	const maxHeight = parseInt(getComputedStyle(menuElement).getPropertyValue('max-height'), 10);
+	return Math.min(menuElement.scrollHeight, (isNaN(maxHeight) ? Infinity : maxHeight));
+}
+
+@theme(css)
+export class Menu extends ThemeableMixin(WidgetBase)<MenuProperties> {
+	protected _wasOpen = false;
```
mwistrand: Ah; didn't realize about the prefix. There is a lot of shared functionality between `Menu` and a future `DropDown` widget that will extend it, in which case using `private` is not really an option.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+	private _hideTimer: Handle;
+	private _initialRender = true;
+
+	onLabelClick() {
```
mwistrand: These types of functions aren't private in the other widgets?

agubler: They should be if they are not an API intended to be overridden to modify behaviour in widgets that extend it. I think @bitpshr did get round to updating Combo.

src/menu/Menu.ts
agubler: The properties should be the master (I don't see how that is any less tiresome)

src/menu/MenuItem.ts
agubler: Do @bitpshr or @smhigley have any suggestions for the ARIA properties of `Menu`?

src/menu/Menu.ts
bitpshr: The visibility throughout was updated as part of the ComboBox PR. Most things are `private` as @agubler said.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+
+		if (!disabled && expandOnClick) {
+			this._toggleDisplay();
+		}
+	}
+
+	onLabelKeypress(event: KeyboardEvent) {
+		const { disabled } = this.properties;
+
+		if (!disabled && event.key === 'Enter') {
```
bitpshr: Safari is [drunk](http://caniuse.com/#feat=keyboardevent-key) when it comes to `event.key`. We need to use `keyCode` for now.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+		}
+
+		if (typeof animate === 'function') {
+			return animate(element, hidden);
+		}
+
+		this._animate(element, hidden);
+	}
+
+	protected _animate(element: HTMLElement, hidden: boolean) {
```
mwistrand: Animating to and from `auto` is not possible with CSS.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+
+	onMenuFocus() {
+		const { disabled, hidden = this._getDefaultHidden() } = this.properties;
+		if (!disabled && hidden) {
+			const onRequestShow = this.properties.onRequestShow;
+			onRequestShow && onRequestShow();
+		}
+	}
+
+	onMenuMouseEnter() {
```
bitpshr: Just curious about UX: if I tab into a menu's label, it opens, and then I can use the enter key to continue to toggle it only if `expandOnClick` is `false`? For a11y, should we remove the `expandOnClick` check when using the enter key?

src/menu/Menu.ts
bitpshr: The animation is being done using CSS. It just needs an explicit value to animate to besides `auto`. @mwistrand see comment above re: `marginTop` and `offsetHeight`.

src/menu/Menu.ts
mwistrand: Ok. I'll have to find some way to test those methods that doesn't require using `<any>` hacks that will break once we start emitting directly to ES6. I should be able to access the event listeners directly of `vnode.properties`, correct?

src/menu/Menu.ts
mwistrand: `isHidden` is a way to force which state is requested. I'll update the parameter name to make that obvious.

src/menu/Menu.ts
bitpshr: +1

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+			return;
+		}
+
+		if (!animate) {
+			// In case `animate` was previously `true`, remove any `height` property set on the node.
+			element.style.height = null;
+			return;
+		}
+
+		if (typeof animate === 'function') {
```
bitpshr: Are we sure this will work correctly since opening and closing is only occurring after updates, not after creation? I think it's possible that some combination of upstream property changes could trigger the Menu to be recreated vs. updated, and if `hidden` changes as part of those upstream property changes, opening and closing should still occur.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+export interface MenuItemProperties extends ThemeableProperties {
+	/**
+	 * Indicates whether the menu is disabled. If true, then the widget will ignore events.
+	 */
+	disabled?: boolean;
+
+	/**
+	 * Applies only when a URL is provided. If `true`, the URL will be opened in a new window.
+	 */
+	external?: boolean;
```
bitpshr: Wouldn't a user just put an anchor as the child to this MenuItem? I don't think the MenuItem should handle link behavior.

src/menu/MenuItem.ts
bitpshr: @smhigley may have specifics, but you can check https://www.w3.org/WAI/tutorials/menus/. @smhigley and I have been adding a camel-cased property without the "aria" for each ARIA attribute that needs to be set externally. [Example](https://github.com/dojo/widgets/blob/master/src/textinput/TextInput.ts#L16). We should stay consistent.

src/menu/MenuItem.ts
bitpshr: And exporting!

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+		} = this.properties;
+
+		const ariaProperties = getAriaProperties && getAriaProperties() || Object.create(null);
+		const classes = this.classes(
+			hasMenu ? css.menuLabel : css.menuItem,
+			disabled ? css.disabled : null,
+			selected ? css.selected : null
+		);
+
+		const labelNode = v('a', assign(ariaProperties, {
```
bitpshr: You could spread `ariaProperties` instead of assigning.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+			'aria-disabled': disabled,
+			classes,
+			href: url || undefined,
+			onclick: this.onClick,
+			onkeypress: this.onKeypress,
+			role: 'menuitem',
+			tabIndex : disabled ? -1 : tabIndex,
+			target: url && external ? '_blank' : undefined
+		}), label ? [ label ] : undefined);
```
bitpshr: This is odd to me. Menus don't have to contain links so I don't think we should be always creating an anchor. I also don't think the MenuItem should handle hrefs or targets or anything links-specific. It should just put its children in an element so it can still capture events, and that's it. Let the user make the child an anchor if they want.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+	 * between the visible and hidden states, or a function that manually handles the animation. If true (the default),
+	 * then the menu will slide into and out of view like an accordion. If false, then any animation should be handled
+	 * in the CSS, as the menu will just snap open/shut.
+	 */
+	animate?: boolean | ((element: HTMLElement, hidden: boolean) => void);
+
+	/**
+	 * Indicates whether the menu is disabled. If true, then the widget will ignore events.
+	 */
+	disabled?: boolean;
```
mwistrand: If `label` is specified, then a `MenuItem` is created and `disabled` is passed to it, which keeps it from being expanded.

src/menu/Menu.ts
bitpshr: All of our existing tests use the hacky `(<any> menu)._onFoo()` method for testing. I'd just use that for now; we'll have to do a full sweep of the components once @kitsonk's harness lands anyway.

src/menu/Menu.ts
mwistrand: Good catch. I'll add a guard to use `event.key` when it's available (since `keyCode` has been removed from the standards).

src/menu/MenuItem.ts
mwistrand: I'll update to explicitly list all possible ARIA attributes, which will also make destructuring possible/`assign`ing unnecessary.

src/menu/Menu.ts
mwistrand: So the same instance can be "created" more than once? In that case, some updates will be required.

src/menu/example/index.html
```diff
@@ -0,0 +1,26 @@
...
+<html>
+<head>
+	<title>Menu Example</title>
+</head>
+<body>
+	<script src="../../../../node_modules/@dojo/loader/loader.min.js"></script>
+	<link rel="stylesheet" href="../../common/styles/widgets.css">
+	<link rel="stylesheet" href="./styles/app.css">
```
mwistrand: Sounds good. I knew that was being separated out eventually, so I just did that here. I'll move it back into the main `menu.css`, perhaps with notes disambiguating any styles that could go either way.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+	role?: Role;
+}
+
+function getMenuHeight(menuElement: HTMLElement): number {
+	const maxHeight = parseInt(getComputedStyle(menuElement).getPropertyValue('max-height'), 10);
+	return Math.min(menuElement.scrollHeight, (isNaN(maxHeight) ? Infinity : maxHeight));
+}
+
+@theme(css)
+export class Menu extends ThemeableMixin(WidgetBase)<MenuProperties> {
```
agubler: Yup sorry, I should have put that (basically need to export all the things that are used publicly)

src/menu/Menu.ts
agubler: Okay. Do you think that we should expose a property that can provide an overriding function for the animation?

src/menu/Menu.ts
agubler: What do you think about using argument defaults?

```ts
class Test { 

    properties: any = {};

    private _getDefaultHidden(): boolean {
        return true;
    }

    private _toggleDisplay(isHidden = this.properties.hidden || this._getDefaultHidden()) { 

    }
}
```

agubler: Could be cleaner if `this_getDefaultHidden` returns the property `hidden` if it existed

src/menu/Menu.ts
bitpshr: I think...@agubler @matt-gadd can you confirm?

src/menu/Menu.ts
mwistrand: @agubler We already do. `properties.animate` can be a boolean or a function. If it's a boolean, then it enables/disables the default animation. If it's a function, then the built-in `_animate` function is bypassed in favor of the property function.

agubler: I guess my question was, do we think it is valid to accept an overriding function for the animate?

(i'm happy with the boolean switch)

mwistrand: Gotcha. `animate` as a function was a last minute add. Given that as long as the height is set in the build-in function, most animations can be handled correctly in CSS, I'll remove this option.

src/menu/Menu.ts
mwistrand: There's a dedicated `keypress` event for using the enter key (`click` fires on enter for links, but not in every scenario), so keyboard interaction still works correctly.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+	/**
+	 * The text or properties for a MenuItem widget that is used to control the menu.
+	 */
+	label?: string | MenuItemProperties;
+
+	/**
+	 * Indicates whether the menu is nested within another menu. Useful for styling, this does not affect
+	 * functionality. Defaults to false.
+	 */
+	nested?: boolean;
```
mwistrand: Strictly speaking, we could use child selectors to manage the same thing. But since `composes` only works with single selectors, I added the extra class to be safe.

src/menu/Menu.ts
```diff
@@ -0,0 +1,311 @@
...
+	 */
+	onRequestShow?: () => void;
+
+	/**
+	 * The value to use for the menu's `role` property. Defaults to "menu".
+	 */
+	role?: Role;
+}
+
+function getMenuHeight(menuElement: HTMLElement): number {
```
mwistrand: @bitpshr That is pretty clean, and using that would remove the need to reset the transition in `animate` when collapsing. Since the `animate` property lets the user switch the built-in animation off , I'd have to make sure setting the top margin doesn't conflict with the default `.hidden { height: 0 }` and `.visible { height: auto }`, but I'll prototype it out.

mwistrand: I got this working (for the most part) with `margin-top`, but noticed when tabbing into the menu, it feels a little odd having to wait for the top of menu to slide in before seeing the first link focused. So, I left this alone for the moment.

src/menu/Menu.ts
```diff
@@ -0,0 +1,299 @@
...
+		const {
+			disabled,
+			expandOnClick = true
+		} = this.properties;
+
+		if (!disabled && !expandOnClick) {
+			this._hideTimer && this._hideTimer.destroy();
+			this.toggleDisplay(true);
+		}
+	}
```
smhigley: When the dropdown can be opened on hover, we should probably consider having it open on focus as well

src/menu/Menu.ts
```diff
@@ -0,0 +1,299 @@
...
+
+	protected onLabelKeypress(event: KeyboardEvent) {
+		const { disabled } = this.properties;
+		const key: string | number = 'key' in event ? event.key : event.keyCode;
+
+		if (!disabled && (key === 'Enter' || key === 13)) {
+			this.toggleDisplay();
+		}
+	}
```
smhigley: The menu should show when you hit the down, enter, or space keys. All of them should place focus on the first item in the sub menu when menu opens, and focus should return to the menu trigger before it closes.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,100 @@
...
+		if (!disabled && typeof onClick === 'function') {
+			onClick(event);
+		}
+	}
+
+	protected onKeypress(event: KeyboardEvent) {
+		const { disabled, onKeypress } = this.properties;
+		if (!disabled && typeof onKeypress === 'function') {
+			onKeypress(event);
+		}
```
smhigley: I think we should manage key events in a little more depth:
- up/down arrow keys should move focus to the previous or next menu item, and wrap beginning <--> end. 
- left/right should do the same on horizontal menus
- space/enter: activates the menu item and closes the dropdown menu (if applicable)
- space on a checkbox: toggles the checkbox but doesn't close the menu
- escape: closes the menu

src/menu/Menu.ts
```diff
@@ -0,0 +1,299 @@
...
+	disabled?: boolean;
+	expandOnClick?: boolean;
+	hideDelay?: number;
+	hidden?: boolean;
+	id?: string;
+	label?: DNode;
+	nested?: boolean;
+	onRequestHide?: () => void;
+	onRequestShow?: () => void;
+	role?: Role;
```
smhigley: We should probably also have an `orientation`  attribute that defaults to `horizontal`, at least for top-level menus.

src/menu/Menu.ts
```diff
@@ -0,0 +1,299 @@
...
+		if (label) {
+			return w(MenuItem, {
+				controls: id,
+				disabled,
+				expanded: !hidden,
+				hasMenu: true,
+				overrideClasses: overrideClasses || css,
+				onClick: this.onLabelClick,
+				onKeypress: this.onLabelKeypress
+			}, [ label ]);
```
smhigley: This should probably have an `id`, so the dropdown can refer back to it

src/menu/Menu.ts
```diff
@@ -0,0 +1,299 @@
...
+
+	render(): DNode {
+		const {
+			id = uuid(),
+			nested,
+			role = 'menu'
+		} = this.properties;
+
+		const label = this.renderLabel(id);
+		const menu = v('nav', {
```
smhigley: Let's also add an `aria-labelledby` that points back to the label, just for funsies.

src/menu/Menu.ts
smhigley: Oh, and all the `keypress` events should probably be `keydown` so they catch arrow keys

src/menu/Menu.ts
mwistrand: An `orientation` attribute makes sense, but given the generic use case for the `Menu` widget, do we really want to make it the default? @bitpshr?

src/menu/Menu.ts
```diff
@@ -0,0 +1,299 @@
...
+export class Menu extends MenuBase<MenuProperties> {
+	protected wasOpen = false;
+	private _hideTimer: Handle;
+	private _initialRender = true;
+
+	render(): DNode {
+		const {
+			id = uuid(),
+			nested,
+			role = 'menu'
```
mwistrand: Again, this depends on the default use case. If we are assuming that `Menu` will be used predominantly for main app navigation, then, yes, `menubar` should be the default.

smhigley: True. It should probably be tied to whatever determines whether or not we use `nav`, if we decide to control that.

src/menu/Menu.ts
smhigley: Yeah, I wasn't all that sure about `horizontal` either. I don't have strong opinions either way, I just suggested `horizontal` because I think that's the default browser assumption for menus.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,109 @@
...
+			type = 'item'
+		} = this.properties;
+
+		const classes = this.classes(
+			hasMenu ? css.menuLabel : css.menuItem,
+			disabled ? css.disabled : null,
+			selected ? css.selected : null
+		);
+
+		return v(tag, assign(Object.create(null), properties, {
```
tomdye: It is necessary to pass the `tag` through here, I don't believe we are doing this anywhere else.

tomdye: Is there any need for the `assign` call here over simple passing `properties` ?

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,109 @@
...
+			'aria-controls': controls,
+			'aria-expanded': String(expanded),
+			'aria-haspopup': hasPopup ? 'true' : undefined,
+			'aria-disabled': String(disabled),
+			classes,
+			key: 'menu-item',
```
tomdye: I think that the `key` should be set by the parent or should at least use the menu item `id` or there will be no difference.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,109 @@
...
+			onclick: this.onClick,
+			onkeydown: this.onKeydown,
+			role: roleMap[type],
+			tabIndex: active ? 0 : -1
+		}), this.children);
+	}
+
+	protected onClick(event: MouseEvent) {
```
tomdye: these are usually `private`

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	protected onElementUpdated(element: HTMLElement) {
+		this.properties.active && element.focus();
+	}
+
+	protected onKeydown(event: KeyboardEvent) {
+		const { disabled, onKeydown } = this.properties;
+		if (!disabled) {
+			if (event.keyCode === 32) { // space
```
tomdye: do you think it would be better to define a `const SPACE_KEY = 32` and use that here instead of needing a comment?

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,109 @@
...
+				(<HTMLElement> event.target).click();
+			}
+			typeof onKeydown === 'function' && onKeydown(event);
```
tomdye: The pattern elsewhere for consistency is `onKeyDown && onKeyDown(event)`.
Can `onKeydown` take any form other than undefined or a function? Also do you think `down` should be changed to have a capital `D`?

src/menu/styles/menu.m.css
```diff
@@ -0,0 +1,66 @@
+@import '../../common/styles/variables.css';
+
+.container {
```
tomdye: can you change this to `root` please?

src/menu/styles/menu.m.css
```diff
@@ -0,0 +1,66 @@
...
+
+.subMenu {
+	box-shadow: 0 0;
+	box-sizing: border-box;
+	overflow: hidden;
+	padding: 0 15px;
+	transition: height 0.5s ease-in-out;
+}
+
+.hidden { height: 0; }
```
tomdye: I think that these classes should be using the `base.css` classes. We should have a consistent approach to `show` / `hidden`

src/menu/styles/menu.m.css
```diff
@@ -0,0 +1,66 @@
...
+.nestedMenu {
+	width: 100%;
+}
+
+.subMenu {
+	box-shadow: 0 0;
+	box-sizing: border-box;
+	overflow: hidden;
+	padding: 0 15px;
+	transition: height 0.5s ease-in-out;
```
tomdye: Can you please use the `short animation time` variable here

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,109 @@
...
+			onkeydown: this.onKeydown,
+			role: roleMap[type],
+			tabIndex: active ? 0 : -1
+		}), this.children);
+	}
+
+	protected onClick(event: MouseEvent) {
+		const { disabled, onClick } = this.properties;
+		if (!disabled && typeof onClick === 'function') {
+			onClick(event);
```
tomdye: Would it be more advantageous to return the item id / key here instead of the click event? There's likely no need to expose the dom click event to the parent `Menu`.

src/menu/MenuItem.ts
mwistrand: I'll update to `onKeyDown` since that's the pattern used everywhere else.

src/menu/MenuItem.ts
mwistrand: Thanks for catching that. I thought I'd updated all of those.

src/menu/MenuItem.ts
mwistrand: I'd be fine with internally determining the tag name based on the specified `type` property. As for `assign`, it cleanly separates the properties that are specific to and required by `MenuItem` from those needed only by the underlying DOM node and have nothing to do with `Menu` functionality. Also, since other widgets (e.g., `Button`) use generic property names for the `aria-` properties, I tried to do the same thing here.

src/menu/MenuItem.ts
mwistrand: Correct; initially I included all the functionality for dropdowns within `Menu`, but after the widget logic became too complicated decided that it might be better to have a dedicated `DropDown` widget that extends `Menu`, in which case using `private` limits what can be done.

src/menu/MenuItem.ts
mwistrand: The `Menu` widget does not need access to the event, but there may be instances in which an application needs to call `event.preventDefault` on a `MenuItem` click event.

src/menu/styles/menu.m.css
mwistrand: Unfortunately `.hidden` in `base.css` uses `display: none`, which makes animation difficult; hence the need for the local classes. One option would be to change the class name to `collapsed`, but when applied to drop downs `hidden` feels more natural.

src/menu/MenuItem.ts
mwistrand: Hmm. Both `ComboBox` and `SlidePane` use a static key for controlling how `onElementCreated` behaves.

src/menu/Menu.ts
```diff
@@ -0,0 +1,435 @@
...
+		const menu = v('div', {
+			classes: this.classes.apply(this, this.getMenuClasses()),
+			id: this._id,
+			key: 'menu',
+			onclick: this.onItemActivate,
+			onfocus: this.onMenuFocus,
+			onfocusout: this.onMenuFocusOut,
+			onkeydown: this.onMenuKeyDown,
+			role,
+			tabIndex: this.state.active || label ? -1 : 0
```
smhigley: There are also some weird things happening with my ability to focus on the second menu after I've opened and closed its sub-menu. I think everything would be simplified by always having exactly one top-level menu item with `tabIndex: 0`, and directly moving focus as little as possible.

src/menu/example/index.ts
```diff
@@ -0,0 +1,205 @@
...
+				return w(MenuItem, {
+					key,
+					disabled: name === 'Open Location...',
+					onKeyDown: (event: KeyboardEvent) => {
+						if (event.keyCode === 13) {
+							toggleSelected();
+						}
+					},
+					onClick: toggleSelected,
+					selected: <boolean> this.state[`${key}Selected`]
```
smhigley: These seem like they're functioning as checkboxes with the `selected` property. Should they have `role: checkbox` even though the text seems to imply otherwise?

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+			hasMenu ? css.menuLabel : css.menuItem,
+			disabled ? css.disabled : null,
+			selected ? css.selected : null
+		);
+
+		const itemProperties: { [key: string]: any; } = {
+			'aria-controls': controls,
+			'aria-expanded': String(expanded),
+			'aria-haspopup': hasPopup ? 'true' : undefined,
+			'aria-disabled': String(disabled),
```
mwistrand: Hmm. The [WAI-ARIA example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-1/menubar-1.html) always shows `aria-expanded`.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+		if (key === 'root') {
+			this.properties.active && element.focus();
+		}
+	}
+
+	protected onKeyDown(event: KeyboardEvent) {
+		const { disabled, onKeyDown } = this.properties;
+		if (!disabled) {
+			if (event.keyCode === SPACE_KEY) {
+				(<HTMLElement> event.target).click();
```
mwistrand: If clicking a `MenuItem` just opens a link in a new window, for example, then calling `onClick` 1) requires the `MenuItem` to manually navigate to the underlying URL within its `onClick` handler and 2) potentially causes problems with popup blockers that are circumvented by just calling `click()`.

mwistrand: By "popup blocker", I mean built-in browser functionality.

src/menu/MenuItem.ts
smhigley: @mwistrand yeah, sorry, it should be `aria-expanded: hasPopup ? String(expanded) : null`

src/menu/example/index.ts
mwistrand: That's a good point. Since they are using `selected`, they should have `role: checkbox`.

src/menu/Menu.ts
```diff
@@ -0,0 +1,435 @@
...
+	}
+
+	protected getDefaultHidden() {
+		const { disabled, label } = this.properties;
+
+		if (label && disabled) {
+			return true;
+		}
+
+		return label ? true : false;
```
mwistrand: lol. You're right.

src/menu/Menu.ts
mwistrand: I see; so essentially all that changes is `tabIndex`? `document.activeElement` won't actually change during in-menu keyboard navigation?

mwistrand: Based on our private conversation, I can update `MenuItem` to store a reference to its DOM node so that focus can be moved out of `onElementUpdated` and changed in response to keyboard events. That way, this ~hack~ workaround can be removed.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+	protected onClick(event: MouseEvent) {
+		const { disabled, onClick } = this.properties;
+		if (!disabled && onClick) {
+			onClick(event);
+		}
+	}
+
+	protected onElementUpdated(element: HTMLElement, key: string) {
+		if (key === 'root') {
+			this.properties.active && element.focus();
```
mwistrand: `MenuItem`s now have both an `active` property for determining whether it should be focused and a `focusable` property for determining the `tabIndex`.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,118 @@
...
+		};
+
+		if (type === 'checkbox' || type === 'radio') {
+			itemProperties['aria-checked'] = String(selected);
+		}
+
+		return v(tag, assign(Object.create(null), properties, itemProperties), this.children);
+	}
+
+	protected onClick(event: MouseEvent) {
```
mwistrand: Fixed!

mwistrand:
Closing in favor of #104.

PR #84, opened 2017-03-23
Change deps to peer deps

agubler:
**Type:** bug

**Description:**

All dojo2 deps should be peer dependencies over deps as that is the current standard.


rishson:
This is awesome.

PR #94, opened 2017-04-03
Select 05

### LINKED ISSUE #5: Add Form Widgets

smhigley:
**Type:** bug / feature

**Description:**

Resolves #5, adds the last form widget

Adds a select widget with native and non-native options. In addition to the usual form properties, it has the following API:
```
multiple?: boolean;
options?: SelectOption[];
renderOption?(option: SelectOption): DNode;
useNativeSelect?: boolean;
onBlur?(event: FocusEvent): void;
onChange?(option: SelectOption): void;
onClick?(event: MouseEvent): void;
onFocus?(event: FocusEvent): void;
onKeyDown?(event: KeyboardEvent): void;
```

A sample native select could be rendered as follows:
```
const options = [
  { label: 'First', value: 'one' },
  { label: 'Second', value: 'two' }
];

w(Select, {
  options: options,
  useNativeSelect: true,
  value: this.state.selectValue,
  onChange: (option: SelectOption) => this.setState({ selectValue: option.value })
});
```

A non-native select with custom option elements would use `renderOption`:
```
w(Select, {
  options: options,
  renderOption: (option: SelectOption) => v('div', { classes: this.classes(customCss.testClass) }, [ option.label ]),
  value: this.state.selectValue,
  onChange: (option: SelectOption) => this.setState({ selectValue: option.value })
});
```


src/select/Select.ts
```diff
@@ -0,0 +1,411 @@
...
+import uuid from '@dojo/core/uuid';
+import Label, { LabelOptions } from '../label/Label';
+import * as css from './styles/select.m.css';
+
+/**
+ * @type SelectOption
+ *
+ * Properties that can be set on a Select component
+ *
+ * @property label				Text to display to the user
```
bitpshr: TypeDoc should be indented using spaces now for consistency.

src/select/Select.ts
```diff
@@ -0,0 +1,411 @@
...
+
+	private _onBlur (event: FocusEvent) { this.properties.onBlur && this.properties.onBlur(event); }
+	private _onClick (event: MouseEvent) { this.properties.onClick && this.properties.onClick(event); }
+	private _onFocus (event: FocusEvent) { this.properties.onFocus && this.properties.onFocus(event); }
+	private _onKeyDown (event: KeyboardEvent) { this.properties.onKeyDown && this.properties.onKeyDown(event); }
+
+	// native select events
+	private _onNativeChange (event: Event) {
+		const { options = [] } = this.properties;
+		const option = options.filter((option: SelectOption) => option.value === (<HTMLInputElement> event.target).value)[0];
```
bitpshr: I'd break `value` out into its own variable so you don't have to dig into `event` every iteration of `filter`. I'd also use `find` instead of `filter` since you're always using the first item that satisfies the filter condition:

```ts
const value = (<HTMLInputElement> event.target).value;
const option = options.find((option: SelectOption) => option.value === value);
```

src/select/Select.ts
```diff
@@ -0,0 +1,411 @@
...
+		this.properties.onChange && this.properties.onChange(option);
```
bitpshr: You can destructure `onChange`.

src/select/Select.ts
```diff
@@ -0,0 +1,411 @@
...
+
+		const optionNodes = [];
+		let optionNode, option;
+		for (let i = 0; i < options.length; i++) {
+			option = options[i];
+			option.id = option.id || uuid() + '';
+			option.selected = multiple ? option.selected : value === option.value;
+			optionNode = renderOption ? renderOption(option) : option.label;
+
+			optionNodes.push(v('div', {
```
bitpshr: Did you decide to not use the registry approach to allow custom option components?

src/select/Select.ts
```diff
@@ -0,0 +1,411 @@
...
+			invalid ? css.invalid : null,
+			invalid === false ? css.valid : null,
+			multiple ? css.multiselect : null,
+			readOnly ? css.readonly : null,
+			required ? css.required : null
+		];
+
+		let rootWidget;
+
+		const select = useNativeSelect ? this.renderNativeSelect() : multiple ? this.renderCustomMultiSelect() : this.renderCustomSelect();
```
bitpshr: Double ternary statements are hard to read.

src/select/Select.ts
```diff
@@ -0,0 +1,411 @@
...
+			value
+		} = this.properties;
+
+		const {
+			_open,
+			_focusedIndex,
+			_selectId
+		} = this;
+
+		const selectedOption = options.filter((option: SelectOption) => option.value === value)[0] || options[0];
```
bitpshr: `Array.first` instead of `Array.filter`

src/select/Select.ts
```diff
@@ -0,0 +1,411 @@
...
+			case keys.enter:
+				!options[_focusedIndex].disabled && onChange && onChange(options[_focusedIndex]);
+				break;
+			case keys.space:
+				!options[_focusedIndex].disabled && onChange && onChange(options[_focusedIndex]);
+				break;
+			case keys.escape:
+				this._closeSelect();
+				break;
+			case keys.down:
```
bitpshr: Can a `disabled` option be focused using the keyboard? I didn't see any logic to find the next enabled option if the currently-focused option is disabled. Curious because I had to do something like that for TabPane.

src/select/styles/example.m.css
```diff
@@ -0,0 +1,28 @@
+@import '../../common/styles/variables.css';
+
+.option {
```
bitpshr: I think most of the example page CSS can go away. Tom is pulling out theme styles from structural styles at dojo/widgets#77. These aesthetic styles could go into the theme CSS, and any minor page styling could just be done inline. I think we should try to keep `styles` clean of example-specific stuff.

smhigley: Oops, I actually put the example styling inline, and forgot to delete this :P

src/select/Select.ts
smhigley: Yeah, in the end it didn't seem necessary, and I didn't really want to split out custom option nodes into a separate widget. I'm open to hearing reasons to do so, though

src/select/Select.ts
smhigley: Yes, disabled options should be (and are) focusable using the keyboard: https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_disabled_controls

src/select/Select.ts
smhigley: Quick question -- IE supports `Array.filter` but not `Array.find`. Should I still use find, and is there a dojo helper or something I should use to get IE11 support? (I didn't see anything in `dojo/core/lang` or `dojo/core/util`)

bitpshr: We have a shim in `@dojo/shim/array`.

src/select/Select.ts
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	// custom select events
+	private _openSelect() {
+		this._open = true;
+		this._ignoreBlur = false;
+		this._focusedIndex = this._focusedIndex || 0;
+		this.invalidate();
+	}
+	private _closeSelect() {
```
bitpshr: New line

src/select/Select.ts
```diff
@@ -0,0 +1,450 @@
...
+			name,
+			options = [],
+			readOnly,
+			required,
+			value
+		} = this.properties;
+
+		/* create option nodes */
+		const optionNodes = [];
+		for (let option of options) {
```
bitpshr: `map` may be cleaner

smhigley:
Waiting for https://github.com/dojo/widget-core/issues/442 to fix test issue causing build failure.

src/select/Select.ts
```diff
@@ -0,0 +1,419 @@
...
+	value?: string;
+	onBlur?(event: FocusEvent): void;
+	onChange?(option: OptionData): void;
+	onClick?(event: MouseEvent): void;
+	onFocus?(event: FocusEvent): void;
+	onKeyDown?(event: KeyboardEvent): void;
+}
+
+// This should have a lookup somewhere
+const keys = {
```
bitpshr: I wonder if it's time to pull this out into `common/keys` or something similar.

src/select/Select.ts
bitpshr: I need to update ComboBox in that case. Thanks.

src/select/Select.ts
```diff
@@ -0,0 +1,419 @@
...
+		const {
+			describedBy,
+			disabled,
+			invalid,
+			options = [],
+			readOnly,
+			required
+		} = this.properties;
+
+		const {
```
bitpshr: I think you can combine this destructuring:

```ts
const {
	_focusedIndex,
	properties: {
		describedBy,
		disabled,
		invalid,
		options = [],
		readOnly,
		required
	}
} = this;
```

smhigley: Ohhh, nice! I didn't even think about nested destructuring.

src/select/Select.ts
smhigley: Probably. I think it might be better to do it in a separate PR, though, and change it in ComboBox, Tabs, and wherever else as well.

src/select/Select.ts
smhigley: Cool 👍. It's not universal (i.e. not every single disabled focusable thing should continue to get focus), but it's one of the exceptions outlined in the spec. I think Combobox could probably have focusable disabled options too, though :)

rishson:
@smhigley do you want to change `useNativeSelect` to `useNativeElement` in this PR?

src/select/Select.ts
```diff
@@ -0,0 +1,418 @@
...
+			v('span', {
+				classes: this.classes(css.arrow)
+			})
+		]);
+	}
+
+	renderCustomMultiSelect(): DNode {
+		const {
+			_focusedIndex,
+			properties = {
```
rishson: causing tslint 'unused variable' error.

PR #104, opened 2017-04-10
`MenuItem`, `Menu`, `SubMenu`, and `MenuBar` widgets

### LINKED ISSUE #52: Menu widget

mwistrand:
**Type:** feature

**Description:**

Continues the work in #79, simplifying the `Menu` widget by splitting it into two widgets: `Menu` for managing just the list of menu items and their interactions, and `SubMenu` (admittedly not the best name; I'm very open to suggestions) for rendering a trigger that toggles a menu, whether that be an inlined menu, a drop down, or a popup. Also includes a simple `MenuBar` widget that takes a numerical breakpoint, beneath which all children are rendered to a `SlidePane`.

Resolves #52


src/menu/Menu.ts
```diff
@@ -0,0 +1,212 @@
...
+export type Orientation = 'horizontal' | 'vertical';
+
+export type Role = 'menu' | 'menubar';
+
+/**
+ * @type MenuProperties
+ *
+ * Properties that can be set on a Menu component.
+ *
+ * @property active				Determines whether the menu has focus.
```
bitpshr: We've been using spaces for TypeDoc comments.

src/menu/Menu.ts
```diff
@@ -0,0 +1,212 @@
...
+ * @property activeIndex		Determines the index of the focused item.
+ * @property disabled			Determines whether the menu is disabled.
+ * @property id					The widget ID. Defaults to a random string.
+ * @property nested				Indicates whether the menu is nested within another menu.
```
bitpshr: Would a nested menu not be a SubMenu? It looks like this is only used for styling.

src/menu/Menu.ts
```diff
@@ -0,0 +1,212 @@
...
+
+		if (menuId === id && !isNaN(index)) {
+			this._activeIndex = index;
+		}
+	}
+
+	private _renderChildren() {
+		const { activeIndex = this._activeIndex, id = this._id } = this.properties;
+		const focusIndex = this._isActive() ? activeIndex : undefined;
+		this._activeIndex = activeIndex;
```
bitpshr: Let's say a parent component passes in an `activeIndex` then the active element is changed via the arrow keys. Wouldn't the new index immediately be overwritten since the parent doesn't have any way to know to clear out the `activeIndex` property it was passing in?

src/menu/Menu.ts
```diff
@@ -0,0 +1,212 @@
...
+		let itemNode = <HTMLElement> event.target;
+		while (!itemNode.hasAttribute('data-dojo-index') && itemNode.parentElement) {
+			itemNode = itemNode.parentElement;
+		}
+
+		const { id = this._id } = this.properties;
+		const index = parseInt(itemNode.getAttribute('data-dojo-index') || '', 10);
+		const menuId = itemNode.getAttribute('data-dojo-menuid');
+
+		if (menuId === id && !isNaN(index)) {
```
bitpshr: I wonder if it would be cleaner to put a mouse down handler on each child menu item. That way you can be sure any item triggering this handler is a child of this menu, so these checks could go away.

src/menu/MenuBar.ts
```diff
@@ -0,0 +1,96 @@
...
+			onRequestClose,
+			open,
+			slidePaneButtonLabel,
+			slidePaneStyles
+		} = this.properties;
+
+		return v('div', {
+			classes: this.classes(css.navBar, css.slidePane)
+		}, [
+			v('button', {
```
bitpshr: Part of me feels like we should accept any arbitrary DNode and use that as the trigger for the SlidePane instead of always using a button.

src/menu/MenuBar.ts
```diff
@@ -0,0 +1,96 @@
...
+		window.removeEventListener('resize', listener);
+	};
+});
+
+const MenuBarBase = ThemeableMixin(WidgetBase);
+
+@theme(css)
+export class MenuBar extends MenuBarBase<MenuBarProperties> {
+	constructor() {
+		/* istanbul ignore next: disregard transpiled `super`'s "else" block */
```
bitpshr: Does this fix the coverage?! Amazing find.

src/menu/MenuBar.ts
```diff
@@ -0,0 +1,96 @@
...
+	breakpoint: number;
+	slidePaneButtonLabel?: string;
+	slidePaneStyles?: any;
+	onRequestClose?: () => void;
+	onRequestOpen?: () => void;
+	open?: boolean;
+}
+
+// TODO: Make this a hot observable with Evented?
+// TODO: Should this be moved into a common utility method that can be used by any widget?
```
bitpshr: Could you remove these? Also, I do think that this should go into a utility method in `common`, this is nice and useful.

src/menu/MenuBar.ts
```diff
@@ -0,0 +1,96 @@
...
+import { v, w } from '@dojo/widget-core/d';
+import ThemeableMixin, { theme, ThemeableProperties } from '@dojo/widget-core/mixins/Themeable';
+import WidgetBase from '@dojo/widget-core/WidgetBase';
+import * as css from './styles/menuBar.m.css';
+import SlidePane from '../slidepane/SlidePane';
+
+export interface MenuBarProperties extends ThemeableProperties {
+	breakpoint: number;
+	slidePaneButtonLabel?: string;
+	slidePaneStyles?: any;
```
bitpshr: Smart.

src/menu/MenuBar.ts
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	private _onSlidePaneClick() {
+		this.properties.onRequestOpen && this.properties.onRequestOpen();
+	}
+
+	private _observeViewport() {
+		// TODO: If the current vw is stored as a private variable, then it would
+		// be possible to invalidate only when the vw crosses the breakpoint.
+		const viewportSubscription = viewportSource.subscribe({
```
bitpshr: I agree with your TODO here, it would be good to conditionally invalidate.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,176 @@
...
+	properties?: VirtualDomProperties;
+	selected?: boolean;
+	tag?: string;
+	type?: MenuItemType;
+}
+
+export const MenuItemBase = ThemeableMixin(WidgetBase);
+
+const SPACE_KEY = 32;
+const ENTER_KEY = 13;
```
bitpshr: We should really pull these out into a common module. I think this, TabPane, ComboBox, and a few others use internal key maps, so they can all be updated as well. You don't have to do that as part of this PR, but I won't stop you :)

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,176 @@
...
+	expanded?: boolean;
+	focusable?: boolean;
+	hasMenu?: boolean;
+	hasPopup?: boolean;
+	id?: string;
+	index?: number;
+	menuId?: string;
+	onClick?: (event: Event) => void;
+	onKeyDown?: (event: KeyboardEvent) => void;
+	properties?: VirtualDomProperties;
```
bitpshr: What was the motivation for this properties bag?

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,176 @@
...
+					// within the next task, whereas others may take multiple tasks before the element switches
+					// to visible.
+					let callCount = 0;
+					const scheduleFocus = () => {
+						requestAnimationFrame(() => {
+							visibility = getComputedStyle(element).getPropertyValue('visibility');
+
+							if (this.properties.active) {
+								if (visibility === 'hidden') {
+									if (callCount > 30) {
```
bitpshr: A little confused by this. What's the purpose of the `callCount` check here?

src/menu/SubMenu.ts
```diff
@@ -0,0 +1,445 @@
...
+	focusable?: boolean;
+	hidden?: boolean;
+	hideOnBlur?: boolean;
+	hideDelay?: number;
+	hideOnActivate?: boolean;
+	index?: number;
+	label: DNode;
+	labelId?: string;
+	labelStyles?: any;
+	menuStyles?: any;
```
bitpshr: This is a little confusing. It's not obvious that a SubMenu creates a Menu internally, so a developer could naively pass `overrideClasses` to the SubMenu thinking they would theme it properly. Maybe you could pass `this.properties.overrideClasses` to the Menu that's created and lose `menuStyles` altogether?

src/menu/Menu.ts
mwistrand: It is only used for styling, and since `SubMenu` renders its menu to a container node it can control when animating, and since there are other classes for positioning, this is probably unnecessary. I'll remove it.

src/menu/Menu.ts
mwistrand: I added the `menuId` property to prevent menus from responding to event within nested menus (for example, clicking an item within a dropdown menu should not cause an ancestor `MenuBar` to update it's own index). We could deliberately add a `child.properties.onMouseDown` to menu items from within `_renderChildren`, which might solve this problem.

src/menu/Menu.ts
mwistrand: The parent would need a means of determining whether the `activeIndex` should be set. For example, `SubMenu` uses `activeIndex: this.properties.hidden ? 0 : undefined`, so control is handed over to `Menu` as soon as it is visible. In practice, parents should avoid setting `activeIndex` as much as possible, and it was added specifically so that `SubMenu` would have a way of indicating to the child menu that keyboard navigation should start from the first item after closing the menu.

@smhigley and I were discussing "internal use" properties that naturally arise from creating distinct components that are intended to work with each other, and how it would be nice if we had a natural way of marking them as such. One option is to use the format `__propertyName__` as is done with the `__render__` method, but I can open an issue for discussion.

src/menu/MenuBar.ts
mwistrand: It does, and I'm fine using it here since Istanbul is catching on an `else` block in the transpiled code that we don't need to care about. Once SitePen/remap-istanbul#106 is resolved, these checks can be removed.

src/menu/MenuItem.ts
mwistrand: Agreed. I'll add a `Keys` map and the vw observer to a `common/util.ts` module, and we can expand on that/separate functionality later if the need arises.

src/menu/MenuBar.ts
agubler: neat.

src/menu/MenuItem.ts
mwistrand: It's just a safety guard against infinitely checking whether a hidden `MenuItem` has become visible, since the `MenuItem` does not control its own visibility.

src/menu/SubMenu.ts
mwistrand: Since `Menu` can be used independently of `SubMenu`, I separated the styles into two different CSS files. If we need to merge the two, or if there is a way to clarify the current implementation, I'll make the necessary changes.

src/menu/MenuItem.ts
mwistrand: Since the user can specify the tag used for the item's DOM node, it makes sense to also allow properties specific to that tag to be used (e.g., `href` with `<a>`). It also keeps those properties separate from those required by `MenuItem` to function properly.

src/menu/MenuBar.ts
mwistrand: Since the trigger needs to open/close the slide pane on click, how about a function property that takes the click handler and returns a `DNode`?
`slidePaneTrigger?: string | ((onClick: () => void) => DNode)`

src/menu/Menu.ts
bitpshr: Ok, makes sense.

src/menu/MenuItem.ts
bitpshr: Got it. Maybe pull that out into a constant.

src/menu/SubMenu.ts
bitpshr: I vote to merge them so that `SubMenu` could pass its `overrideClasses` straight on to `Menu`, but it's your call.

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,186 @@
...
+ * @property hasPopup       Indicates whether the widget has a drop down child.
+ * @property id             Specifies the ID for the widget.
+ * @property index          Specifies the index of the item within a parent menu.
+ * @property onClick        Called when the widget is activated via a click or space key.
+ * @property onKeyDown      Called when keys are pressed while the widget has focus.
+ * @property onMouseDown    Called during a mousedown event.
+ * @property properties     Additional properties for the widget's vnode.
+ * @property selected       Determines whether the widget is marked as selected.
+ * @property tag            The HTML tag name to use for the widget's vnode. Defaults to 'a'.
+ * @property type           The type of the menu item. Defaults to 'item'.
```
bitpshr: Should this be `role` to stay consistent?

src/menu/MenuItem.ts
```diff
@@ -0,0 +1,186 @@
...
+		if (key === 'root') {
+			let visibility = getComputedStyle(element).getPropertyValue('visibility');
+
+			if (this.properties.active) {
+				if (visibility === 'hidden') {
+					// When animating from `visibility: hidden`, some browsers will mark the element as visible
+					// within the next task, whereas others may take multiple tasks before the element switches
+					// to visible.
+					let callCount = 0;
+					const scheduleFocus = () => {
```
bitpshr: Why did you have to wrap the rAF call in an anonymous function?

src/menu/Menu.ts
```diff
@@ -0,0 +1,186 @@
...
+
+export const enum Operation {
+	decrease,
+	increase
+}
+
+export const MenuBase = ThemeableMixin(WidgetBase);
+
+@theme(css)
+export class Menu extends MenuBase<MenuProperties> {
```
bitpshr: We should probably have a custom element descriptor for all of these components.

src/common/util.ts
```diff
@@ -0,0 +1,32 @@
+import Observable, { Observer, Subscription } from '@dojo/shim/Observable';
+
+export const enum Keys {
```
mwistrand: Yes. I just saw that PR had landed.

src/menu/MenuItem.ts
mwistrand: It was `role` originally, but @smhigley and I agreed that allowing users to specify `item`, `checkbox`, or `radio` instead of `menuitem`, `menuitemcheckbox` or `menuitemradio` would be better, in which case `role` became somewhat misleading.

src/menu/MenuItem.ts
mwistrand: Since rAF may need to be called multiple times before the element becomes visible in certain browsers, depending on the animation.

src/menu/Menu.ts
```diff
@@ -0,0 +1,186 @@
...
+import { v } from '@dojo/widget-core/d';
+import { DNode } from '@dojo/widget-core/interfaces';
+import ThemeableMixin, { theme, ThemeableProperties } from '@dojo/widget-core/mixins/Themeable';
+import WidgetBase from '@dojo/widget-core/WidgetBase';
+import * as css from './styles/menu.m.css';
+import { Keys } from '../common/util';
+
+export type Orientation = 'horizontal' | 'vertical';
+
+export type Role = 'menu' | 'menubar';
```
mwistrand: `MenuRole` or `RoleType`?

bitpshr: Ah, yeah. I guess `RoleType`.

src/menu/Menu.ts
```diff
@@ -0,0 +1,186 @@
...
+		super();
+		// TODO: Remove once focus management is implemented.
+		this.own(createHandle(() => {
+			this._domNode = null;
+		}));
+	}
+
+	render(): DNode {
+		const {
+			id = this._id,
```
tomdye: seems off to me that this widget will create and store `this._id` when it may never be used and has no bearing to the rendered ui if an id is specified.

src/menu/Menu.ts
```diff
@@ -0,0 +1,186 @@
...
+			onfocusin: this._onMenuFocus,
+			onfocusout: this._onMenuFocusOut,
+			onkeydown: this._onMenuKeyDown,
+			role
+		}, this._renderChildren());
+	}
+
+	protected onElementCreated(element: HTMLElement, key: string) {
+		if (key === 'root') {
+			this._domNode = element;
```
tomdye: Why do you need to store a reference to the rendered element?

src/menu/Menu.ts
```diff
@@ -0,0 +1,186 @@
...
+
+		if (orientation === 'horizontal') {
+			classes.push(css.horizontal);
+		}
+
+		return classes;
+	}
+
+	private _isActive() {
+		return this._active || this.properties.active;
```
tomdye: should the `active` state not always come from passed in properties?

src/menu/Menu.ts
```diff
@@ -0,0 +1,186 @@
...
+		const previousIndex = this._activeIndex;
+		this._activeIndex = operation === Operation.decrease ?
+			previousIndex - 1 < 0 ? max - 1 : previousIndex - 1 :
+			Math.min(previousIndex + 1, max) % max;
+
+		this.invalidate();
+	}
+
+	private _onMenuFocus() {
+		if (!this._isActive()) {
```
tomdye: in a controlled pattern I would have expected to see a `onRequestFocus` function called here such that the parent component can deal with focus / active state.

src/menu/Menu.ts
```diff
@@ -0,0 +1,186 @@
...
+			this._active = true;
+			this.invalidate();
+		}
+	}
+
+	private _onMenuFocusOut() {
+		if (this._isActive()) {
+			requestAnimationFrame(() => {
```
tomdye: why do you need to requestanimationframe here? widget renders always occur on the subsequent animation frame

src/menu/Menu.ts
```diff
@@ -0,0 +1,186 @@
...
+			this._activeIndex = index;
+		}
+	}
+
+	private _renderChildren() {
+		const { activeIndex = this._activeIndex } = this.properties;
+		const focusIndex = this._isActive() ? activeIndex : undefined;
+		this._activeIndex = activeIndex;
+
+		this.children.filter((child: any) => child && child.properties)
```
tomdye: Are children rendered inside a `<nav>` element at any point?

src/menu/MenuBar.ts
```diff
@@ -0,0 +1,105 @@
+import { createHandle } from '@dojo/core/lang';
+import { debounce } from '@dojo/core/util';
+import { v, w } from '@dojo/widget-core/d';
+import { DNode } from '@dojo/widget-core/interfaces';
+import ThemeableMixin, { theme, ThemeableProperties } from '@dojo/widget-core/mixins/Themeable';
+import WidgetBase from '@dojo/widget-core/WidgetBase';
+import * as css from './styles/menuBar.m.css';
+import SlidePane from '../slidepane/SlidePane';
+import { observeViewport } from '../common/util';
```
tomdye: Why does the widget need to observe the viewport? The viewport should be passed via properties.

src/menu/MenuBar.ts
```diff
@@ -0,0 +1,105 @@
...
+	onRequestClose?: () => void;
+	onRequestOpen?: () => void;
+	open?: boolean;
+}
+
+const MenuBarBase = ThemeableMixin(WidgetBase);
+
+@theme(css)
+export class MenuBar extends MenuBarBase<MenuBarProperties> {
+	private _vw = document.body.offsetWidth;
```
tomdye: what is `_vw` ?

tomdye:
I don't understand why we have created such a complicated header / sidebar / menu thing that tries to be clever about it's surroundings and lay itself out appropriately when those kind of things should be determined at app level. In my opinion the `menu` should just be a `<nav>` with links / list items that can be configured via the registry. There should be a separate widget created for `Header` and we already have the `SlidePane` and at the app level decisions should be made as to if a header or a slidepane are created to hold the menu.

src/menu/Menu.ts
mwistrand: No, parents will need to render to a `<nav>` if needed since not every menu warrants a `<nav>` (e.g., dropdown menus within a parent `<nav>`).

src/menu/Menu.ts
mwistrand: It's so the widget has a way of knowing whether the current `document.activeElement` is a descendant of the widget node. When focus management is added, this will go away.

src/menu/Menu.ts
mwistrand: When clicking within a menu, `document.activeElement` is equivalent to `document.body` until the next frame, making the `node.contains(document.activeElement)` check useless without the rAF.

mwistrand:
@tomdye A good portion of the [issue](https://github.com/dojo/widgets/issues/52) discusses a menu bar that responds to viewport resizing by rendering contents in a `SlidePane` (and I wouldn't 
 call `MenuBar` complicated). And the separate `Menu` widget is just a list of `MenuItem`s, but it also has to manage things like menu item navigation, which is more complicated than it is with, for example, a select or combobox since the menu items are tangible elements that need to receive focus. Not to mention that "menu" is such a vague term that can be applied to a number of UI components, so the goal here was to keep `Menu` generic enough to fit those different use cases.

src/menu/Menu.ts
rishson: Will direct RAF usage go away if we implement some kind of focus trapping, in the same way that references to the rendered element will disappear?

mwistrand: As long as the focus trap jumps through any hoops it needs to track changes to `document.activeElement`, then widgets shouldn't need to use rAF themselves to determine whether they still have focus.

mwistrand:
@rishson Yes, it definitely is closer to the DOM than anyone would like, but to change that we'd need a way to:

1) abstract programmatic focus-setting (which neither maquette nor `widget-core` currently supports).
2) animate to/from `height: auto`, which is unfortunately not possible without calculating dimensions.
3) abstract viewport observation. If everyone decides `MenuBar` should be removed in favor requiring applications to handle the same functionality, I would be okay with that (although the utility method in `src/common/util` is still useful even if gets moved somewhere else like `widget-core`).

tomdye:
I strongly feel that `observeViewport` needs to be removed from menu bar and the `orientation` should become a property that can be set.

tomdye:
Can you ensure that the `theme` property is passed into and child widgets please?

src/common/tests/unit/util.ts
```diff
@@ -18,5 +29,23 @@ registerSuite({
...
+	},
+
+	observeViewport() {
+		let next = sinon.spy();
+		const subscriber = observeViewport({ next });
+		const listener: () => void = (<any> window.addEventListener).args[0][1];
+		listener();
+
+		assert.isTrue((<any> window.addEventListener).calledWith('resize'));
+		assert.isNumber(next.args[0][0], '`next should be passed the current viewport width.`');
```
tomdye: Please use `next.firstCall.calledWith` where possible

src/menu/Menu.ts
```diff
@@ -0,0 +1,186 @@
+import { createHandle } from '@dojo/core/lang';
+import uuid from '@dojo/core/uuid';
+import { v } from '@dojo/widget-core/d';
+import { DNode } from '@dojo/widget-core/interfaces';
+import ThemeableMixin, { theme, ThemeableProperties } from '@dojo/widget-core/mixins/Themeable';
+import WidgetBase from '@dojo/widget-core/WidgetBase';
+import * as css from './styles/menu.m.css';
+import { Keys } from '../common/util';
+
+export type Orientation = 'horizontal' | 'vertical';
```
tomdye: This would work well as an ENUM

src/menu/styles/subMenu.m.css
```diff
@@ -0,0 +1,57 @@
...
+}
+
+.dropDown.hidden,
+.popup.hidden {
+	opacity: 0;
+	pointer-events: none;
+}
+
+/* Animations */
+.fade {
```
tomdye: should be in the animations.css file

mwistrand:
@tomdye Without the viewport observation, is it even worth having a menu bar widget? In my opinion, it's simple enough that the only benefit it provides is resetting the wrapper widget in response to viewport changes. And since there will be only at most one or two menu bars in an application, I see no reason why the application couldn't handle that itself.

bitpshr:
What's the status of this PR?

mwistrand:
@bitpshr That's a good question. It seems the biggest complaint is how low-level the menu functionality is, and that we need abstractions for animation and focus management. But beyond that I am not sure how others want this to change before giving the go-ahead to land it while we wait for those abstractions to be developed.

kitsonk:
Let either update and land this, or close it.

tomdye:
Closing this for now as it has been stagnant for some time. Can be used as a reference in future when working on a Menu widget.

PR #114, opened 2017-04-17
Util for key codes

smhigley:
**Type:** feature

**Description:**

Moves key codes to `common/util`, and updates all widgets that use them (Select, ComboBox, and TabPane).

### DIFF
--- a/src/combobox/ComboBox.ts
+++ b/src/combobox/ComboBox.ts
@@ -9,6 +9,7 @@ import ResultMenu, { ResultMenuProperties } from './ResultMenu';
 import TextInput, { TextInputProperties } from '../textinput/TextInput';
 import uuid from '@dojo/core/uuid';
 import WidgetRegistry from '@dojo/widget-core/WidgetRegistry';
+import { Keys } from '../common/util';
 
 import * as css from './styles/comboBox.m.css';
 
@@ -67,14 +68,6 @@ export interface ComboBoxProperties extends ThemeableProperties {
 export const enum Operation {
 	increase = 1,
 	decrease = -1
-};
-
-// This can go away when Safari decides to support KeyboardEvent.key
-const keys: {[key: string]: number} = {
-	escape: 27,
-	enter: 13,
-	up: 38,
-	down: 40
 };
 
 export const ComboBoxBase = ThemeableMixin(WidgetBase);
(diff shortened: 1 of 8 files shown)


src/combobox/ComboBox.ts
```diff
@@ -169,19 +162,19 @@ export default class ComboBox extends ComboBoxBase<ComboBoxProperties> {
 	private _onInputKeyDown(event: KeyboardEvent) {
 		const { results = [] } = this.properties;
 
-		switch (event.keyCode) {
-			case keys.up:
+		switch (event.which) {
```
agubler: Should you change this as part of this PR?

src/tabpane/TabButton.ts
```diff
@@ -85,33 +86,33 @@ export default class TabButton extends TabButtonBase<TabButtonProperties> {
 		}
 
 		// Accessibility
-		switch (event.keyCode) {
+		switch (event.which) {
```
agubler: Same question.

src/combobox/ComboBox.ts
smhigley: I thought it would be good to standardize use of `event.keyCode` or `event.which`, since we currently use both. Should we bother/should it be in a different PR?

msssk:
I think Dojo 2 should, [like Dojo 1](https://github.com/dojo/dojo/blob/master/keys.js), provide support for a decent subset (if not the entirety) of the [W3 spec for key values](https://www.w3.org/TR/DOM-Level-3-Events-key/#key-value-tables).

This PR seems like a small stop-gap measure. It's fine that very few keys are supported since that can be updated later with no impact, but I think we should decide where the `keys` module will reside sooner than later to reduce duplication and refactoring.

[`dojo/core` issue](https://github.com/dojo/core/issues/322) (would this be better in `dojo/meta`?)

src/combobox/ComboBox.ts
msssk: `which` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which)

smhigley: so is `keyCode`, but `key` doesn't have browser support.

smhigley: `event.which` normalizes `event.keyCode` and `event.charCode` which doesn't matter much for these keys, but if we're choosing a standard I'd rather go with `which`.

Safari supports `event.key` as of about a week ago, but the Android browser does not. I forget if we support it, but if not I'd be happy to switch to `.key` and drop this entire util.

msssk: Sounds like `which`, although being the most poorly named, is the best option to go with.

Regardless, we do still want a `keys` module, right? Even if we're able to go with `event.key` it will be nice to be able to do comparisons against `keys.ArrowDown` etc.

smhigley: If we could use `event.key` I think it would make sense to just compare against `'ArrowDown'` vs. `keys.ArrowDown`. Might be moot though, unless we can ignore the Android browser and Safari 10.0/iOS Safari <10.3.

smhigley:
@msssk Yeah, it's definitely a stop-gap, and not necessarily a permanent solution. I think it'll eventually be moved out of `@dojo/widgets`, but there wasn't really agreement on which repo it should be in. It'd just be nice to use it in current widgets before too many more are created, and I hope this change will make it easier to switch approaches in the future (i.e. the only change will be the import, not all the references to `keys` in the code).

agubler:
@bitpshr because it's an unrelated change, which we should try and avoid.

PR #115, opened 2017-04-17
Aria cleanup

smhigley:
**Type:** bug

**Description:**

- Updates widgets that use `aria-invalid` to only show the property when it is set to `true`. Previously, widgets were displaying `aria-invalid="undefined"`.
- Also updates Select option id's to be correctly referenced in `aria-activedescendant`.

### DIFF
--- a/src/select/Select.ts
+++ b/src/select/Select.ts
@@ -242,9 +242,9 @@ export default class Select extends SelectBase<SelectProperties> {
 
 		// add ids to options for use with aria-activedescendant
 		if (includes(evt.changedPropertyKeys, 'options')) {
-			for (let option of options) {
+			options.forEach((option) => {
 				option.id = option.id || uuid();
-			}
+			});
 		}
 	}
 
--- a/src/select/Select.ts
+++ b/src/select/Select.ts
@@ -218,7 +218,7 @@ export default class Select extends SelectBase<SelectProperties> {
 			index: i,
 			key: i + '',
 			optionData: assign({}, option, <any> {
-				id: option.id || uuid() + '',
+				id: option.id,
 				selected: multiple ? option.selected : value === option.value
 			}),
 			onMouseDown: this._onOptionMouseDown,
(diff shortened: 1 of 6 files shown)


src/select/Select.ts
```diff
@@ -231,12 +231,21 @@ export default class Select extends SelectBase<SelectProperties> {
...
 
+		// update custom option registry
 		if ( !this.registry || includes(evt.changedPropertyKeys, 'customOption')) {
 			this.registry = this._createRegistry(customOption);
 		}
+
+		// add ids to options for use with aria-activedescendant
+		if (includes(evt.changedPropertyKeys, 'options')) {
+			for (let option of options) {
+				option.id = option.id || uuid() + '';
```
agubler: Curious why `+ ''`?

smhigley: hah, I don't remember :D. I think I thought it was a number, but I clearly didn't cast it earlier, so... ¯\_(ツ)_/¯

src/select/Select.ts
```diff
@@ -231,12 +231,21 @@ export default class Select extends SelectBase<SelectProperties> {
...
 		} = this.properties;
 
+		// update custom option registry
 		if ( !this.registry || includes(evt.changedPropertyKeys, 'customOption')) {
 			this.registry = this._createRegistry(customOption);
 		}
+
+		// add ids to options for use with aria-activedescendant
+		if (includes(evt.changedPropertyKeys, 'options')) {
+			for (let option of options) {
```
dylans: should we be using native for...of or the [forOf shim](https://github.com/dojo/shim/blob/master/src/native/iterator.ts#L66)

smhigley: I forgot about IE11 :P. I think I'd rather use a regular for loop over a shim.

dylans: sorry to spoil your ES6 zen ;)

agubler: Curious why not a `forEach`?

smhigley: `forEach` is fine, would you prefer that?

agubler: We do generally use the newer array constructs, I was just wondering if there was a reason not to.

Issue #117, opened 2017-04-18
Common Select/dropdown base widget


smhigley:
**Enhancement**

The dropdowns in Select, ComboBox, and TimePicker are similar enough that I think we should discuss using a common base for all three, possibly by splitting Select into three separate widgets and using custom Select as a generic dropdown-selection widget.

Benefits:
- Standardized keyboard interaction, controlled and updated in one place (e.g. adding home/end keys would be easy)
- Standard approach to focusing (either moving the focus directly on the options or using `aria-activedescendant`)
- Styling and theming is in one place
- Easier to add features like position detection for rendering the menu above or below the menu trigger

dylans:
I certainly like not having multiple implementations of almost the same thing, and dealing with the inconsistency that introduces (see Dijit's ComboBox, Select, and FilteringSelect for this problem in action).

sebilasse:
👍 
this is just a cc. for https://github.com/Semantic-Org/Semantic-UI/issues/5420 because it's Dropdown could inherit from this base class.

PR #118, opened 2017-04-18
TimePicker widget

### LINKED ISSUE #100: TimePicker widget
# Specification
This widget provides time picker behaviour.
 
screen shot 2017-04-06 at 2 33 56 pm

## Features
- can support AM/PM labels
- can support 12 and 24hr times
- can be passed a value that sets the currently displayed time

## Features excluded from this version
- timezone support, e.g. optional picker after the time [16:20][PM]**[BST]**
- default to current time if no time passed in
- falls back to native mobile UI on mobile

Ability yo choose times

`onHide`
`onShow`

Ability to scroll through the times when choosing, using mouse and keyboard.

Mandatory
Disabled
Enabled

Controlled

**List of any icons needed**
Clock icon for input addon (used to launch the time picker)

Look and feel of the widget.

**Any other considerations**:

## Questions
What format should be used to get/set the time data?

## Acceptance criteria
When I click the 'clock' icon on the input, the timepicker should be shown in a direction that makes best use of screen real estate, e.g. if the input is at the bottom of the screen, then the timepicker should be shown above the input
When I set and invalid time, e.g. `25:10`, then the input will show as `invalid`.

mwistrand:
**Type:** feature

**Description:**

Implement a time picker widget that renders a text input, a button, and a menu of possible times within a specified range. Borrows heavily from `ComboBox`.

Resolves #100


src/timepicker/styles/timePicker.m.css
```diff
@@ -0,0 +1,77 @@
...
+	display: inline-block;
+	position: relative;
+	width: var(--form-field-width);
+}
+
+.controls {
+	position: relative;
+}
+
+.disabled {}
```
bitpshr: Ah, I need to do this for ComboBox so theming can work properly. Nice catch.

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+		return w(TextInput, {
+			...inputProperties,
+			bind: this,
+			disabled,
+			invalid,
+			onBlur: this._onInputBlur,
+			onFocus: this._onInputFocus,
+			onInput: this._onInput,
+			onKeyDown: this._onInputKeyDown,
+			overrideClasses: css,
```
agubler: There are a few significant breaking changes in the next release of widget-core, which will need to be planned accordingly. This shouldn't hold up the PR.

src/timepicker/styles/timePicker.m.css
```diff
@@ -0,0 +1,77 @@
...
+
+.root,
+.root * {
+	box-sizing: border-box;
+}
+
+.root {
+	display: inline-block;
+	position: relative;
+	width: var(--form-field-width);
```
tomdye: Do you think that fields should have a width, or that they should expand to take up the space they are given?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+}
+
+// Enum used when traversing items using arrow keys
+export const enum Operation {
+	increase = 1,
+	decrease = -1
+};
+
+// This can go away when Safari decides to support KeyboardEvent.key
+const keys: {[key: string]: number} = {
```
tomdye: this should be moved out to a keys module that can be loaded. I thought that @bitpshr had already done this?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+	escape: 27,
+	enter: 13,
+	up: 38,
+	down: 40
+};
+
+function getTime(units: TimeUnits, date = new Date()) {
```
tomdye: can you add jsdoc for this.

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+ * Convert a standard time string into an object with `hour`, `minute`, and `second` number properties.
+ *
+ * For example, '12:30' is converted to `{ hour: 12, minute: 30, second: 0 }`, and '19:03:27' is converted
+ * to `{ hour: 19, minute: 3, second: 27 }`.
+ *
+ * @param value   A standard time string or an object with `hour`, `minute`, and `second` properties.
+ * @throws Error  For any string not in the format 'HH:mm' or 'HH:mm:ss'.
+ * @return        An object containing `hour`, `second`, and `number` properties.
+ */
+export const parseUnits = (function () {
```
tomdye: Do you think  these functions would be better put into a util module?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+export interface TimePickerProperties extends ThemeableProperties {
+	autoBlur?: boolean;
+	buttonText?: string;
+	customOptionItem?: any;
+	customOptionMenu?: any;
+	disabled?: boolean; // defaults to false
+	end?: string | TimeUnits;
+	formId?: string;
+	getOptionLabel?: (option: TimeUnits) => string;
+	inputProperties?: TextInputProperties;
```
tomdye: Why are you abstracting these away rather than having them at the top level alongside the other properties? Doesn't seem clear to a user which properties belong where?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+	onChange?: (value: string) => void;
+	onFocus?: (value: string) => void;
+	onRequestOptions?(value: string, options: ReadonlyArray<TimeUnits>): void;
+	openOnFocus?: boolean;
+	options?: TimeUnits[];
+	readOnly?: boolean;  // defaults to false
+	required?: boolean; // defaults to false
+	start?: string | TimeUnits;
+	step?: number; // defaults to 60
+	value?: string;
```
tomdye: Is there not an option to use the native input available? This would be crucial for usage in mobile first projects.

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+ * @property start              The minimum time to display in the menu (defaults to '00:00:00')
+ * @property step               The number of seconds between each option in the menu (defaults to 60)
+ * @property value              Value to set on the input
+ */
+export interface TimePickerProperties extends ThemeableProperties {
+	autoBlur?: boolean;
+	buttonText?: string;
+	customOptionItem?: any;
+	customOptionMenu?: any;
+	disabled?: boolean; // defaults to false
```
tomdye: Do you think these comments are necessary?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+	private _ignoreBlur: boolean;
+	private _ignoreFocus: boolean;
+	private _inputElement: HTMLInputElement;
+	private _open: boolean;
+	private _options: ReadonlyArray<TimeUnits>;
+	private _startTime: number;
+	private _step: number;
+
+	render(): DNode {
+		const { disabled, formId, invalid, label, readOnly, required } = this.properties;
```
tomdye: Do the optional properties here need to have defaults added?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+		const options = this.renderOptions();
+		const menuId = options ? (<ResultMenuProperties> options.properties).id : '';
```
tomdye: Does this work if `options` does not contain `properties` or `properties.id` ?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+
+		let controls: DNode = v('span', {
```
tomdye: Do you think it's clear to use the same `controls` variable here for both cases?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+	protected renderOptions() {
+		if (!this._open) {
+			return null;
+		}
+
+		const { isOptionDisabled, options = [] } = this.properties;
+		return w('result-menu', <ResultMenuProperties> {
+			bind: this,
+			getResultLabel: this._getOptionLabel,
+			id: uuid(),
```
tomdye: is this id necessary?

src/timepicker/styles/timePicker.m.css
bitpshr: We probably need to revisit this when we get an actual theme. Currently all input-based elements use `--form-field-width`.

tomdye:
I'm finding the code in `TimePicker.ts` hard to follow if I'm honest so will take another look in the morning with fresh eyes. I'd envisaged a simple widget with an optional fallback to native that takes an array of children to display in a common results-list style popup when clicked on. All of the keyboard navigation etc.. would be in the common component and the formatting of the timestrings would be done by a util module or by the controlling component.

src/timepicker/TimePicker.ts
mwistrand: Perhaps not. I added them mostly out of habit.

src/timepicker/TimePicker.ts
mwistrand: A lot of the functionality for this widget was taken directly from `ComboBox` (hence #117), including this property. So rather than flattening out all possible `TextInputProperties`, I kept this consistent with `ComboBox`.

src/timepicker/TimePicker.ts
mwistrand: That PR just landed this morning, I think. I'll update to use `Keys` from `src/common/util`.

src/timepicker/TimePicker.ts
mwistrand: The issue excludes the fallback to native from the initial implementation. If @rishson agrees to hold off landing until that is added, I'll do so.

src/timepicker/TimePicker.ts
mwistrand: Perhaps, but at the moment it's exclusive to this widget.

src/timepicker/TimePicker.ts
mwistrand: It is not. I'll remove it.

mwistrand: My mistake, it is needed for the `aria-controls` property.

src/timepicker/TimePicker.ts
mwistrand: `renderOptions` returns a `WNode`, which will always have properties, and since the returned menu always has an ID, `menuId` will always be a string.

src/timepicker/TimePicker.ts
mwistrand: I consider the `Label` to be part of the `controls` so don't see how it would be confusing. That said, if it would be clearer to set a `label` variable and then check for that, I'll update.

mwistrand:
@tomdye I agree that changes will need to be made to all select-style components once the functionality is added to a common `SelectMixin`, but discussions are still taking place to determine what all that will look like. And I also agree that generating a list of options and passing that to the parent seems odd, but it's equally odd (at least to me) to have a time picker widget that can't actually fulfill one of its foundational requirements: show a list of times in between a specified start/end. And as I mentioned in an earlier comment, the issue excludes a native fallback from the initial implementation, and I resisted my usual urge to add every possible feature from the start; but if @rishson is fine with adding `useNativeElement` before this lands, I'm happy to do so.

src/timepicker/TimePicker.ts
smhigley: Some of those properties (disabled, invalid, readOnly, required) are probably better leaving undefined. Right now we treat undefined and false as the same, but in the future we could decide to allow users to add something like `aria-invalid="false"` vs. just leaving it off entirely.

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,481 @@
...
+				formId,
+				label
+			}, [ controls ]);
+		}
+
+		return v('span', {
+			'aria-expanded': this._open ? 'true' : 'false',
+			'aria-haspopup': 'true',
+			'aria-readonly': readOnly ? 'true' : 'false',
+			'aria-required': required ? 'true' : 'false',
```
smhigley: These two should probably be `readOnly/required ? 'true' : null`. I currently don't know of any benefit of including `aria-readonly` and `aria-required` if false.

src/timepicker/TimePicker.ts
smhigley: Not sure how much we care about this, but calling `uuid()` here will change the `id` every single render call, and require dom changes in both this and the button.

rishson:
@mwistrand feel free to add `useNativeElement` if you wish to do so.

agubler:
@mwistrand Needs a rebase to for the latest `beta1` release of `widget-core` and `test-extras`.

src/timepicker/TimePicker.ts
tomdye: Could this be changed to an exported function?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,356 @@
...
+
+			const [ hour, minute, second = 0 ] = value.split(':').map(unit => parseInt(unit, 10));
+			return { hour, minute, second } as TimeUnits;
+		}
+
+		return value;
+	};
+})();
+
+export const TimePickerBase = ThemeableMixin(WidgetBase);
```
tomdye: does this need to be exported?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,356 @@
...
+		} = this.properties;
+
+		return w(ComboBox, {
+			autoBlur,
+			bind: this,
+			clearable,
+			customResultItem: customOptionItem,
+			customResultMenu: customOptionMenu,
+			disabled,
+			extraClasses: extraClasses || css,
```
tomdye: not sure that `|| css` is right here

src/timepicker/TimePicker.ts
mwistrand: I don't see why it would be, but the `Base` class is exported in the other widgets, so I did so here for consistency.

agubler: You will have to export it because it is required to create the typings when doing a `grunt dist`.

An example of the error you would get `src/checkbox/Checkbox.ts(72,39): error TS4020: 'extends' clause of exported class 'Checkbox' has or is using private name 'CheckboxBase'.`

tomdye: Fair play @agubler . I remember that now

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,359 @@
...
+import Label, { LabelOptions } from '../label/Label';
+import { TextInputProperties } from '../textinput/TextInput';
+
+/**
+ * @type TimePickerProperties
+ *
+ * Properties that can be set on a TimePicker component
+ *
+ * @property autoBlur           Determines whether the input should blur after value selection
+ * @property buttonText         The text for the menu trigger (defaults to '')
```
bitpshr: Doesn't look like this property exists anymore.

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,359 @@
...
+	options?: TimeUnits[];
+	readOnly?: boolean;
+	required?: boolean;
+	start?: string;
+	step?: number;
+	useNativeElement?: boolean;
+	value?: string;
+}
+
+export interface TimeUnits {
```
bitpshr: We've been documenting all interfaces with at least a TypeDoc description.

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,359 @@
...
+	const i = step * 1000;
+	const options: number[] = [];
+
+	while (time <= endTime) {
+		options.push(time);
+		time += i;
+	}
+
+	return options.map((time: number) => {
+		const date = new Date(time);
```
bitpshr: This is heavy; I guess there's no way around this?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	@onPropertiesChanged()
+	protected onPropertiesChanged(event: PropertiesChangeEvent<this, TimePickerProperties>) {
+		if (
+			!this._getOptions ||
+			includes(event.changedPropertyKeys, 'start') ||
+			includes(event.changedPropertyKeys, 'end') ||
+			includes(event.changedPropertyKeys, 'step')) {
+			this._getOptions = (() => {
```
bitpshr: Why do you have to dynamically define / redefine `_getOptions`?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	private _getOptionLabel(value: TimeUnits) {
+		const { getOptionLabel } = this.properties;
+		const units = parseUnits(value);
+		return getOptionLabel ? getOptionLabel(units) : this._formatUnits(units);
+	}
+
+	private _onNativeBlur() {
+		this.properties.onBlur && this.properties.onBlur(this._nativeInputNode.value);
```
bitpshr: Couldn't you just use `event.target` for these native input node event handlers and do away with `_nativeInputNode` altogether?

src/timepicker/styles/timePicker.m.css
```diff
@@ -0,0 +1,25 @@
+@import '../../common/styles/variables.css';
+
+.root,
+.root * {
+	box-sizing: border-box;
+}
+
+.root {
+	display: inline-block;
+	/* width: var(--form-field-width); */
```
bitpshr: Can we remove this?

src/timepicker/TimePicker.ts
mwistrand: I might be able to reuse the same `Date` object so I'm not instantiating one each time.

src/timepicker/TimePicker.ts
mwistrand: `_getOptions` caches the list of options after creating it so it doesn't have to generate the default 1,440 values each time the results menu is shown. If either `start`, `end`, or `step` changes, then that cached list needs to be regenerated.

Alternatively, I could instead add a private `_options` instance variable so that it's the list that gets invalidated rather than the callback.

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,355 @@
...
+			.join(':');
+	}
+
+	private _getOptionLabel(value: TimeUnits) {
+		const { getOptionLabel } = this.properties;
+		const units = parseUnits(value);
+		return getOptionLabel ? getOptionLabel(units) : this._formatUnits(units);
+	}
+
+	private _getOptions() {
```
smhigley: I'm not sure I understand the purpose of this private `_getOptions` in conjunction with the exported `getOptions`. Since `properties.onRequestOptions` is getting passed this private function a user couldn't use the `start`/`end`/`step` params to easily filter options in any case, so why pass a function instead of the array of generated options? Also, why is it split into two functions?

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,355 @@
...
+
+/**
+ * Generate an array of time unit objects from the specified start date to the specified end date.
+ *
+ * @param start    The start time. Defaults to midnight.
+ * @param end      The end time. Defaults to 23:59:59.
+ * @param step     The amount of time in seconds between each step. Defaults to 60.
+ * @return         An array of time unit objects.
+ */
+export function getOptions(start: string = '00:00:00', end: string = '23:59:59', step = 60): TimeUnits[] {
```
smhigley: Is this only exported for unit tests?

mwistrand: Mostly, yes.

src/timepicker/TimePicker.ts
mwistrand: If app developers want to use a different method for generating options, they can without the added cost of having this additional array generated for them. `getOptions` generates the `TimeUnits` array; `_getOptions` caches the result so the list isn't generated multiple times for the same inputs.

mwistrand:
@smhigley The `getOptionLabel` property can be used to format values (e.g., using `@dojo/i18n/date`). Also, the `describedBy` property can be added to the `inputProperties` property, but I'll recycle any relevant properties from that object for the native `<input type="time">`.

src/timepicker/TimePicker.ts
smhigley: Hm, I see the benefit of caching the results, but I'm not sure it's necessary to avoid passing back the actual options array. If someone wants to use a different method to generate options, I think they should just extend the widget. @bitpshr do you have an opinion one way or the other?

mwistrand: I agree passing an array is cleaner; we probably do not need to worry about allowing customization of the `options` array.

smhigley: Cool, yeah, I like that better 👍. I'd also be in favor of `getOptions` being `protected` instead of `private` so it can be overridden

mwistrand: Good call. I should have thought of that... ;)

smhigley:
@mwistrand nice, I missed the `getOptionLabel` prop. For the a11y stuff, even with `describedBy` technically available I think it might be nice to have something like a `@property formatDescription: Add accessible text that describes the desired time format, e.g. "24-hour time rounded to the nearest half hour"`. It makes it slightly more obvious that it's something widget authors should think about.

mwistrand:
@smhigley I would much rather keep the current `describedBy` property with a note about how to use it properly in a README than add an additional property and have to figure out which one to pass to the `TextInput` widget. Perhaps `inputProperties` could be flattened and merged with `ComboBoxProperties` so we don't end up with two levels of properties affecting the root widget.

src/timepicker/example/index.ts
```diff
@@ -0,0 +1,172 @@
...
+				'Use Dojo Theme ',
+				v('input', {
+					type: 'checkbox',
+					onchange: this.themeChange
+				})
+			]),
+			v('h3', [ 'Filter options on input' ]),
+			w(TimePicker, {
+				inputProperties: {
+					placeholder: 'Enter a value'
```
smhigley: Since we're going the `inputProperties` route for a11y, could we add `describedBy` to all the `inputProperties` examples to model good a11y behavior? E.g.
```js
v('p', {
   id: 'foo',
   classes: this.classes().fixed(baseCss.visuallyHidden)
}, [ 'Accepts 24-hour time with a leading zero, rounded to the nearest half hour' ])
...
inputProperties: {
   placeholder: 'Do things',
   describedBy: 'foo'
}
```
I just want to make sure we model doing this because otherwise there's no indication of what the time format should be if you can't see the combobox options.

mwistrand:
@smhigley I added the readme, along with a 12-hour formatting example, and added `inputProperties.describedBy` to all input examples.

src/timepicker/TimePicker.ts
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/**
+ * Convert a standard time string into an object with `hour`, `minute`, and `second` number properties.
+ *
+ * For example, '12:30' is converted to `{ hour: 12, minute: 30, second: 0 }`, and '19:03:27' is converted
+ * to `{ hour: 19, minute: 3, second: 27 }`.
+ *
+ * @param value   A standard time string or an object with `hour`, `minute`, and `second` properties.
+ * @throws Error  For any string not in the format 'HH:mm' or 'HH:mm:ss'.
```
bitpshr: Do we need this `Error` type with TypeDoc for `@throws`?

src/timepicker/example/index.ts
```diff
@@ -0,0 +1,241 @@
...
+import { StatefulMixin } from '@dojo/widget-core/mixins/Stateful';
+import { theme, ThemeableMixin, ThemeableProperties } from '@dojo/widget-core/mixins/Themeable';
+import { v, w } from '@dojo/widget-core/d';
+import { WidgetBase } from '@dojo/widget-core/WidgetBase';
+import setLocaleData from './setLocaleData';
+import TimePicker, { TimeUnits } from '../TimePicker';
+import dojoTheme from '../../themes/dojo/theme';
+import * as baseCss from '../../common/styles/base.m.css';
+
+// Load all CLDR data required to render 12-hour time
```
bitpshr: I'm not sure these comments add any clarity, the code is easy enough to follow.

src/timepicker/README.md
```diff
@@ -0,0 +1,127 @@
...
+	readOnly: false,
+	required: true,
+	start: '00:00',
+	step: 60,
+	useNativeElement: true,
+	value: '10:30'
+});
+```
+
+## Themeing
```
smhigley: Just noticed -- this is misspelled, should be "Theming"

PR #121, opened 2017-04-21
SplitPane

### LINKED ISSUE #110: SplitPane
# Specification
This widget provides split pane behavior. Each SplitPane has two children and can be nested inside another SplitPane.
 
Split pane example

## Features
- vertical or horizontal orientation
- ability to specify a minimum panel size
- ability to specify a maximum panel size
- ability to specify an initial panel size
- ability to nest split panes

**Value-add**
Ability to divide content into resizable horizontal or vertical containers

**Callbacks**
`onResizeStart`
`onResizeEnd`
`onResize`

**Mouse/keyboard interactions**
Ability to click and drag divider line

**Mandatory/valid/empty/wait states**
None

**Controlled/uncontrolled**
Uncontrolled

**List of any icons needed**
None

Look and feel of the center splitter

**Any other considerations**:

## Acceptance criteria
- I can orient a SplitPane vertically or horizontally
- I can test SplitPanes inside other SplitPanes
- I can click the divider bar and resize the primary panel

bitpshr:
**Type:** feature

**Description:**

This PR adds a SplitPane component and associated tests. I decided to make the component fully controlled. This allowed for a really nice API, as properties like `minSize` and `maxSize` could go away entirely since a parent widget could simply not update `size` above or below some threshold. The component takes a `leading` and `trailing` property, each of which can be a `DNode`, that corresponds to the primary and secondary components to show in the SplitPane. SplitPanes can also be nested indefinitely to create any number of complex layouts.

_Note: Minor updates were made to a few components to get back to 100% coverage, things like moving member initialization to a constructor. No actual code logic was changed. I also alphabetized some properties for components I've written._

**Initial API**

```ts
direction?: Direction;
leading?: DNode;
onResize?(size: number): void;
size?: number;
trailing?: DNode;
```

Resolves #110 
Related to #214


dylans:
Never mind, I see the comment at the end of your description!

agubler:
@bitpshr Needs a rebase to for the latest `beta1` release of `widget-core` and `test-extras` and some conflicts to resolve.

Also not sure why the latest commit didn't run CI.

tomdye:
Why are there changes to combo box etc.. in this PR?

bitpshr:
@tomdye see the note in the PR description. I can back those out if necessary. I'll make sure not to spruce up unrelated components going forward.

src/combobox/ComboBox.ts
```diff
@@ -318,15 +318,15 @@ export default class ComboBox extends ComboBoxBase<ComboBoxProperties> {
...
 			getResultLabel: this._getResultLabel,
+			id: uuid(),
 			isResultDisabled: this.properties.isResultDisabled,
 			onResultMouseDown: this._onResultMouseDown,
 			onResultMouseEnter: this._onResultMouseEnter,
-			onResultMouseUp: this._onResultMouseUp
+			onResultMouseUp: this._onResultMouseUp,
+			registry: this.registry,
+			results,
+			selectedIndex: this._activeIndex
```
tomdye: you need to ensure you pass `theme` down to each widget you create

src/combobox/ComboBox.ts
```diff
@@ -352,39 +352,39 @@ export default class ComboBox extends ComboBoxBase<ComboBoxProperties> {
...
 				invalid,
-				readOnly,
-				required,
-				value,
 				onBlur: this._onInputBlur,
 				onFocus: this._onInputFocus,
 				onInput: this._onInput,
 				onKeyDown: this._onInputKeyDown,
-				overrideClasses: css
+				overrideClasses: css,
```
tomdye: has changed to `extraClasses`. Plus I think `css` is the wrong thing to pass here

src/splitpane/SplitPane.ts
```diff
@@ -0,0 +1,190 @@
...
+
+import * as css from './styles/splitPane.m.css';
+
+/**
+ * Direction of this SplitPane
+ */
+export const enum Direction {
+	column,
+	row
+};
```
mwistrand: Do these enum values need to be capitalized? E.g., `column` => `Column`.

src/splitpane/SplitPane.ts
```diff
@@ -0,0 +1,190 @@
...
+		}
+		else {
+			return event.changedTouches ? event.changedTouches[0].clientY : event.clientY;
+		}
+	}
+
+	private _onDragStart(event: MouseEvent & TouchEvent) {
+		this._dragging = true;
+		this._position = this._getPosition(event);
+	}
```
mwistrand: On mobile, without `event.preventDefault` will this result in the scroll position being changed as the divider is dragged?

src/splitpane/tests/unit/SplitPane.ts
```diff
@@ -0,0 +1,308 @@
...
+			selector: ':nth-child(2)' /* this should be the divider */
+		});
+		widget.sendEvent('touchend', {
+			selector: ':nth-child(2)' /* this should be the divider */
+		});
+
+		assert.strictEqual(called, 2);
+	},
+
+	'onResieze should be called'() {
```
smhigley: `onResize` 😄

src/splitpane/tests/unit/SplitPane.ts
```diff
@@ -0,0 +1,308 @@
...
+
+		splitPane.__setProperties__({ onResize: () => called++ });
+
+		(<any> splitPane)._root = { offsetWidth: 50, offsetHeight: 50 };
+		(<any> splitPane)._onResize();
+
+		splitPane.__setProperties__({
+			direction: Direction.column,
+			size: 25
+		});
```
smhigley: I'm not sure I understand the point of this test. Lines 297-302 can be removed without it failing, and it seems to only test that `this._onResize` calls `this.properties.onResize`, even though it's covered by earlier tests.

src/splitpane/SplitPane.ts
```diff
@@ -0,0 +1,190 @@
...
+			onResize,
+			size = DEFAULT_SIZE
+		} = this.properties;
+
+		const currentPosition = this._getPosition(event);
+		let newSize = (this._lastSize === undefined ? size : this._lastSize) + currentPosition - this._position;
+
+		const maxSize = direction === Direction.row ?
+			this._root.offsetWidth - this._divider.offsetWidth :
+			this._root.offsetHeight - this._divider.offsetHeight;
```
smhigley: This shouldn't block the PR, but the dimensions part of the animation POC would be useful here as an alternative to saving a reference to the dom. Would it make sense to add a TODO with a link to a github issue for it?

src/splitpane/SplitPane.ts
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	private _onResize() {
+		const {
+			direction = Direction.row,
+			onResize,
+			size = DEFAULT_SIZE
+		} = this.properties;
+
+		const rootSize = direction === Direction.row ? this._root.offsetWidth : this._root.offsetHeight;
```
bitpshr: It should be OK. This was being thrown off because I was using `vh` / `vw`.

smhigley: Hm, it's still a problem for me even when I change `vw`/`vh`. I get the leading div width set to the same pixel width as the container, so the divider ends up outside it.

src/splitpane/tests/unit/SplitPane.ts
bitpshr: This test verifies that `onResize` is called if a window resize causes the root resize. I updated the test to be a bit better.

src/splitpane/SplitPane.ts
bitpshr: Could you try with the latest now? (Just pushed some changes.) Also, could you attach a .gif or something so I could see what you're talking about? I may be trying to fix the wrong thing.

bitpshr:
@mwistrand @smhigley @tomdye rebased, feedback addressed

src/splitpane/SplitPane.ts
smhigley: It was actually the `flex: 0 1 auto` that fixed it, since now the leading div will shrink to allow for other fixed-with elements. It might still be a good idea to subtract `this._divider.offsetWidth` anyway to allow the leading div to start from the correct pixel width.

src/splitpane/SplitPane.ts
bitpshr: None of our existing enum values are capitalized. If we want to standardize on this, we should probably do it in a separate PR.

src/splitpane/SplitPane.ts
bitpshr: @smhigley I actually switched back to using `flex: 0 0 auto` for the leading div and `flex: 1 1 0%` for the trailing div. This is the more intuitive user experience to me, and how I originally coded it: the user can size the SplitPane, the end. The SplitPane doesn't attempt any cleverness during a window resize that may or may not cause the pane to resize. Things got messy when multiple levels of nested SplitPanes had incorrect explicit widths / heights set after window resizing since flexbox only uses these values as suggestions in certain cases.

smhigley: I'm fine with that. I still think the `rootSize` here should take into account the width of the divider. Right now if you resize down, you can lose the divider control entirely and be unable to resize the panes. If you change it to `const rootSize = direction === Direction.row ? this._root.offsetWidth - this._divider.offsetWidth : this._root.offsetHeight - this._divider.offsetHeight;`, it should be fixed.

bitpshr:
@smhigley I thought about this more and removed any window resize logic from the component itself. It seemed against our reactive pattern to have this one component listen to window resizing all to make sure individual panes in a SplitPane stay within range; that should really be a user concern.

This is ready for another review.

src/splitpane/tests/unit/SplitPane.ts
```diff
@@ -0,0 +1,304 @@
...
+})();
+
+let widget: Harness<SplitPaneProperties, typeof SplitPane>;
+
+registerSuite({
+	name: 'SplitPane',
+
+	beforeEach() {
+		widget = harness(SplitPane);
+		sinon.stub(util, 'observeViewport').returns({ unsubscribe: sinon.spy() });
```
smhigley: You don't need this anymore, right? (Also in `afterEach()`)

src/common/util.ts
```diff
@@ -12,3 +14,23 @@ export const enum Keys {
 	Tab = 9,
 	Up = 38
 }
+
+export const observeViewport = (function () {
```
tomdye: This isn't being used anywhere as part of this PR, can you remove please?

bitpshr:
I'm going to land this based on the approval of @smhigley and @tomdye since @mwistrand approved apart from minor nits.

PR #131, opened 2017-05-02
Calendar Widget

### LINKED ISSUE #1: Calendar Widget
@smhigley commented on [Thu Jan 19 2017](https://github.com/dojo/widget-core/issues/264)

**Enhancement**

Add an accessible calendar widget based on this mock: http://codepen.io/smhigley/pen/wodZbj

Related to #85.

smhigley:
**Type:** feature

**Description:**

Resolves #1 

This PR adds a standalone calendar widget. It doesn't include text input and dropdown, which will be split out to a separate datepicker widget as per the issue discussion. The calendar allows all text (months, weeks, and a11y messages) to be passed in via properties for i18n, defaulting to English. It's controlled via `year`, `month`, and `selectedDate` properties for the displayed year/month, and the user-selected date. Initial API is:

```javascript
customDateCell?: any; // widget constructor that extends CalendarCell
labels?: CalendarMessages; // a11y labels
month?: number;
monthNames?: { short: string; long: string; }[];
selectedDate?: Date;
weekdayNames?: { short: string; long: string; }[];
year?: number;
renderMonthLabel?(month: number, year: number): string;
renderWeekdayCell?(day: { short: string; long: string; }): DNode;
onMonthChange?(month: number): void;
onYearChange?(year: number): void;
onDateSelect?(date: Date): void;
```


tomdye:
Haven't got time to do a full review right now but can you make sure you pass the `theme` to each sub widget you create please.

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,364 @@
...
+	weekdayNames?: { short: string; long: string; }[];
+	year?: number;
+	renderMonthLabel?(month: number, year: number): string;
+	renderWeekdayCell?(day: { short: string; long: string; }): DNode;
+	onMonthChange?(month: number): void;
+	onYearChange?(year: number): void;
+	onDateSelect?(date: Date): void;
+};
+
+const TODAY = new Date();
```
agubler: What happens if you don't reload the module before the next day?

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,364 @@
...
+
+const DEFAULT_MONTHS = [
```
agubler: Should these be in an `i18n` bundle? As with the default weekdays and default labels.

src/calendar/Calendar.ts
smhigley: Once the calendar is open, I don't think it should change months or years without user interaction. `TODAY` only affects which month shows up if one isn't passed with properties.

agubler: But that's calculated when the module is loaded, not when the calendar is opened?

smhigley: Should it get updated in `render()`, then? It seems like it would add a fairly frequent extra function that ideally shouldn't be needed. Another option is Calendar could require `month` and `year`, and leave it up to the parent widget to default to `TODAY`

src/calendar/Calendar.ts
smhigley: That's fine with me! Should it go right in the Calendar widget?

mwistrand: Eventually at least the `DEFAULT_LABELS` should be in an `i18n` bundle, but that is true for more than just this widget. Adding the months to a bundle is trickier, since this functionality is already provided by `@dojo/i18n/date`. However, those methods require CLDR data that applications must supply themselves. So do we treat these as bundles messages, or include the necessary CLDR data here and use `@dojo/i18n/date`? This is probably a larger question that deserves its own issue, and might require a resolution to dojo/i18n#82.

agubler: Could you include minimal CLDR data for these (or a copy of the minimal) that gets used unless the an application has provided it's own CLDR data?

Agreed on these being a shared concern regardless of the i18n approach.

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,311 @@
...
+		const { monthNames, renderMonthLabel } = this.properties;
+		return renderMonthLabel ? renderMonthLabel(month, year) : monthNames[month].long + ' ' + year;
+	}
+
+	private _renderMonthRadios() {
+		const { month, theme = {} } = this.properties;
+
+		return this.properties.monthNames.map((monthName, i) => w(Radio, {
+			key: this._radiosName + i,
+			overrideClasses: { root: css.monthRadio, input: css.monthRadioInput, checked: css.monthRadioChecked },
```
tomdye: Can you update this to `extraClasses` please?

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,367 @@
...
+	year?: number;
+	renderMonthLabel?(month: number, year: number): string;
+	renderWeekdayCell?(day: { short: string; long: string; }): DNode;
+	onMonthChange?(month: number): void;
+	onYearChange?(year: number): void;
+	onDateSelect?(date: Date): void;
+};
+
+const DEFAULT_MONTHS = [
+	{short: 'Jan', long: 'January'},
```
tomdye: Should these be moved to `nls` ?

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,311 @@
...
+			checked: i === month,
+			label: {
+				content: `<abbr title="${monthName.long}">${monthName.short}</abbr>`,
```
tomdye: should this not be separate vdom call?

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,367 @@
...
+				this._goToDate(monthLengh);
+				break;
+			case Keys.Enter:
+			case Keys.Space:
+				const { onDateSelect } = this.properties;
+				onDateSelect && onDateSelect(new Date(year, month, this._focusedDay));
+		}
+	}
+
+	private _onMonthIncrement(direction: 'up' | 'down') {
```
smhigley: `onMonthCrement` 😄 
I'm really bad at naming things. I'd be happy to change it if you have a better name

src/calendar/MonthPicker.ts
smhigley: That's an interesting question. It's currently typed as `content: string` in form widget properties, and I'm not sure I'd want to encourage adding DNodes to label content. It could be changed, though.

src/calendar/Calendar.ts
smhigley: Ant mentioned that too, and unfortunately that thread got collapsed. As @mwistrand  said, there's still some discussion needed around CLDR data in i18n (e.g. https://github.com/dojo/i18n/issues/82)

src/calendar/Calendar.ts
smhigley: Also, in math at least, an increment can be positive or negative. But mostly I couldn't think of a better name :P

devpaul: I just happened to see this in dojo notifications and that it involved math so it piqued my curiosity... but, if you split this into two methods `_onMonthIncrement` and `_onMonthDecrement` you'd have fewer conditional blocks and wouldn't need to pass a direction.

smhigley: Hm, I think there would actually be more conditional blocks. Two functions = two if/else blocks vs. one if/else if/else block. It would mean no direction, though. I'm ambivalent; I actually started with separate functions and merged them b/c they shared a bunch of logic. The function has simplified since then though, so maybe it's time to split again.

smhigley: I'm good with that. Originally there was more repetitive logic involved so it made sense to only have one function. Now I think separating it might be best, and I like getting rid of `direction`

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+
+	private _onDateClick(date: number, disabled: boolean) {
+		const { onDateSelect } = this.properties;
+		let {
+			month,
+			year
+		} = this._getMonthYear();
+
+		if (disabled && date < 15) {
+			({ month, year } = this._onMonthIncrement());
```
agubler: do you need the wrapping brackets?

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+			month,
+			year
+		} = this._getMonthYear();
+		const { theme = {} } = this.properties;
+
+		const currentMonthLength = this._getMonthLength(month, year);
+		const previousMonthLength = this._getMonthLength(month - 1, year);
+		const initialWeekday = new Date(year, month, 1).getDay();
+
+		let dayIndex = 0,
```
agubler: should have a `let` per variable (plus the formatting is a bit funky!)

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+				dayIndex++;
+
+				// set isSelectedDay if the dates match
+				if (isCurrentMonth && selectedDate && new Date(year, month, date).toDateString() === selectedDate.toDateString()) {
+					isSelectedDay = true;
+				} else {
+					isSelectedDay = false;
+				}
+
+				days.push(w('date-cell', <CalendarCellProperties> {
```
agubler: This should be `w<DateCell>('date-cell', ...)` now

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+
+		return weeks;
+	}
+
+	private _renderWeekdayCell(day: { short: string; long: string; }): DNode {
+		const { renderWeekdayCell } = this.properties;
+		return renderWeekdayCell ? renderWeekdayCell(day) : v('abbr', { title: day.long }, [ day.short ]);
+	}
+
+	render() {
```
agubler: should be `protected render(): DNode {`

src/calendar/CalendarCell.ts
```diff
@@ -0,0 +1,89 @@
...
+		const { date, disabled = false, onClick } = this.properties;
+		onClick && onClick(date, disabled);
+	}
+
+	private _onKeyDown(event: KeyboardEvent) {
+		const { onKeyDown } = this.properties;
+		onKeyDown && onKeyDown(event);
+	}
+
+	formatDate(date: number): DNode {
```
agubler: `public`, `protected` or `private`?

src/calendar/CalendarCell.ts
```diff
@@ -0,0 +1,89 @@
...
+		return v('span', [ String(date) ]);
+	}
+
+	render() {
```
agubler: `protected render(): DNode {`

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,300 @@
...
+		if (
+			event.which === Keys.Escape ||
+			event.which === Keys.Enter ||
+			event.which === Keys.Space
+		) {
+			this._closePopup();
+		}
+	}
+
+	private _onRadioChange(event: Event) {
```
agubler: I think that there is a `TargetedEventObject` in `widget-core` that you could use like `TargetedEventObject<HTMLInputElement>`

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,300 @@
...
+				break;
+			case Keys.Left:
+				this._decreaseYear();
+				break;
+		}
+	}
+
+	private _renderMonthLabel(month: number, year: number) {
+		const { monthNames, renderMonthLabel } = this.properties;
+		return renderMonthLabel ? renderMonthLabel(month, year) : monthNames[month].long + ' ' + year;
```
agubler: could use template string here?

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,300 @@
...
+				content: `<abbr title="${monthName.long}">${monthName.short}</abbr>`,
+				before: false
+			},
+			name: this._radiosName,
+			// TODO: replace this with a method to "unfocus" components?
+			// tabIndex: open ? 0 : -1,
+			theme,
+			value: i + '',
+			onChange: this._onRadioChange,
+			onMouseUp: () => {
```
agubler: Any reason it couldn't just be `onMouseUp: this._closePopup`?

src/calendar/tests/unit/Calendar.ts
```diff
@@ -0,0 +1,474 @@
...
+
+let widget: Harness<CalendarProperties, typeof Calendar>;
+const testDate = new Date('June 3 2017');
+
+const compareId = compareProperty((value: any) => {
+	return typeof value === 'string';
+});
+
+const expectedDateCell = function(widget: any, date: number, active: boolean) {
+	return w('date-cell', <CalendarCellProperties> {
```
agubler: Same comment as earlier, it should be `w<DateCell>(....)`

src/calendar/tests/unit/Calendar.ts
```diff
@@ -0,0 +1,474 @@
...
+		selected: false,
+		onClick: widget.listener,
+		onFocusCalled: widget.listener,
+		onKeyDown: widget.listener
+	});
+};
+
+const expected = function(widget: any) {
+	return v('div', { classes: widget.classes(css.root) }, [
+		w(MonthPicker, <any> {
```
agubler: shouldn't need an `any` here?

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+		return lastDate.getDate();
+	}
+
+	private _getMonthYear() {
+		const {
+			month,
+			selectedDate = this._defaultDate,
+			year
+		} = this.properties;
+		return { month: typeof month === 'number' ? month : selectedDate.getMonth(), year: year ? year : selectedDate.getFullYear() };
```
agubler: this is a bit hard to digest

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+		}
+
+		this._focusedDay = day;
+		this._callDateFocus = true;
+		this.invalidate();
+	}
+
+	private _onDateClick(date: number, disabled: boolean) {
+		const { onDateSelect } = this.properties;
+		let {
```
agubler: For examples like this do we need to have a carriage return for each variable?

```ts
let { month, year } = this._getMonthYear();
```

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+				this._goToDate(monthLengh);
+				break;
+			case Keys.Enter:
+			case Keys.Space:
+				const { onDateSelect } = this.properties;
+				onDateSelect && onDateSelect(new Date(year, month, this._focusedDay));
+		}
+	}
+
+	private _onMonthDecrement() {
```
agubler: I think you recently split these out but it does seem they share quite a lot of logic 😛

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+			onMonthChange,
+			onYearChange
+		} = this.properties;
+
+		if (month === 0) {
+			onMonthChange && onMonthChange(11);
+			onYearChange && onYearChange(year - 1);
+			return { month: 11, year: year - 1 };
+		}
+		else {
```
agubler: can drop the `else` here

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+			onMonthChange,
+			onYearChange
+		} = this.properties;
+
+		if (month === 11) {
+			onMonthChange && onMonthChange(0);
+			onYearChange && onYearChange(year + 1);
+			return { month: 0, year: year + 1 };
+		}
+		else {
```
agubler: same here can drop the `else`

src/calendar/example/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import { ProjectorMixin } from '@dojo/widget-core/mixins/Projector';
+import { v, w } from '@dojo/widget-core/d';
+import Calendar from '../../calendar/Calendar';
+
+const AppBase = StatefulMixin(WidgetBase);
+export class App extends AppBase<WidgetProperties> {
+	render() {
+		return v('div', {}, [
+			w(Calendar, {
+				month: <number> this.state.month,
```
agubler: do you need the casts here?

src/calendar/tests/functional/Calendar.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+function clickDate(remote: any) {
+	const today = new Date();
+	const firstDay = new Date(today.getFullYear(), today.getMonth(), 1).getDay();
+	return remote
+		.get('http://localhost:9000/_build/common/example/?module=calendar')
+		.setFindTimeout(5000)
+		.findByCssSelector(`tbody > tr:first-child > td:nth-child(${firstDay + 1})`)
+			.click()
```
agubler: is this formatting normal?

src/common/tests/unit/all.ts
```diff
@@ -2,7 +2,10 @@ import '@dojo/test-extras/support/loadJsdom';
 import './main';
 import './util';
 import '../../../dialog/tests/unit/Dialog';
-import '../../../combobox/tests/unit/ComboBox';
+import '../../../calendar/tests/unit/Calendar';
+import '../../../calendar/tests/unit/CalendarCell';
+import '../../../calendar/tests/unit/MonthPicker';
+// import '../../../combobox/tests/unit/ComboBox';
```
agubler: should this be commented out?

src/button/Button.ts
```diff
@@ -35,6 +35,7 @@ export interface ButtonProperties extends ThemeableProperties {
 	content?: string;
 	describedBy?: string;
 	disabled?: boolean;
+	id?: string;
```
agubler: Should probably have a unit test to cover this addition.

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,391 @@
...
+ * @property weekdayNames      Customize or internationalize weekday names and abbreviations
+ * @property year              Set the currently displayed year
+ * @property renderMonthLabel  Format the displayed current month and year
+ * @property renderWeekdayCell Format the weekday column headers
+ * @property onMonthChange     Function called when the month changes
+ * @property onYearChange      Function called when the year changes
+ * @property onDateSelect      Function called when the user selects a date
+ */
+export interface CalendarProperties extends ThemeableProperties {
+	customDateCell?: any;
```
agubler: Is this really an `any`?

src/calendar/Calendar.ts
smhigley: It didn't work for me otherwise, and that's the syntax I see in the MDN examples: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

agubler: Fair, I just tried it in node... where it is does work without the brackets.

src/calendar/Calendar.ts
smhigley: I'll be perpetually stuck in a loop of splitting them out and merging them again 😂 .
I'm really fine either way, they seem on the edge of annoying repetitiveness vs. annoying conditionals

src/calendar/Calendar.ts
smhigley: spaces > tabs :trollface:

src/calendar/MonthPicker.ts
smhigley: I just love me some extra anonymous arrow functions 😂

src/common/tests/unit/all.ts
smhigley: Oh wow, I have no idea why I did that

src/calendar/tests/functional/Calendar.ts
smhigley: Should I indent the `.then()` blocks? I've mostly been going off formatting in other widget functional tests.

src/calendar/tests/unit/Calendar.ts
smhigley: Switched it to individual properties using `compareProperty` 👍

src/calendar/example/index.ts
smhigley: not anymore!

src/calendar/tests/functional/Calendar.ts
mwistrand: And the other functional tests in `@dojo/widgets` are formatted according to the [Intern example](http://theintern.github.io/intern/#writing-functional-test), which indents the operations that pertain to a specific element.

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,389 @@
...
+
+				// set isSelectedDay if the dates match
+				if (isCurrentMonth && selectedDate && new Date(year, month, date).toDateString() === selectedDate.toDateString()) {
+					isSelectedDay = true;
+				} else {
+					isSelectedDay = false;
+				}
+
+				days.push(w<CalendarCell>('date-cell', {
+					key: date + '-' + (isCurrentMonth ? month : 'inactive'),
```
bitpshr: May be a nice place for a template string.

src/calendar/CalendarCell.ts
```diff
@@ -0,0 +1,89 @@
...
+export default class CalendarCell extends CalendarCellBase<CalendarCellProperties> {
+	protected onElementCreated(element: HTMLElement, key: string) {
+		const { callFocus, onFocusCalled } = this.properties;
+		if (callFocus) {
+			element.focus();
+			onFocusCalled && onFocusCalled();
+		}
+	}
+	protected onElementUpdated(element: HTMLElement, key: string) {
+		const { callFocus, onFocusCalled } = this.properties;
```
bitpshr: You could pull out the common focus logic of `onElementCreated` and `onElementUpdated` into a private method, like `_focus`.

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,299 @@
...
+import uuid from '@dojo/core/uuid';
+import { Keys } from '../common/util';
+import { DNode, TypedTargetEvent } from '@dojo/widget-core/interfaces';
+import Radio from '../radio/Radio';
+import Button from '../button/Button';
+import * as css from './styles/calendar.m.css';
+import * as baseCss from '../common/styles/base.m.css';
+
+// will need to localize messages
+export interface CalendarMessages {
```
bitpshr: I think this needs proper TypeDoc so it plays nicely with the doc generator.

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,299 @@
...
+	onRequestOpen?(): void;
+	onRequestClose?(): void;
+	onRequestMonthChange?(month: number): void;
+	onRequestYearChange?(year: number): void;
+};
+
+export const MonthPickerBase = ThemeableMixin(WidgetBase);
+
+@theme(css)
+export default class MonthPicker extends MonthPickerBase<MonthPickerProperties> {
```
bitpshr: Just curious, why didn't you put MonthPicker in the Calendar registry? I guess it's probably unlikely that consumers would need to customize it often.

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,299 @@
...
+	private _buttonId = uuid();
+	private _dialogId = uuid();
+	private _labelId = uuid();
+	private _yearSpinnerId = uuid();
+	private _radiosName = uuid();
```
bitpshr: I'm not sure there'd be a huge performance gain, but couldn't you use a single `uuid()` with prefixes or suffixes for each of these?

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,299 @@
...
+
+	private _callTriggerFocus = false;
+	private _callPopupFocus = false;
+
+	// move focus when opening/closing the popup
+	protected onElementUpdated(element: HTMLElement, key: string) {
+		// TODO: find a way to call focus on button widget
```
bitpshr: Are you still investigating this? Should this prevent the PR from landing?

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,299 @@
...
+		return this.properties.monthNames.map((monthName, i) => w(Radio, {
+			key: this._radiosName + i,
+			extraClasses: { root: css.monthRadio, input: css.monthRadioInput, checked: css.monthRadioChecked },
+			checked: i === month,
+			label: {
+				content: `<abbr title="${monthName.long}">${monthName.short}</abbr>`,
+				before: false
+			},
+			name: this._radiosName,
+			// TODO: replace this with a method to "unfocus" components?
```
bitpshr: Same question about this TODO.

src/calendar/MonthPicker.ts
smhigley: Hm, I guess I didn't really see it as something that people would customize. Right now everything in there apart from `render` is private, so it's not easily extendable. I wonder if I should start making more things protected over private, both in MonthPicker and in general?

src/calendar/MonthPicker.ts
smhigley: The problem here (and below with the radio buttons) is it involves handling focus and focusability within other widgets. I could either modify all form widgets to be able to be focused and add `tabIndex` to properties or get a working focus manager. I'm inclined towards the latter, but if Calendar needs to wait for a focus manager it'll be here a while :P. I could update the comment to say it needs to be fixed once the focus manager lands and put in a link to the issue

src/calendar/Calendar.ts
agubler: It's cool, lets leave them as two separate functions for now 👍

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,311 @@
...
+		const {
+			month,
+			year,
+			labelId = `${this._idBase}_label`,
+			labels,
+			open = false
+		} = this.properties;
+
+		return v('div', { classes: this.classes(css.header) }, [
+			// button
```
tomdye: I don't think these comments are required.

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,385 @@
...
+	year?: number;
+	renderMonthLabel?(month: number, year: number): string;
+	renderWeekdayCell?(day: { short: string; long: string; }): DNode;
+	onMonthChange?(month: number): void;
+	onYearChange?(year: number): void;
+	onDateSelect?(date: Date): void;
+};
+
+export const DEFAULT_MONTHS = [
+	{short: 'Jan', long: 'January'},
```
smhigley: Agreed this needs to change; per OSS conversation, will track in this issue and raise a separate PR: https://github.com/dojo/widgets/issues/213

src/calendar/Calendar.ts
smhigley: @agubler @tomdye so far all the other widgets that have accepted custom widget registries in `properties` have typed them as `any` and used the `custom` prefix.

agubler: Well I would say we'd probably want to type them correctly everywhere, I think it would be `Constructor<Cell>` or `Constructor<WidgetBase>`? No? Assuming it accepts a `WidgetConstructor`?

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,385 @@
...
+	chooseMonth: 'Choose Month',
+	chooseYear: 'Choose Year',
+	previousMonth: 'Previous Month',
+	nextMonth: 'Next Month'
+};
+
+export const CalendarBase = ThemeableMixin(WidgetBase);
+
+@theme(css)
+@diffProperty('customDateCell', DiffType.REFERENCE)
```
bitpshr: I thought by default all object properties were diffed by reference, no? If not, we have a few bugs in other registry-based components I think.

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,385 @@
...
+				else {
+					date++;
+				}
+				dayIndex++;
+
+				// set isSelectedDay if the dates match
+				dateString = new Date(year, month, date).toDateString();
+				if (isCurrentMonth && selectedDate && dateString === selectedDate.toDateString()) {
+					isSelectedDay = true;
+				} else {
```
bitpshr: `else` should be on a new line

src/calendar/MonthPicker.ts
```diff
@@ -0,0 +1,311 @@
...
+			}, [
+				v('span', { classes: this.classes().fixed(baseCss.visuallyHidden) }, [ labels.chooseMonth ])
+			]),
+			v('label', {
+				id: labelId,
+				classes: this.classes(css.currentMonthLabel),
+				'aria-live': 'polite',
+				'aria-atomic': 'false'
+			}, [ this._renderMonthLabel(month, year) ]),
+			// previous/next month buttons
```
bitpshr: You can lose these comments probably.

src/calendar/Calendar.ts
smhigley: `Constructor<CalendarCell>` seems to work, I'll use that 👍

src/calendar/Calendar.ts
agubler: If it's a widget constructor, then it's actually by default ignored (because it's a function) - so to detect changes in between the custom constructors it does need the reference DiffType.

fwiw, object properties are swallow diff'd by default.

smhigley: @bitpshr I noticed it b/c tests weren't hitting `if (this._registry)`. Have you run across that in any other widgets that take a constructor in properties?

src/calendar/MonthPicker.ts
smhigley: Hah, @tomdye said the same thing. It's a pretty large chunk of VDOM, though. I have trouble finding what I'm looking for at a glance, and I wrote it 😕. Are you sure we shouldn't have comments in there for later?

bitpshr: NO COMMENTS ALLOWED

No, if you needed them, I'd say leave them for sure.

src/calendar/README.md
```diff
@@ -0,0 +1,91 @@
...
+### i18n Features
+- Localized versions of `CalendarMessages`, `monthNames`, and `weekdayNames` can be passed in via properties.
+
+## Example Usage
+
+*Basic Example*
+```js
+import Calendar from '@dojo/widgets/calendar/Calendar';
+
+protected render: DNode () {
```
bitpshr: I don't think we need to show the `import` of the module and a pseudocoded `render` function. This isn't valid code anyway, so why not just show the `w` call to keep boilerplate to a minimum? We also don't know if a user will be importing from the index module, this module, etc.

src/calendar/MonthPicker.ts
smhigley: I made them a little more consistent and specific. If you still don't like them, I can burn them all with 🔥

src/calendar/Calendar.ts
```diff
@@ -0,0 +1,389 @@
...
+	{short: 'Sun', long: 'Sunday'},
+	{short: 'Mon', long: 'Monday'},
+	{short: 'Tue', long: 'Tuesday'},
+	{short: 'Wed', long: 'Wednesday'},
+	{short: 'Thu', long: 'Thursday'},
+	{short: 'Fri', long: 'Friday'},
+	{short: 'Sat', long: 'Saturday'}
+];
+
+export const DEFAULT_LABELS: CalendarMessages = {
```
agubler: Are we nls'ing this? I cannot remember the outcome of this conversation.

smhigley: I believe the outcome was that it'll be a separate PR based on the consensus in https://github.com/dojo/widgets/issues/213

PR #134, opened 2017-05-04
Add Dialog functional tests

mwistrand:
**Type:** feature

**Description:**

Adds basic functional tests for `Dialog`.

It should be noted that I have not been able to figure out how to get a consistent read on the window/dialog dimensions. In Chrome, for example, the registered height of the underlay is always ~60px less than the window height, even though during the tests themselves the underlay clearly occupies the entire screen. As such, rather liberal estimates have to be used to determine whether the underlay spans the entire screen, and whether the dialog is actually centered. If anyone has ideas for getting a more accurate reading, I would greatly appreciate the input.

### DIFF
--- a/src/dialog/tests/functional/Dialog.ts
+++ b/src/dialog/tests/functional/Dialog.ts
@@ -1,0 +1,149 @@
+import * as registerSuite from 'intern!object';
+import * as assert from 'intern/chai!assert';
+
+interface Options {
+	closeable?: boolean;
+	modal?: boolean;
+	underlay?: boolean;
+}
+
+const CONTAINER_SELECTOR = 'div > div';
+const DELAY = 400;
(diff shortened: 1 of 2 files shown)


tomdye:
Please update to use css-module selectors

src/dialog/tests/functional/Dialog.ts
```diff
@@ -0,0 +1,150 @@
...
+import * as assert from 'intern/chai!assert';
+import * as css from '../../styles/dialog.m.css';
+
+interface Options {
+	closeable?: boolean;
+	modal?: boolean;
+	underlay?: boolean;
+}
+
+const CONTAINER_SELECTOR = 'div > div';
```
smhigley: Could this be `.${css.root}`?

pottedmeat:
Moved to https://github.com/dojo/widgets/pull/317

Issue #75, opened 2017-03-17
Use `root` classname consistently for outermost HTMLElement of widgets


tomdye:
@tomdye commented on [Tue Mar 14 2017](https://github.com/dojo/widget-core/issues/407)

**Enhancement** 

Do we feel that it would be advantageous to consistently use `root` as the outermost classname for all of our dojo 2 widgets?

- Would provide a consistent way to pass `overrideClasses` to widget outer HTMLElement.
- Would lower the barrier to entry to applying classes to 'black box' widgets.
- Allows the widget author to determine the `root` element in the case of widgets such as `Dialog`
- Allows `spacing` type rules to be applied safely to a widget with greater ease.

**Example**

``` css
/* myLayout.css */
.marginClass {
   margin: 10px;
}
```

``` typescript
import * as myLayout from './myLayout.css';

/* can safely target the root class */
w(Button, { overrideClasses: { root: myLayout.marginClass } });
```

---

@tomdye commented on [Tue Mar 14 2017](https://github.com/dojo/widget-core/issues/407#issuecomment-286503637)

@rishson  Do we feel this is `beta1` or `beta2`? I suspect the latter.

---

@matt-gadd commented on [Fri Mar 17 2017](https://github.com/dojo/widget-core/issues/407#issuecomment-287382044)

@tomdye this is in the wrong repo no?

---

@tomdye commented on [Fri Mar 17 2017](https://github.com/dojo/widget-core/issues/407#issuecomment-287382161)

correct, can it be moved to `widgets` or do i have to create a new one?

rishson:
If we are going to do this, then we need to check the following are compliant:

- [ ] Button
- [ ] CheckBox
- [ ] ComboBox
- [ ] Dialog
- [ ] Label
- [ ] Menu
- [ ] NavBar
- [ ] Radio
- [ ] SlidePane
- [ ] Slider
- [ ] TabPane
- [ ] TextArea
- [ ] TextInput
- [ ] TimePicker

tomdye:
Dialog was missing `root` class. Label can have `root` class added once it's `classes` property is removed in another PR.

PR #167, opened 2017-05-09
Label: update API and update unit tests to use test-extras

### LINKED ISSUE #139: Label: Update Unit Tests
Update Label unit tests to use `test-extras`.

msssk:
**Type:** enhancement

**Description:**

In addition to test updates, this PR changes the `Label` API:

* remove the custom `classes` property and use the `extraClasses` property from `ThemeableMixin` instead
* exports a `parseLabelClasses` function to facilitate using `extraClasses`

Resolves #139 #75


src/checkbox/Checkbox.ts
```diff
@@ -141,7 +141,7 @@ export default class Checkbox extends CheckboxBase<CheckboxProperties> {
 
 		if (label) {
 			checkboxWidget = w(Label, {
-				classes: this.classes(css.root, ...stateClasses),
+				extraClasses: this.classes(css.root, ...stateClasses),
```
tomdye: @msssk This usage of `extraClasses` isn't quite right. The idea is to pass an object that references the class name key you wish to add classes to and a space-separated list of the classes you want to add to it. The idea being that in most cases you'd only be adding one or two classes via `extraClasses`. I guess that label is a bit of an edge case as it's being used to wrap an input. Something like the following should get you where you need to be:

``` typescript
function getExtraClasses(classes: { [ string ]: boolean }): string {
    return Object.keys(classes).reduce((classNamesString: string, className) => {
        if (classes[className] === true) {
            classNamesString += ` ${className}`;
        }
        return classNamesString;
    }, '').trim();
}

if (label) {
	checkboxWidget = w(Label, {
		extraClasses: { 'root': getExtraClasses(this.classes(css.root, ...stateClasses).get()) },
		formId,
		label
	}, children);
}
else {
	checkboxWidget = v('div', {
		classes: this.classes(css.root, ...stateClasses)
	}, children);
}
```

tomdye: As this is going to be called in each instance that a `label` widget is used, I think that `parseLabelClasses` should be a function exported from `label`.

msssk: Thank you, this was very helpful - updated.

tomdye: You're welcome

src/label/Label.ts
```diff
@@ -30,17 +30,30 @@ const labelDefaults = {
...
+			classNamesString += ` ${className}`;
+		}
+
+		return classNamesString;
+	}, '').trim();
+}
+
 export const LabelBase = RegistryMixin(ThemeableMixin(WidgetBase));
 
 @theme(baseCss)
```
tomdye: Finally, to enable this to actually work and to finish of #75 Could you please create `label.m.css` under the `label/styles` directory, change this to `@theme(css)` and pass an empty `css.root` to the label's root node please. Otherwise `extraClasses` will have no target `root` to apply the classes to.

src/radio/Radio.ts
```diff
@@ -119,7 +119,7 @@ export default class Radio extends RadioBase<RadioProperties> {
 
 		if (label) {
 			radioWidget = w(Label, {
-				classes: this.classes(css.root, ...stateClasses),
+				extraClasses: this.classes(css.root, ...stateClasses),
```
tomdye: This usage is still wrong at the moment. Please see the example previously.

tomdye:
Can you update the title of the PR please to represent the changes to the label api.

msssk:
@tomdye 
> Can you update the title of the PR please to represent the changes to the label api.

What changes to the API? What should the title be changed to?

tomdye:
Question for @bitpshr @smhigley, do we still think this is the best approach, ie. use of `extraClasses` given the theming discussion issues?

smhigley:
@tomdye I'm not a huge fan of it. Since this is meant to be a Label-only usage, maybe we should go back to having a Label-only property that takes a `ClassesFunctionChain` or whatever the correct type for `this.classes()` is.

I'm also a little curious how outside users will respond to cases like this. I can see wanting to apply extra styling classes being a somewhat common use case, so I wonder if we should work a little more on demoing good patterns for extending widgets and applying custom styles. Not necessarily in Label, since it needs additional themed classes rather than just a custom theme, but perhaps in an article or README.

smhigley:
That conversation is partly why I don't really like it here anymore. For the second part of my comment, I just meant we should have better documentation around how and when to extend a widget to apply individual styling (i.e. when you're not just theming all instances of a widget).

agubler:
@smhigley we shouldn't be passing the function, but perhaps could go back to passing the resulting class object after calling `.get()`.

agubler:
@bitpshr @smhigley @tomdye Does anyone know the status of this? I think it is blocking a whole load of the other testing changes.

tomdye:
Last conversation I had with @smhigley I was hoping she could change `label` to not take any classes at all and just use it's own based off of state. Not sure where she got with it.

agubler:
Yup let's do something as this is blocking other pull requests afaik

smhigley:
@agubler I put a potential solution at https://github.com/msssk/widgets/compare/label-test-update...smhigley:label-test-update, but I got the sense you and @tomdye didn't like the widget extension.

I'm also not big on dumping everything directly into `Label`. Should we have e.g. `isToggleCheckbox: boolean` in `LabelProperties` just to correctly handle the one checkbox class, even though it has no bearing on label functionality? That also doesn't solve the broader problem of users sometimes needing to add additional classes to existing widgets.

PR #168, opened 2017-05-10
Add `Button` functional tests.

### LINKED ISSUE #149: Button:  Add Functional Tests

lzhoucs:
**Type:** functional tests

Resolves #149

### DIFF
--- a/src/button/tests/functional/Button.ts
+++ b/src/button/tests/functional/Button.ts
@@ -13,11 +13,11 @@ registerSuite({
 	'button should be visible'(this: any) {
 		return getPage(this.remote)
 			.findByCssSelector('button:first-of-type')
-		// as per mwistrand, `isDisplayed` is broken in Safari 10, commented out for now
-			// .isDisplayed()
-			// .then((displayed: boolean) => {
-			// 	assert.isTrue(displayed, 'The button should be displayed.');
-			// })
+		// `isDisplayed` is broken in Safari 10, commented out for now. See https://github.com/SeleniumHQ/selenium/issues/3029
+		// .isDisplayed()
+		// .then((displayed: boolean) => {
+		// 	assert.isTrue(displayed, 'The label should be displayed.');
+		// })
 			.getSize()
 			.then(({ height, width }: { height: number; width: number; }) => {
 				assert.isAbove(height, 0, 'The button height should be greater than zero.');
--- a/src/button/tests/functional/Button.ts
+++ b/src/button/tests/functional/Button.ts
@@ -13,11 +13,6 @@ registerSuite({
 	'button should be visible'(this: any) {
 		return getPage(this.remote)
 			.findByCssSelector('button:first-of-type')
-		// `isDisplayed` is broken in Safari 10, commented out for now. See https://github.com/SeleniumHQ/selenium/issues/3029
-		// .isDisplayed()
-		// .then((displayed: boolean) => {
-		// 	assert.isTrue(displayed, 'The label should be displayed.');
-		// })
 			.getSize()
 			.then(({ height, width }: { height: number; width: number; }) => {
 				assert.isAbove(height, 0, 'The button height should be greater than zero.');
--- a/src/button/tests/functional/Button.ts
+++ b/src/button/tests/functional/Button.ts
@@ -1,0 +1,73 @@
+import * as registerSuite from 'intern!object';
+import * as assert from 'intern/chai!assert';
+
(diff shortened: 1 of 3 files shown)


src/button/tests/functional/Button.ts
```diff
@@ -0,0 +1,73 @@
...
+		.setFindTimeout(5000);
+}
+
+registerSuite({
+	name: 'Button',
+
+	'button should be visible'(this: any) {
+		return getPage(this.remote)
+			.findByCssSelector('button:first-of-type')
+		// as per mwistrand, `isDisplayed` is broken in Safari 10, commented out for now
```
dylans: Could we add a link to a Selenium driver issue listed somewhere so we can come back and fix this later?

dylans: Also comment indentation is off

lzhoucs: I just added the Selenium issue link and fixed the indentation. Please let me know if there's any other issues. Thanks!

lzhoucs:
Yeah. It is the known Safari issue on browserstack for today: `NoSuchDriver`.

src/button/tests/functional/Button.ts
```diff
@@ -0,0 +1,73 @@
...
+		.get('http://localhost:9000/_build/common/example/?module=button')
+		.setFindTimeout(5000);
+}
+
+registerSuite({
+	name: 'Button',
+
+	'button should be visible'(this: any) {
+		return getPage(this.remote)
+			.findByCssSelector('button:first-of-type')
```
tomdye: Please change all `findByCssSelector` calls to use the generate `css-module` class names by importing the appropriate `m.css` file.

src/button/tests/functional/Button.ts
```diff
@@ -0,0 +1,73 @@
...
+		// `isDisplayed` is broken in Safari 10, commented out for now. See https://github.com/SeleniumHQ/selenium/issues/3029
```
tomdye: Rather than leaving this in, can we please remove it in a separate commit? Then it can be added back in again as and when it's appropriate to do so. I think it's better not to have commented out code in our master branch.

src/button/tests/functional/Button.ts
lzhoucs: Thanks for the review Tom. I looked at https://github.com/dojo/widgets/pull/166/commits/6612379717688a6b3300fccd97d2546c32233846 where we changed to use `css-module` class to access the internals of a widget, and this approach makes sense to me. However, in my case here where the widget structure is simple and I am not trying to exercise the internal components of the widget, but the widget itself - I am trying to locate the first widget on the page (there's several button widgets in the example), so I am not sure if `css-module` class would be any help locating what I am looking for. For example, by using `css-module`, it seems I still have to do something like this: ``.findByCssSelector(`button.${css.root}`:first-of-type)``. That said, the example page probably should have included ids for each widget for external tests like this one.

tomdye: I see what you are saying @lzhoucs, however you are assuming that the tag will be `button`. This is not an assumption that you can make in all cases. More importantly, these tests are also to help with dojo 2 adopters and it's important that we consistently adopt `the right way` to deliver our code and tests. As per our `css-modules` approach, no css selector at any point should ever reference the tagname.

src/button/tests/functional/Button.ts
lzhoucs: Dylan suggested a slightly different approach: https://github.com/dojo/widgets/pull/168#discussion_r116070666, so that's why the comment is there. I've removed the commented code, however I do have a concern that it might get lost over time once the issue with Safari is fixed.

src/button/tests/functional/Button.ts
lzhoucs: I agree that tag name shouldn't be used. I've added a wrapper div with id for each widget in the example, together with `css-module` class to locate the widget. Please take another look when you get chance, @tomdye, thanks!

tomdye: Much better, although I'd have used a class over an id. 😉

lzhoucs: I think id is not always bad. In the example page, each example is unique. There will always be one `example-1` on the page. And I am not trying to consistently style multiple examples, but access each of them individually.

lzhoucs:
@tomdye , thanks for the review. I've addressed your feedback. Please take another look when you get chance and let me know if there's any other issues. Thanks!

PR #169, opened 2017-05-10
Fix slidepane animations with custom theme.

### LINKED ISSUE #129: Slidepane example is broken when the dojo theme is applied
**Bug** 

Package Version: master

**Description**
After the theme split work was completed the slidepane example page does not work correctly when the dojo theme is applied. The pane does not drag open / closed properly.

mwistrand:
**Type:** bug

**Description:**

Remove direct DOM manipulation from `SlidePane` so that its animations and functionality do not depend directly on `src/slidepane/styles/slidePane.m.css`.

Previously, `SlidePane` treated the `open`, `left`, and `right` CSS classes as fixed classes. However, this made it impossible for themes to specify combinations like `left.slideIn`, since the corresponding `left` class in the generated CSS is different than the fixed class actually applied to the DOM node. As such, `open`, `left`, and `right` were downgraded to normal classes.

Resolves #129

### DIFF
--- a/src/slidepane/SlidePane.ts
+++ b/src/slidepane/SlidePane.ts
@@ -93,7 +93,7 @@ export default class SlidePane extends SlidePaneBase<SlidePaneProperties> {
 	}
 
 	private _onSwipeEnd(event: MouseEvent & TouchEvent) {
-		const { target, type } = event;
+		const { changedTouches, pageX, target, type } = event;
 		this._swiping = false;
 
 		const {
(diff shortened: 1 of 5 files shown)


mwistrand:
@tomdye That would work for the slidepane not closing when the underlay is clicked; I'll update to use a fixed class for that since I guess it is not important/necessary that consumers be able to modify that class. The other changes allow the themed SlidePane to animate opened/closed properly.

tomdye:
Thanks for the clarification @mwistrand . I think that's the better way to go about it.

src/slidepane/SlidePane.ts
```diff
@@ -101,7 +103,7 @@ export default class SlidePane extends SlidePaneBase<SlidePaneProperties> {
 		} = this.properties;
 
 		// Current pointer position
-		const currentX = event.type === 'touchend' ? event.changedTouches[0].screenX : event.pageX;
+		const currentX = type === 'touchend' ? event.changedTouches[0].screenX : event.pageX;
```
bitpshr: If you're going to grab `target` and `type`, might as well destructure fully and grab `changedTouches` too.

PR #174, opened 2017-05-12
Remove bind references in v() and w() calls

agubler:
**Type:** feature

**Description:**

Prepping for the next release of `widget-core` that removes `bind` from both the `v()` and `w()` property interfaces. `bind` will automatically be added to all nodes from a `render` call using an `afterRender` in `WidgetBase`.

This means there is no way to pass `bind` when trying to assert with test-extras, so https://github.com/dojo/test-extras/issues/38 will need to be addressed first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,7 @@
     "@dojo/has": "beta1",
     "@dojo/i18n": "beta1",
     "@dojo/shim": "beta1",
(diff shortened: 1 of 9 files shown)


agubler:
Bombed out because needs release of widget-core and test-extras :)

PR #177, opened 2017-05-15
Icon css file with Select and ComboBox usage

### LINKED ISSUE #54: Add font awesome for icons and create icon.css
Tasks:

- [ ] add fa to our codebase
- [ ] implement in at least 1 widget

Questions:
Where should we implement an icon to test the integration?
Do we need to replace the down and up arrows on `ComboBox` and `TitlePane` or keep as text content?
Should we use as few icons as possible or use in as many places as possible for a consistent look and feel?
We could always add 'type' to `Dialog` e.g. `error|warn|info` and have nice icons by the dialog content header..

tomdye:
**Type:** feature

**Description:**

- Adds an `icons.m.css` file containing an icomoon app generate subset of the font-awesome icons.
- Added sample usage for `select` and `combobox` widgets as well as aligning their classname usage for consistency and predictability.
- Adds copy font tasks to grunt file

Resolves #54


src/combobox/ComboBox.ts
```diff
@@ -385,7 +387,7 @@ export default class ComboBox extends ComboBoxBase<ComboBoxProperties> {
 			clearable ? v('button', {
 				'aria-controls': menuId,
 				bind: this,
-				classes: this.classes(css.clear),
+				classes: this.classes(css.clear, iconCss.icon, iconCss.times),
```
bitpshr: Is it really necessary to pass `icon` every time we pass a specific icon class, like `times`? Can't we write the CSS in a way where this isn't needed?

src/common/styles/icons.m.css
```diff
@@ -0,0 +1,148 @@
...
+.plus:before {
+	content: "\f067";
+}
+.minus:before {
+	content: "\f068";
+}
+.check:before {
+	content: "\f00c";
+}
+.close:before {
```
bitpshr: How does this work with themes? It would be nice to show different icons when the theme check box is toggled. I feel like that would really cover all aspects of theming.

src/combobox/ComboBox.ts
smhigley: `composes` should do that, right?

src/combobox/ComboBox.ts
```diff
@@ -394,7 +396,7 @@ export default class ComboBox extends ComboBoxBase<ComboBoxProperties> {
 			v('button', {
 				'aria-controls': menuId,
 				bind: this,
-				classes: this.classes(css.arrow),
+				classes: this.classes(css.trigger, iconCss.icon, iconCss.chevronDown),
```
smhigley: Should we include such specific class names within the widget code? This could get semantically awkward if someone wants to style the button with a "+"-icon or something. I'm not sure of the best way to get around it, but my first thought is to leave `iconCss` out of `ComboBox` and handle it within the css modules using variables and imports.

src/common/styles/icons.m.css
```diff
@@ -0,0 +1,148 @@
+@font-face {
+	font-family: 'dojo2';
+	src:
+		url('fonts/dojo2.ttf?k76fx7') format('truetype'),
+		url('fonts/dojo2.woff?k76fx7') format('woff'),
```
smhigley: Can we get woff2 icon file? It's getting better support, and there are some online converters if whatever we're using to generate the font files doesn't output it.

src/combobox/ComboBox.ts
tomdye: yes and no, so `composes` *can* do that, BUT it creates other problems and duplicated code. The `icons.css` file is adapted from the `icomoon` output. Adding `composes` or similar would mean further changes to the standard output and the actual output would just be two classes anyway.

src/common/styles/icons.m.css
tomdye: So when I worked on the ill-fated `icon widget` i showed that this works with theming. I can add a material icon version in the `dojo` theme for clarity if we feel this is particularly important.

src/common/styles/icons.m.css
tomdye: Again, this is the limited output available from icomoon. I think that if we have a premium account then we might get access to the `woff2` but I'm not sure. Might be worth finding out if they have an open source offering outside of their paid model. @dylans any idea here?

src/combobox/ComboBox.ts
tomdye: I think this is a pretty common way of doing things. If we left `iconCss` out then the icons would not be themeable so I don't think that is an issue. I believe we made a decision to use an icon font to provide icons as part of the original issue and this way we have the css-modules providing the icon type-ahead etc... It's quick and easy for a user to see what icons are available to them.

smhigley: I definitely like the intellisense that comes with `iconCss`, and I think it should be used like this when users are creating their own VNodes. My only concern is for edge cases within themed widgets. The first is purely semantic: if someone wants to change one icon to a different icon, they could theoretically theme `iconCss` so `.chevronDown` is a plus sign, but it seems messy. The other problem is if anyone ever wants to theme two separate uses of the same icon differently. Is that possible with the current usage?

src/common/styles/icons.m.css
bitpshr: Premium does expose woff2 files fwiw.

dylans:
This one needs a rebase @tomdye

tomdye:
I'm going to update the icon class names as ensure that the `icomoon` selections file is put somewhere too such that anyone can regenerate the css. Also, as per https://github.com/dojo/widget-core/issues/538 I am going to suffix icon classnames with `Icon` to avoid such issues going forward.

src/common/styles/fonts/dojo2.svg
```diff
@@ -0,0 +1,26 @@
...
+<glyph unicode="&#xf078;" glyph-name="down" d="M961.714 416l-424-423.429c-14.286-14.286-37.143-14.286-51.429 0l-424 423.429c-14.286 14.286-14.286 37.714 0 52l94.857 94.286c14.286 14.286 37.143 14.286 51.429 0l303.429-303.429 303.429 303.429c14.286 14.286 37.143 14.286 51.429 0l94.857-94.286c14.286-14.286 14.286-37.714 0-52z" />
+<glyph unicode="&#xf0c9;" glyph-name="bars" horiz-adv-x="878" d="M877.714 109.714v-73.143c0-20-16.571-36.571-36.571-36.571h-804.571c-20 0-36.571 16.571-36.571 36.571v73.143c0 20 16.571 36.571 36.571 36.571h804.571c20 0 36.571-16.571 36.571-36.571zM877.714 402.285v-73.143c0-20-16.571-36.571-36.571-36.571h-804.571c-20 0-36.571 16.571-36.571 36.571v73.143c0 20 16.571 36.571 36.571 36.571h804.571c20 0 36.571-16.571 36.571-36.571zM877.714 694.857v-73.143c0-20-16.571-36.571-36.571-36.571h-804.571c-20 0-36.571 16.571-36.571 36.571v73.143c0 20 16.571 36.571 36.571 36.571h804.571c20 0 36.571-16.571 36.571-36.571z" />
+</font></defs></svg>
```
agubler: missing line at end of the file

src/common/styles/fonts/dojo2Selection.json
```diff
@@ -0,0 +1,485 @@
...
+      "useClassSelector": true,
+      "color": 0,
+      "bgColor": 16777215,
+      "classSelector": ".icon"
+    },
+    "historySize": 50,
+    "showCodes": true,
+    "gridSize": 16
+  }
+}
```
agubler: missing line at end of the file

tomdye: This is a generated file @agubler

smhigley:
One other thing -- could you add `role="presentation"` and `aria-hidden="true"` to the icon elements? Without them (more the second), screen readers still sometimes read css `content`.

PR #178, opened 2017-05-17
Add `Slider` functional tests

### LINKED ISSUE #158: Slider: Add Functional Tests

lzhoucs:
**Type:** functional tests

Resolves #158

### DIFF
--- a/src/slider/tests/functional/Slider.ts
+++ b/src/slider/tests/functional/Slider.ts
@@ -3,8 +3,12 @@ import * as assert from 'intern/chai!assert';
 import * as css from '../../styles/slider.m.css';
 import * as keys from 'leadfoot/keys';
 
-function getPage(remote: any) {
-	return remote
+function getPage(test: any) {
+	const { browserName } = test.remote.environmentType;
+	if (browserName.toLowerCase() === 'microsoftedge') {
+		test.skip('example page currently doesn\'t work in edge.');
+	}
+	return test.remote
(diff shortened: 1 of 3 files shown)


lzhoucs:
all tests are failing in edge. I saw the same issue locally and I thought it was a browserstack issue. 
 But it seems it has to do with my tests. Will look into it asap.

src/slider/tests/functional/Slider.ts
```diff
@@ -0,0 +1,255 @@
...
+			.findByCssSelector(`.${css.input}`)
+			.getProperty('value')
+			.then((value: string) => {
+				currentValue = parseInt(value, 10);
+			})
+			.end()
+			.findByCssSelector(`.${css.fill}`)
+			.getAttribute('style')
+			.then((style: string) => {
+				const absWidthRegex = /width:\s*(\d+)\.?\d*%/;
```
tomdye: can this not be done via a property or similar? regex seems odd.

lzhoucs: I agree. I tried the property approach before coming up with this but it didn't work. I used `.getProperty('style')` but what got returned is an array of property names(`['width']`) in chrome, instead of an object with which I can then use `style.width`. If you point me to an example on how this can be done with property, I'd love to make the change.

lzhoucs:
The tests were working originally but started failing after recent rebase. Needs to look into it.

pottedmeat:
Moved to https://github.com/dojo/widgets/pull/322

PR #198, opened 2017-05-22
Fix: clear SlidePane transform on reopen.

mwistrand:
**Type:** bug

**Description:**

After reopening a `SlidePane`, the CSS `transform` property should be reset to avoid interfering with animations.


mwistrand:
@bitpshr There is a failing test case on master. I merged two PRs this morning that passed independently, but fail together.

PR #203, opened 2017-05-25
Fix focus styles

smhigley:
**Type:** bug
Some focus styles had been removed without replacement in recent theming shuffles.


mwistrand:
The only thing I see missing is the focus style for the close button on a `TabPane` tab. Other than that, all looks good to me.

smhigley:
Good catch!

PR #215, opened 2017-06-01
added fixed classes

### LINKED ISSUE #214: Move purely structural widget styles to use `fixed`
**Enhancement** 

To better support widget theming we should move certain styles / classes to use `this.classes().fixed()` instead.
This avoids the need to replicate structural styles in theme css.

This issue should be used whenever such work is carried out.

tomdye:
**Type:** feature

**Description:**

Creates `fixed` classes for `Slider` to ease future theming.

Related to  #214


src/slider/Slider.ts
```diff
@@ -126,7 +126,7 @@ export default class Slider extends SliderBase<SliderProperties> {
 		const outputNode = output ? output(value) : value + '';
 
 		const slider = v('div', {
-			classes: this.classes(css.inputWrapper),
+			classes: this.classes().fixed(css.inputWrapper),
```
bitpshr: I've been wondering about situations like this, as I've run into them pretty frequently: should we always provide a themeable class on node's that have a fixed class? It seems odd that our components' themeable classes are only determined by what our default theme necessitates. What if some new theme needs to target `inputWrapper` to satisfy their mock ups, but can't simply because our default theme didn't need a themeable class there?

I don't think this should hold up the PR at all, and I think a pattern will become obvious from community usage and after our own initial theme is done.

tomdye:
@bitpshr The issue is going to remain open to include refactors as and where required. If someone wants to take the time and go through to split these out then then can go for it at some point :)

src/slider/Slider.ts
tomdye: I think we should only add a themeable class then there's something to be themed. Ie `inputWrapper` does not need a themeable class in this case.

src/slider/styles/slider.m.css
```diff
@@ -31,11 +41,14 @@
 	height: 2em;
 	left: 50%;
 	margin-left: -1em;
-	position: absolute;
 	top: -0.625em;
 	width: 2em;
```
smhigley: Should dimensions be under the fixed class? The size of the track and thumb seems like one of the main things someone might want to theme.

tomdye: they're not in the fixed class....

smhigley: Hah, you're totally right. Sorry! Sneaky line collapsing fooled me 😆

PR #221, opened 2017-06-05
Remove unnecessary type casting in examples

### LINKED ISSUE #217: Remove unnecessary state type casting in examples
**Enhancement**

Earlier versions of widget-core required that `state` be type casted before being passed into child components. This should be removed from all component example pages.

Package Version: latest

**Code**

```ts
w(Button, {
	pressed: <boolean> this.state.buttonPressed
})
```

bitpshr:
**Type:** feature

**Description:**

This PR removes once-necessary type casting from examples. The issue was evident when grabbing values from state and passing them to child components as properties.

Resolves #217

### DIFF
--- a/src/select/example/index.ts
+++ b/src/select/example/index.ts
@@ -110,7 +110,9 @@ export class App extends AppBase<WidgetProperties> {
 				useNativeElement: true,
 				value: this.state.value1,
 				theme: this._theme,
-				onChange: (option: OptionData) => this.setState({ value1: option.value })
+				onChange: (option: OptionData) => {
+					this.setState({ value1: option.value });
+				}
 			}),
 			v('p', {
 				innerHTML: 'Result value: ' + this.state.value1
@@ -123,7 +125,9 @@ export class App extends AppBase<WidgetProperties> {
 				options: this._selectOptions,
 				value: this.state.value2,
 				theme: this._theme,
-				onChange: (option: OptionData) => this.setState({ value2: option.value })
+				onChange: (option: OptionData) => {
+					this.setState({ value2: option.value });
+				}
 			}),
 			v('h2', {}, [ 'Native multiselect widget' ]),
 			w(Select, {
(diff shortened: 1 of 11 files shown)


src/slidepane/example/index.ts
```diff
@@ -32,12 +32,10 @@ export class App extends StatefulMixin(WidgetBase)<WidgetProperties> {
...
-				open: <boolean> this.state['open'],
-				underlay: <boolean> this.state['underlay'],
-				align: <Align> this.state['align'],
-				onRequestClose: () => {
-					this.setState({ open: false });
-				},
+				open: this.state.open,
+				underlay: this.state.underlay,
+				align: this.state.align,
+				onRequestClose: () => this.setState({ open: false }),
```
agubler: this doesn't actually return anything so shouldn't be inline

agubler: I see it is done elsewhere also, I think the style guide also says that if there is no return then the implicit return shouldn't be used.

src/select/example/index.ts
```diff
@@ -108,26 +108,22 @@ export class App extends AppBase<WidgetProperties> {
...
 				customOption: CustomOption,
 				label: 'Custom box!',
 				options: this._selectOptions,
-				value: <string> this.state['value2'],
+				value: this.state.value2,
 				theme: this._theme,
-				onChange: (option: OptionData) => {
-					this.setState({ value2: option.value });
-				}
+				onChange: (option: OptionData) => this.setState({ value2: option.value })
```
smhigley: Same comment as @agubler made below about not using implicit return

src/textinput/example/index.ts
```diff
@@ -31,49 +31,49 @@ export class App extends AppBase<WidgetProperties> {
 				w(TextInput, {
 					key: 't1',
 					label: 'Name',
-					type: <TextInputType> 'text',
+					type: 'text',
```
smhigley: Do things like `ButtonType` and `TextInputType` still need to be exported now that we don't need to do this?

bitpshr: I think they should still be made available to downstream users.

bitpshr:
@smhigley @agubler updated all examples to not implicitly return.

PR #222, opened 2017-06-05
Fix registry creation in widgets

smhigley:
**Type:** bug
Since `onPropertiesChanged` won't fire on initial render if there are no properties passed, registry creation should happen in the constructor. Adds custom `@diffProperty` decorators for now, until https://github.com/dojo/widget-core/issues/567 is resolved.


bitpshr:
Looks fine. Did coverage take a big hit?

smhigley:
I'm not sure I really understand the coverage diff. It shows coverage went up (as it should have, by eliminating the previously uncoverable branch), but the diff is too low?

PR #232, opened 2017-06-08
TitlePane (and README, test-extras, functional tests)

bitpshr:
**Type:** feature

**Description:**

This PR builds on the work @msssk did on the TitlePane to bring it to completion based on all feedback to date.

Resolves nothing (no issue for TitlePane?) but works towards resolving #14.


src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,132 @@
...
+
+	private _onTitleClick() {
+		this._toggle();
+	}
+
+	private _onTitleKeyUp(event: KeyboardEvent) {
+		const { closeable = true } = this.properties;
+		if (!closeable) {
+			return;
+		}
```
smhigley: You already have this logic in `_toggle()`, not sure if it's necessary here too

src/titlepane/example/index.ts
```diff
@@ -0,0 +1,123 @@
...
+import { WidgetProperties } from '@dojo/widget-core/interfaces';
+
+import TitlePane from '../TitlePane';
+import dojoTheme from '../../themes/dojo/theme';
+
+export class App extends StatefulMixin(WidgetBase)<WidgetProperties> {
+	private _theme: {};
+
+	themeChange(event: Event) {
+		const checked = (<HTMLInputElement> event.target).checked;
```
smhigley: can type as `event: TypedTargetEvent<HTMLInputElement>`, and then you don't need to cast `event.target`

src/titlepane/README.md
```diff
@@ -0,0 +1,59 @@
...
+`TitlePane` supports standard keyboard navigation for toggling the content open or closed.
+
+**TitleBar Events**:
+
+- Space bar: toggles the pane content of a closeable `TitlePane`
+
+### Accessibility Features
+
+- The titlebar and aria-labelledby have screen-reader-accessible labels and instructions
+- The text within a closeable titlebar is contained within a `<button>` with screen-reader-accessible instructive text
```
smhigley: I don't think this is entirely accurate, since we're using `role="button"` and not `<button>` (I'm still not entirely sure why we don't just use `<button>`, actually). If I were to write the a11y features, I might do something like this:
```markdown
- The titlebar has a role of heading and accepts a `headingLevel` property to set the correct heading level
- The accordion panel has `aria-labelledby` set to the titlebar
- The accordion opens on Enter and Space
- `aria-expanded` is set to `true` or `false` on the toggle button depending on `properties.open`
- `aria-controls` on the toggle button points to the accordion content
```

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,127 @@
...
+					id: this._titleId,
+					role: 'button',
+					tabIndex: closeable ? 0 : -1
+				}, [ title ])
+			]),
+			v('div', {
+				'aria-labelledby': this._titleId,
+				classes: this.classes(css.content),
+				id: this._contentId,
+				key: 'content'
```
smhigley: Sorry, I missed something last time :P. Could we add `aria-hidden: open ? null : 'true'` to this?

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,127 @@
...
+	private _titleId = uuid();
+
+	private _afterRender(element: HTMLElement) {
+		// Conditionally adjust top margin. Done manually instead of through Maquette
+		// so the underlying DOM is accessible, as we need to know the content height.
+		// Put in a rAF to push this operation to the next tick, otherwise
+		// element.offsetHeight can be incorrect (e.g. before styling is applied)
+		// Note that this will go away when meta support is added to widget-core
+		requestAnimationFrame(() => {
+			const { open = true } = this.properties;
```
tomdye: is this still needed as you're using `max-height` for the animation?

src/titlepane/TitlePane.ts
```diff
@@ -0,0 +1,127 @@
...
+			open = true,
+			title
+		} = this.properties;
+
+		return v('div', {
+			classes: this.classes(css.root)
+		}, [
+			v('div', {
+				'aria-level': headingLevel ? String(headingLevel) : null,
+				classes: this.classes(css.title, closeable ? css.closeable : null),
```
tomdye: for things like the closeable clickable area you might want to add a fixed class for `cursor: pointer`.

src/titlepane/styles/titlePane.m.css
```diff
@@ -0,0 +1,24 @@
+@import '../../common/styles/variables.css';
+
+.root {
+	overflow: hidden;
```
tomdye: i think that most of these should have `fixed` counterparts

src/titlepane/tests/functional/TitlePane.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+const DELAY = 400;
+
+registerSuite({
+	name: 'TitlePane',
+
+	'Should be fully visible when `open`'(this: any) {
+		return getPage(this.remote)
+			.findByCssSelector('#titlePane2 > div > :last-child')
```
tomdye: can you please import the `css` file for `titlepane` and use that to target nodes in `findByCssSelector`.

src/titlepane/tests/unit/TitlePane.ts
```diff
@@ -0,0 +1,270 @@
...
+			closeable: true,
+			open: false,
+			onRequestOpen() {
+				called = true;
+			},
+			title: 'test'
+		});
+
+		titlePane.sendEvent('click', {
+			selector: '[role="heading"]'
```
tomdye: I think that these selectors can be replaced with css-classes as well

src/themes/dojo/titlePane.m.css
```diff
@@ -0,0 +1,34 @@
...
+	border-width: var(--border-thin);
+	color: var(--title-bar-color);
+	font-size: var(--title-bar-font-size);
+	padding: var(--title-padding);
+	position: relative;
+	z-index: 1;
+}
+
+.closeable {
+	cursor: pointer;
```
bitpshr: It is, but since the theme replaces the base CSS entirely, we also have to put it here.

src/themes/dojo/variables.css
```diff
@@ -40,6 +40,7 @@
 	/* Composed values */
 	--title-bar-bg: var(--dojo-blue);
 	--title-bar-color: var(--dojo-white);
+	--title-padding: 5px;
```
bitpshr: Nothing exists yet beyond `--container-padding`. I renamed this to `--header-padding` to be more generic.

src/titlepane/TitlePane.ts
bitpshr: No, I'm not using `max-height` for the animation. We're using `margin` so we could avoid the bugs that plagued using `max-height`, it was a trick I discovered while working on this a while ago. The animations you mention were leftover and have been removed from this PR.

src/themes/dojo/titlePane.m.css
bitpshr: Disregard, added a fixed class as mentioned below.

src/themes/dojo/variables.css
tomdye: thanks

src/titlepane/TitlePane.ts
bitpshr: Done.

src/common/tests/unit/main.ts
```diff
@@ -9,15 +9,16 @@ registerSuite({
...
 		assert.isDefined(widgets.Label);
 		assert.isDefined(widgets.Radio);
 		assert.isDefined(widgets.Select);
 		assert.isDefined(widgets.SlidePane);
 		assert.isDefined(widgets.Slider);
-		assert.isDefined(widgets.SplitPane);
 		assert.isDefined(widgets.TabPane);
 		assert.isDefined(widgets.Textarea);
 		assert.isDefined(widgets.TextInput);
+		assert.isDefined(widgets.TitlePane);
```
bitpshr: Fixed.

src/titlepane/tests/functional/TitlePane.ts
bitpshr: Done.

tomdye:
good good

PR #245, opened 2017-07-05
Use sinon throughout unit tests

### LINKED ISSUE #234: Use sinon for test assertions rather than `called` booleans
**Enhancement** 

In order to provide easy to read test assertions and a consistent approach to unit tests we should look to leverage sinon `spy` and `stub` where appropriate. This will involve revisiting existing unit tests for `widgets`.

bitpshr:
**Type:** feature

**Description:**

This PR updates all component unit tests to use sinon for mocking property callbacks instead of manually setting and asserting a variable.

Resolves #234

### DIFF
--- a/src/button/tests/unit/Button.ts
+++ b/src/button/tests/unit/Button.ts
@@ -1,6 +1,7 @@
 import * as registerSuite from 'intern!object';
 import * as assert from 'intern/chai!assert';
 import * as Test from 'intern/lib/Test';
+import * as sinon from 'sinon';
 
 import has from '@dojo/has/has';
 import harness, { Harness } from '@dojo/test-extras/harness';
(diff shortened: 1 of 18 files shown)


src/button/tests/unit/Button.ts
```diff
@@ -124,42 +125,42 @@ registerSuite({
...
+			onKeyDown,
+			onKeyPress,
+			onKeyUp,
+			onMouseDown,
+			onMouseUp
 		});
 
 		widget.sendEvent('blur');
-		assert.isTrue(blurred);
+		assert.isTrue(onBlur.called);
```
tomdye: Can we change these to `calledOnce`, `calledTwice` etc please? Simply using `called` won't catch errors where events are triggered multiple times by mistake.

src/button/tests/unit/Button.ts
```diff
@@ -124,42 +125,42 @@ registerSuite({
...
 	events() {
-		let blurred = false;
-		let clicked = false;
-		let focused = false;
-		let keydown = false;
-		let keypress = false;
-		let keyup = false;
-		let mousedown = false;
-		let mouseup = false;
+		const onBlur = sinon.spy();
```
tomdye: I think that a `stub` would be more appropriate here as a `spy` is meant to allow the function to run underneath and simply monitor the args etc that it was called with.

src/calendar/tests/unit/CalendarCell.ts
```diff
@@ -91,41 +92,31 @@ registerSuite({
 	},
```
smhigley: For `'Click handler called with correct arguments'()`, could `onClick` be a spy that tests the arguments passed instead of using `clickedDate` and `clickedDisabled`?

src/calendar/tests/unit/MonthPicker.ts
```diff
@@ -306,11 +307,11 @@ registerSuite({
```
smhigley: Could the `'Popup opens and closes on button click'()`, `'Popup closes with correct keys'()`, `'Arrow keys change year spinner'()`, and `'Clicking buttons changes year'()` use `sinon.stub` or `.spy`?

src/calendar/tests/unit/MonthPicker.ts
```diff
@@ -306,11 +307,11 @@ registerSuite({
...
 		let currentMonth = testDate.getMonth();
-		let closed = false;
+		const onRequestClose = sinon.spy();
 		widget.setProperties({
 			...requiredProps,
-			open: !closed,
-			onRequestClose: () => { closed = true; },
+			open: true,
+			onRequestClose,
 			onRequestMonthChange: (month: number) => { currentMonth = month; }
```
smhigley: I think this can be a `sinon.spy` too, and then test `assert.strictEqual(onRequestMonthChange.args[0], 6)` (I think). Same for the `'Previous/next month buttons'()` test

kitsonk:
We should either resolve the conflicts in the PR or close it.

bitpshr:
This PR is no longer relevant. All tests have since been rewritten, most of which use sinon appropriately.

PR #248, opened 2017-07-11
Theme: SlidePane

bitpshr:
**Type:** feature

**Description:**

This PR adds an initial theme for the SlidePane component.

screen shot 2017-07-11 at 4 18 24 pm


src/themes/dojo/slidePane.m.css
```diff
@@ -1,34 +1,91 @@
...
 .underlayVisible {
 	background: var(--underlay-bg);
 }
 
-.content {
-	font: var(--base-font);
+.pane {
+	background: var(--dojo-white);
+	border: var(--border-thin) solid var(--border-color);
+	box-shadow: var(--drop-shadow);
```
smhigley: I used `--box-shadow-dimensions` for use with custom colors for inputs, but the shadow dimensions were different for those. We should probably think of better naming 😝

src/themes/dojo/slidePane.m.css
```diff
@@ -1,34 +1,91 @@
...
+	font-size: 0;
+	padding: var(--grid-size);
+	position: absolute;
+	right: calc(2 * var(--grid-size));
+	top: 50%;
+	transform: translateY(-50%);
+}
+
+.close:after {
+	content: '✕';
```
smhigley: doesn't need to be now, but at some point this should probably be an icon (I think it is in the designs)

src/themes/dojo/variables.css
```diff
@@ -55,4 +49,23 @@
...
 	--calendar-hover-bg: var(--dojo-light-blue);
 	--calendar-selected: var(--dojo-blue);
+
+	/* New theme variables */
+	--dojo-white: #FFFFFF;
+	--dojo-gray: #D6DDE2;
+	--dojo-light-gray: #F4F6F7;
+	--dojo-dark-gray: #5C6C7C;
+
+	--pane-bg: var(--dojo-white);
```
smhigley: I'm fine leaving out variable names that are this specific. I think if the name references a specific widget, it doesn't need to be a separate variable.

bitpshr: Agreed. I use `--pane-bg` in a few components so far, so I'll leave this in for now.

tomdye:
Can you please make sure that the outermost element (including the underlay) has a `root` class on it. Previously when I was trying to theme slidepane for a client project I had trouble increasing the z-index as this was not available. 
Also likely that you're going to need to add some `fixed` classes to maintain layout when this is themed.

bitpshr:
@tomdye: Good catch. SlidePane especially needs to use fixed classes to maintain even basic usability. I took care of this and updated the test harness expected renders.

PR #250, opened 2017-07-14
Select theme

smhigley:
**Type:** feature

**Description:**

Adds theme for Select widget:
screen shot 2017-07-14 at 11 17 08 am

### DIFF
--- a/src/select/Select.ts
+++ b/src/select/Select.ts
@@ -250,7 +250,7 @@ export default class Select extends SelectBase<SelectProperties> {
 		});
 	}
 
-	renderNativeSelect(): DNode {
+	protected renderNativeSelect(): DNode {
 		const {
 			describedBy,
 			disabled,
@@ -296,7 +296,7 @@ export default class Select extends SelectBase<SelectProperties> {
 		]);
 	}
 
-	renderCustomMultiSelect(): DNode {
+	protected renderCustomMultiSelect(): DNode {
 		const {
 			_focusedIndex
 		} = this;
@@ -327,7 +327,7 @@ export default class Select extends SelectBase<SelectProperties> {
 		]);
 	}
 
-	renderCustomSelect(): DNode {
+	protected renderCustomSelect(): DNode {
 		const {
 			describedBy,
 			disabled,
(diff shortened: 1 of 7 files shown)


src/select/Select.ts
```diff
@@ -250,7 +250,7 @@ export default class Select extends SelectBase<SelectProperties> {
 		});
 	}
 
-	renderNativeSelect(): DNode {
+	protected renderNativeSelect(): DNode {
```
bitpshr: I'm all for this change, but I'm not sure we should touch visibility as part of this theme-specific PR.

src/select/Select.ts
```diff
@@ -288,15 +288,15 @@ export default class Select extends SelectBase<SelectProperties> {
 				onfocus: this._onFocus,
 				onkeydown: this._onKeyDown
 			}, optionNodes),
-			v('span', { classes: this.classes(css.nativeArrow) }, [
+			multiple ? null : v('span', { classes: this.classes(css.arrow) }, [
```
bitpshr: Does this need a hidden text label? (I may be confusing this with buttons.)

src/themes/dojo/select.m.css
```diff
@@ -13,53 +15,41 @@
 /* custom input styles */
 .trigger {
 	background-color: transparent;
-	border: 3px solid var(--input-border-color);
+	border: var(--border-width) solid var(--color-border);
+	border-bottom-color: var(--color-border-strong);
 	font-size: inherit;
+	line-height: var(--line-height-base);
 	min-width: 200px;
```
bitpshr: Why the min width? Also, are we making all inputs take up 100% of their parent, but giving them a uniform `max-width`?

src/select/Select.ts
smhigley: It's a `<select>` element so it's labeled by its `<label>`. The arrow here is just extra visual window dressing :)

src/select/Select.ts
smhigley: I think it was already visible, just missing the `protected` keyword. The other similar fn's have it, so I thought I'd sneak it in :P

src/themes/dojo/select.m.css
smhigley: I forgot, and let it sneak in. Good catch! I believe we're letting all inputs take up 100% of the parent, with no `max-width`

PR #254, opened 2017-07-17
Radio Theme

smhigley:
**Type:** feature

**Description:**

This PR adds a base theme for the radio widget. In doing so, a `.focused` class was needed on the root node. It seems a common enough styling case to add it for at least the radio and checkbox widgets.

### DIFF
--- a/src/radio/tests/unit/Radio.ts
+++ b/src/radio/tests/unit/Radio.ts
@@ -98,6 +98,21 @@ registerSuite({
 		assert.isFalse(vnode.properties!.classes![css.invalid]);
 	},
 
+	'focused class'() {
+		const radio = new Radio();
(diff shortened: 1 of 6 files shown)


src/radio/tests/unit/Radio.ts
```diff
@@ -98,6 +98,21 @@ registerSuite({
 		assert.isFalse(vnode.properties!.classes![css.invalid]);
 	},
 
+	'focused class'() {
+		const radio = new Radio();
+		let vnode = <VNode> radio.__render__();
+
+		assert.isUndefined(vnode.properties!.classes![css.focused]);
+
+		(<any> radio)._onFocus(<FocusEvent> {});
```
tomdye: is it possible to perform these tests without calling the private function?

PR #256, opened 2017-07-18
Add icons for themes

smhigley:
**Type:** bug / feature

**Description:**

Adds themed icons for the base theme, and updates the available icon classes in `src/common/styles`.


tomdye:
Only thought is that we might want to include any font files under the `theme` dir rather than limiting to the dojo theme, there will be others in the future.

smhigley:
Do you think the other themes will use Material icons too? If we think that will be pretty standard, I'm also in favor of moving them to `/themes/fonts`. I do also want to make sure it's easy to drop themes into projects. Will the CLI handle copying the correct theme css + fonts, so users won't need to remember they need `themes/fonts` in addition to e.g. `themes/dojo`?

smhigley:
I'm going to merge and close this PR. For now, let's leave the icons within the theme, and revisit when we have a second theme.

PR #257, opened 2017-07-19
Slider Theme

smhigley:
**Type:** feature

**Description:**
screen shot 2017-07-19 at 1 34 22 am

Adds base theme for the slider widget. Modified the widget to include a boolean property to create a tooltip tied to the slider handle.

Also added body padding to examples.

### DIFF
--- a/src/slider/Slider.ts
+++ b/src/slider/Slider.ts
@@ -50,6 +50,7 @@ export interface SliderProperties extends ThemeableProperties {
 	min?: number;
 	name?: string;
 	output?(value: number): DNode;
+	outputIsTooltip?: boolean;
 	readOnly?: boolean;
 	required?: boolean;
 	step?: number;
@@ -103,6 +104,7 @@ export default class Slider extends SliderBase<SliderProperties> {
 			min = 0,
 			name,
 			output,
+			outputIsTooltip = false,
 			readOnly,
 			required,
 			step = 1,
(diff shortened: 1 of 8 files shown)


src/slider/Slider.ts
```diff
@@ -174,8 +182,9 @@ export default class Slider extends SliderBase<SliderProperties> {
 				})
 			]),
 			v('output', {
-				classes: this.classes(css.output),
-				for: this._inputId + ''
+				classes: this.classes(css.output, outputIsTooltip ? css.outputTooltip : null),
+				for: this._inputId + '',
```
tomdye: not a big fan of the `+ ''` approach for casting to a string. Would it not be better to use a template string? ie. 
```
for: `${this._inputId}`
```

bitpshr: @smhigley: Looks like @tomdye is right, template strings are [fastest now](https://jsperf.com/es6-string-literals-vs-string-concatenation) (previously String concatenation was still fastest; something must've been improved.)

src/slider/Slider.ts
```diff
@@ -125,6 +127,12 @@ export default class Slider extends SliderBase<SliderProperties> {
 		// custom output node
 		const outputNode = output ? output(value) : value + '';
 
+		// output styles
+		let outputStyles: { left?: string; top?: string } = {};
+		if (outputIsTooltip) {
+			outputStyles = vertical ? { top: ((max - value) / max * 100) + '%' } : { left: (value / max * 100) + '%' };
```
tomdye: Can you use a template string for this? Might look cleaner than `+ '%'`

PR #258, opened 2017-07-19
Theme: ComboBox

bitpshr:
**Type:** feature

**Description:**

This PR adds an initial theme for the ComboBox component. Note that this work was based off @smhigley's TextInput theme PR.

screen shot 2017-07-19 at 10 27 26 am


src/themes/dojo/comboBox.m.css
```diff
@@ -1,55 +1,103 @@
...
+	border: var(--border-width) solid var(--color-border);
+	box-shadow: var(--box-shadow-dimensions-small) var(--color-box-shadow);
+	opacity: 0;
+	position: absolute;
+	transform-origin: 0 0;
+	transform: rotateX(90deg);
+	transition: all var(--transition-duration) var(--transition-easing);
 	width: 100%;
-	padding-right: 2em;
+	z-index: 2;
```
smhigley: We could even put them in variables 😁

src/themes/dojo/comboBox.m.css
```diff
@@ -1,55 +1,103 @@
...
+	background: none;
+	border: none;
+	color: var(--color-border);
+	cursor: pointer;
+	font-size: 0;
+	padding: 0;
+	position: absolute;
+	right: calc(var(--spacing-regular) * 6);
+	top: 50%;
+	transform: translateY(-50%);
```
smhigley: Could possibly use a focus style too

src/themes/dojo/comboBox.m.css
```diff
@@ -1,55 +1,103 @@
...
-	z-index: 1;
+.selected {
+	background-color: var(--color-background-faded);
 }
 
-.result {
-	cursor: pointer;
-	padding: var(--input-padding);
+.invalid .trigger {
+	border-left-color: color(var(--color-error) saturation(-9%) lightness(+37%));
```
smhigley: This color (and the success variation) is getting used in more places than I thought. Do you think it's worth thinking of another variable name?

src/themes/dojo/comboBox.m.css
```diff
@@ -1,55 +1,103 @@
...
-.controls {
-	position: relative;
+.root input {
+	line-height: var(--line-height-base);
+	min-width: 200px;
+	padding: var(--spacing-regular) calc(var(--spacing-regular) * 6) var(--spacing-regular) var(--spacing-regular);
+	width: 100%;
+}
+
+.trigger {
```
tomdye: is it possible to move some of this styling to a shared file? Looks like combo / select / textbox have the same styling

src/themes/dojo/comboBox.m.css
bitpshr: Agreed, good catch. I will do this as part of a separate PR, this will effect SlidePane and Dialog (at least) once they land.

src/themes/dojo/comboBox.m.css
bitpshr: Added.

src/themes/dojo/comboBox.m.css
bitpshr: Totally agreed. We have dojo/widgets#117 to track that progress, as it's more involved than just pulling out common styling. The underlying DOM is still somewhat different and would require some consolidation.

I definitely think we need to take care of that ticket asap though.

bitpshr:
@tomdye @smhigley this should be ready for another review.

bitpshr:
Bump @smhigley and @tomdye for a review. Thanks!

src/combobox/example/index.ts
```diff
@@ -142,7 +142,9 @@ export class App extends WidgetBase<WidgetProperties> {
 	}
 
 	render(): DNode {
-		return v('div', [
+		return v('div', {
+			styles: { maxWidth: '256px' }
```
tomdye: i assume this is to restrict the width of the inputs, should they not have their own max width?

smhigley: Based on talking with Scott, I believe we decided no max width. It makes a certain sort of sense -- desired width can be really variable (no pun intended, mostly), and controlling input width through styles on the containing element seems like a better pattern anyway.

src/themes/dojo/comboBox.m.css
```diff
@@ -1,55 +1,115 @@
...
+	top: 50%;
+	transform: translateY(-50%);
 }
 
-.disabledResult {
-	color: var(--dojo-gray);
-	font-style: italic;
-	cursor: default;
+.clear:focus {
+	border-color: var(--color-highlight);
```
smhigley: I know I'm being super nitpicky (😔), but the border isn't the most beautiful of styles. Maybe Scott would have some ideas?

src/combobox/example/index.ts
smhigley: @bitpshr I actually wouldn't mind this added to textinput/textarea too, if you feel like it

src/themes/dojo/comboBox.m.css
bitpshr: Agreed. I will call this out for Scott!

PR #262, opened 2017-07-21
Slider: min / max value fix

### LINKED ISSUE #259: Slider can accept value larger than max and smaller than min
**Bug** 

When using the slider I have noticed that you can set it's value to much larger than it's supplied `max` prop. This causes the slider thumb to move outside of the widget itself.

bitpshr:
**Type:** bug

**Description:**

This PR updates the Slider component so it never renders a value higher than its `max` or lower than its `min`.

Resolves #259

### DIFF
--- a/src/slider/Slider.ts
+++ b/src/slider/Slider.ts
@@ -111,9 +111,12 @@ export default class Slider extends SliderBase<SliderProperties> {
 			vertical = false,
 			verticalHeight = '200px'
 		} = this.properties;
-		const {
+		let {
 			value = min
 		} = this.properties;
+
+		value = value > max ? max : value;
+		value = value < min ? min : value;
 
 		const stateClasses = [
 			disabled ? css.disabled : null,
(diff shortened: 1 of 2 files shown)


src/slider/Slider.ts
```diff
@@ -109,10 +109,13 @@ export default class Slider extends SliderBase<SliderProperties> {
 			vertical = false,
 			verticalHeight = '200px'
 		} = this.properties;
-		const {
+		let {
 			value = min
 		} = this.properties;
 
+		value = value > max ? max : value;
+		value = value < min ? min : value;
```
smhigley: This isn't a suggestion so much as a question: my initial instinct is towards `Math.max(value, min)` and `Math.min(value, max)` for readability. Just curious if there's a rationale for one vs. the other apart from personal preference?

tomdye: I used the following previously (taken from `npm/clamp`).

```
function clamp(value, min, max) {
  return min < max
    ? (value < min ? min : value > max ? max : value)
    : (value < max ? max : value > min ? min : value)
}
```

bitpshr: @smhigley just preference. Even if I used `Math.max` and `Math.min`, I'd want to check that the `value` was actually greater than the `max` or less than the `min` before calling them, and at that point, I might as well just use the `max` and `min` directly to avoid unnecessary indirection of a `Math` method.

smhigley: cool :)

PR #290, opened 2017-08-31
Slider unit tests updated to use test-extras

### LINKED ISSUE #144: Slider: Update Unit Tests
Update Slider unit tests to use `test-extras`.

smhigley:
**Type:** feature

**Description:**

Resolves #144 
Also updates string casting to consistently use template literals

### DIFF
--- a/src/slider/Slider.ts
+++ b/src/slider/Slider.ts
@@ -130,7 +130,7 @@ export default class Slider extends SliderBase<SliderProperties> {
 		const percentValue = (value - min) / (max - min) * 100;
 
 		// custom output node
-		const outputNode = output ? output(value) : value + '';
+		const outputNode = output ? output(value) : `${value}`;
 
 		// output styles
 		let outputStyles: { left?: string; top?: string } = {};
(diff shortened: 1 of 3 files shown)


src/slider/Slider.ts
```diff
@@ -130,7 +130,7 @@ export default class Slider extends SliderBase<SliderProperties> {
 		const percentValue = (value - min) / (max - min) * 100;
 
 		// custom output node
-		const outputNode = output ? output(value) : value + '';
+		const outputNode = output ? output(value) : `${value}`;
```
kitsonk: both of those are ugly... `String(value)` would be so much better... 🙉

agubler: Template literals for the win 🙈

mwistrand: Personally, I prefer `value + []` for all string casting 🙊

smhigley:
```
`${value}` + String('')
```

kitsonk:
```js
typeof value === 'string' ? value : value.toString()
```

kitsonk:
```ts
[ '' ] + [ value ]
```

🤡

smhigley:
```
Math.random() > 0.5 ? `${value + ''}` : String(value);
```

bitpshr: Template literals are fastest now (we had a JSPerf page going around another PR at some point.)

kitsonk: Ok, fine, though we shouldn't always overvalue performance for code clarity though.

kitsonk: Also, there is no speed advantage as long as we continue to emit to ES5, as the down emit would be:

```js
"" + value
```

But even that is more performant than `String(value)`

agubler: Main reason I like like template literals is that is for consistency as we use them in various scenarios across all our packages. One of which is to effectively stringify a variable :)

kitsonk: Using literals to stringify variables, as long as we are consistent is sufficient and clear.  It is a darn sight better than `value + ''`.  Refs: dojo/meta#198

PR #310, opened 2017-10-10
AccordionPane

### LINKED ISSUE #14: AccordionPane

bitpshr:
**Type:** feature

**Description:**

This PR includes an initial implementation of an AccordionPane component. I intentionally omitted properties like `exclusive` (which would keep only one pane visible at a given time) because this functionality should really be controlled in the parent.

accordion pane

Resolves #14


src/accordionpane/tests/unit/AccordionPane.ts
```diff
@@ -0,0 +1,95 @@
...
+import * as sinon from 'sinon';
+import harness, { Harness } from '@dojo/test-extras/harness';
+
+import AccordionPane, { AccordionPaneProperties } from '../../AccordionPane';
+import TitlePane from '../../../titlepane/TitlePane';
+
+import * as css from '../../styles/accordionPane.m.css';
+
+let pane: Harness<AccordionPaneProperties, typeof AccordionPane>;
+registerSuite({
```
smhigley: Tests should be updated to intern 4

src/accordionpane/AccordionPane.ts
```diff
@@ -0,0 +1,71 @@
...
+				open: includes(openKeys, child!.properties.key),
+				theme
+			});
+
+			return child;
+		});
+	}
+
+	render(): DNode {
+		return v('div', { classes: this.classes(css.root) }, this.renderChildren());
```
smhigley: You know what would be really cool? If we added arrow key navigation through panes 😁 
(am I evil? I might be evil)

src/accordionpane/AccordionPane.ts
```diff
@@ -0,0 +1,71 @@
...
+export const AccordionPaneBase = ThemeableMixin(WidgetBase);
+
+@theme(css)
+export default class AccordionPane extends AccordionPaneBase<AccordionPaneProperties, WNode<TitlePane>> {
+	private _assignCallback(child: WNode<TitlePane>, functionName: string, callback: (key: string) => void) {
+		const existingProperty = (child.properties as any)[functionName];
+		const property = () => { callback.call(this, `${ child.properties.key }`); };
+
+		return existingProperty ? after(existingProperty, property) : property;
+	}
```
smhigley: If we update TitlePane to pass back its key in `onRequest*`, then this function isn't necessary. I think I'd prefer that update, especially since it's currently passing back `this`, which I don't think it should.

src/accordionpane/example/index.ts
```diff
@@ -0,0 +1,61 @@
...
+				w(AccordionPane, {
+					onRequestOpen: (key: string) => {
+						this._openKeys.add(key);
+						this.invalidate();
+					},
+					onRequestClose: (key: string) => {
+						this._openKeys.delete(key);
+						this.invalidate();
+					},
+					openKeys: Array.from(this._openKeys),
```
pottedmeat: Does `Array.from` need to come from the shim?

src/accordionpane/example/index.ts
```diff
@@ -0,0 +1,61 @@
...
+						title: 'Pane 1',
+						key: 'foo'
+					}, [ 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque sodales ante sed massa finibus, at euismod ex molestie. Donec sagittis ligula at lorem blandit imperdiet. Aenean sapien justo, blandit at aliquet a, tincidunt ac nulla. Donec quis dapibus est. Donec id massa eu nisl cursus ornare quis sit amet velit.' ]),
+					w(TitlePane, {
+						title: 'Pane 2',
+						key: 'bar'
+					}, [ 'Ut non lectus vitae eros hendrerit pellentesque. In rhoncus ut lectus id tempus. Cras eget mauris scelerisque, condimentum ante sed, vehicula tellus. Donec congue ligula felis, a porta felis aliquet nec. Nulla mi lorem, efficitur nec lectus vehicula, vehicula varius eros.' ])
+				])
+			])
+		]);
```
smhigley: Since we're saying `exclusive` is the responsibility of the parent, I think it'd be nice to include an example of it here.

src/accordionpane/tests/unit/AccordionPane.ts
```diff
@@ -0,0 +1,95 @@
...
+		const onRequestOpen = sinon.stub();
+
+		pane.setProperties({ onRequestOpen });
+
+		pane.setChildren([
+			w(TitlePane, { title: 'foo', key: 'foo' })
+		]);
+
+		pane.callListener('onRequestOpen', { key: 'foo' });
+		assert.isTrue(onRequestOpen.calledOnce);
```
smhigley: for tests like this one and the next, I've been using `stub.calledWith` to ensure the correct key/any other param is passed through.

src/accordionpane/tests/unit/AccordionPane.ts
```diff
@@ -0,0 +1,95 @@
...
+				onRequestClose: pane.listener,
+				onRequestOpen: pane.listener,
+				open: false,
+				theme: undefined,
+				title: 'baz'
+			})
+		]));
+	},
+
+	'onRequestOpen should be called'() {
```
pottedmeat: Should there be tests to show that children created with/without `onRequest*` have it called when present, and that the accordion pane has `onRequest*` called in either case?

src/accordionpane/AccordionPane.ts
bitpshr: We already support tabbing between panes and using enter to open and close them, so I figured that's sufficient for now until our Ultimate a11y Overhaul Extravaganza.

src/accordionpane/AccordionPane.ts
```diff
@@ -0,0 +1,71 @@
+import { after } from '@dojo/core/aspect';
+import { assign } from '@dojo/core/lang';
+import { DNode, WNode } from '@dojo/widget-core/interfaces';
+import { includes } from '@dojo/shim/array';
```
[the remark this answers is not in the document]
bitpshr: These should already are alphabetized.

src/accordionpane/AccordionPane.ts
```diff
@@ -0,0 +1,71 @@
...
+	onRequestOpen?(key: string): void;
+	openKeys?: string[];
+};
+
+export const AccordionPaneBase = ThemeableMixin(WidgetBase);
+
+@theme(css)
+export default class AccordionPane extends AccordionPaneBase<AccordionPaneProperties, WNode<TitlePane>> {
+	private _assignCallback(child: WNode<TitlePane>, functionName: string, callback: (key: string) => void) {
+		const existingProperty = (child.properties as any)[functionName];
```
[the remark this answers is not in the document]
bitpshr: Smart

src/accordionpane/AccordionPane.ts
bitpshr: This function is still necessary. Its purpose isn't to pass back `key` vs. a TitlePane instance, its purpose is to preserve and aspect an existing `onRequest*` method that's defined on a child TitlePane.

src/accordionpane/example/index.ts
bitpshr: Added.

src/accordionpane/AccordionPane.ts
smhigley: I'll be OK with it if you open a separate issue for adding arrow key navigation :D

bitpshr: Fair, thought we had one. #342

src/accordionpane/AccordionPane.ts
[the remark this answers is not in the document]
pottedmeat: Shouldn't `shim` be before `widget-core`?

src/accordionpane/AccordionPane.ts
smhigley: Ah I see, I didn't know you wanted to allow people to add individual titlepane `onRequest*` methods in Accordion, but that makes sense

src/accordionpane/AccordionPane.ts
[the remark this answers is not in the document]
bitpshr: I alphabetize imports based on the local variable, not the package name. That's how I've always done it and how we've done it in other components as well.

bitpshr:
For reference: @smhigley said she approved this PR after feedback changes but her computer died before she could thumbs it up. I SWEAR I'M NOT LYING. You can trust me 😈.

src/accordionpane/AccordionPane.ts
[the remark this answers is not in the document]
pottedmeat: I've been using https://github.com/dojo/meta/blob/master/STYLE.md#ordering as reference, which I think is still what we should work from?

bitpshr:
PROOF: 
screen shot 2017-10-31 at 11 45 52 am

src/accordionpane/AccordionPane.ts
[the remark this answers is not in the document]
tomdye: Over the top really, I don't believe this kind of nit-picking is being held up across the project so should not hold up this PR.

bitpshr:
@pottedmeat think this is OK despite the ordering? Seems like a hot topic haha.

pottedmeat:
Should be marked approved. I don't really care one way or the other about import order, just trying to make the people that care happy 🥇

PR #311, opened 2017-10-13
Convert to Intern 4

### LINKED ISSUE #263: Update tests to Intern 4

bryanforbes:
**Type:** enhancement

Resolves #263


src/button/tests/functional/Button.ts
```diff
@@ -32,38 +32,36 @@ registerSuite({
...
-					});
-			})
-			.click()
-			.then(function (this: any) {
-				this.getAttribute('aria-pressed')
-					.then((pressed: string) => {
-						assert.strictEqual(pressed, 'false');
-					});
-			})
+			.findByCssSelector(`#example-4 .${css.root}`)
```
bitpshr: I noticed you updated this functional test to target `#example-4` instead of `#example-3`, which like a valid change. Was this passing by accident before and you fixed it intentionally?

bryanforbes: I'm not really sure why this was passing before since `#example-3` isn't toggle-able. I switched it to `#example-4` which is toggle-able.

Issue #102, opened 2017-04-07
Update unit tests to use `test-extras`


rishson:
Use `test-extras` in the unit tests for all of the widgets associated with this Epic.

kitsonk:
Let do what we need to do to close this off.

smhigley:
All unit tests now use test-extras

PR #312, opened 2017-10-17
Listbox

### LINKED ISSUE #117: Common Select/dropdown base widget

smhigley:
**Type:** feature

**Description:**

Refactors Select and ComboBox to use a common Listbox widget. Resolves #117 and #102. This PR also introduces some of the discussed changes to widget patterns, namely splitting the `render()` function into more granular, overridable functions.


src/combobox/ComboBox.ts
```diff
@@ -89,19 +83,18 @@ export default class ComboBox extends ComboBoxBase<ComboBoxProperties> {
...
 
+	private _getMenuId() {
+		return `${this._idBase}-menu`;
+	}
+
 	private _getResultLabel(result: any) {
 		const { getResultLabel } = this.properties;
 
-		return getResultLabel ? getResultLabel(result) : result;
+		return getResultLabel ? getResultLabel(result) : `${result}`;
```
pottedmeat: Could this just be `String(result)`?

src/select/Select.ts
```diff
@@ -18,382 +17,332 @@ import * as iconCss from '../common/styles/icons.m.css';
  *
  * Properties that can be set on a Select component
  *
- * @property CustomOption   Custom widget constructor for options. Should use SelectOption as a base
- * @property describedBy    ID of an element that provides more descriptive text
- * @property disabled       Prevents the user from interacting with the form field
- * @property invalid        Indicates the value entered in the form field is invalid
- * @property label          Label settings for form label text, position, and visibility
- * @property multiple       Whether the widget supports multiple selection
```
pottedmeat: How is this handled now?

src/combobox/example/index.ts
```diff
@@ -61,68 +59,14 @@ const data = [
 	{ value: 'West Virginia' }
 ];
 
-class CustomResultItem extends ResultItem {
```
bitpshr: Why was the example of a custom result item removed?

src/combobox/example/index.ts
```diff
@@ -61,68 +59,14 @@ const data = [
...
-					width: '25px',
-					height: 'auto'
-				}
-			}),
-			getResultLabel(result)
-		]);
-	}
-}
-
-class CustomResultMenu extends ResultMenu {
```
bitpshr: Why was the example of a custom result menuremoved?

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,264 @@
...
+				break;
+		}
+	}
+
+	protected animateScroll(element: HTMLElement, scrollValue: number) {
+		element.scrollTop = scrollValue;
+	}
+
+	@diffProperty('activeIndex', auto)
+	protected calculateScroll(previousProperties: ListboxProperties, { activeIndex = 0 }: ListboxProperties) {
```
bitpshr: What is the purpose of this function? Is this what `_scrollIntoView` used to do?

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,264 @@
...
+
+	protected onElementCreated(element: HTMLElement, key: string) {
+		if (key === 'root' && typeof this._scroll === 'number') {
+			this.animateScroll(element, this._scroll);
+			this._scroll = undefined;
+		}
+	}
+
+	protected onElementUpdated(element: HTMLElement, key: string) {
+		if (key === 'root' && typeof this._scroll === 'number') {
```
bitpshr: Can you refactor this with the login in `onElementCreated`?

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,264 @@
...
+			activeIndex = 0,
+			key = '',
+			getOptionDisabled = (option: any, index: number) => false,
+			getOptionSelected = (option: any, index: number) => false,
+			theme,
+			onActiveIndexChange,
+			onOptionSelect
+		} = this.properties;
+
+		const disabled = getOptionDisabled(option, index);
```
bitpshr: It looks like you call this function a few different places, and also define it in each of those places. Could you pull this out into a static function somewhere? Or consider using a default property.

bitpshr: Never mind, default properties haven't landed.

src/select/Select.ts
```diff
@@ -18,382 +17,332 @@ import * as iconCss from '../common/styles/icons.m.css';
...
+ * @property label             Label settings for form label text, position, and visibility
+ * @property name              The form widget's name
+ * @property options           Array of any type of data for the options
+ * @property readOnly          Allows or prevents user interaction
+ * @property required          Whether or not a value is required
+ * @property useNativeElement  Use the native <select> element if true
+ * @property value             The current value
+ * @property onBlur            Called when the input loses focus
+ * @property onChange          Called when the node's 'change' event is fired
+ * @property onFocus           Called when the input is focused
```
bitpshr: Where did `multiple` go?

src/combobox/ComboBox.ts
```diff
@@ -41,21 +38,19 @@ import * as iconCss from '../common/styles/icons.m.css';
  * @property value              Value to set on the input
  */
 export interface ComboBoxProperties extends ThemeableProperties {
-	autoBlur?: boolean;
 	clearable?: boolean;
-	CustomResultItem?: Constructor<ResultItem>;
```
bitpshr: How are these passed to a ComboBox now?

src/combobox/ComboBox.ts
bitpshr: Template strings are the most optimal way to do string conversion now.

src/combobox/ComboBox.ts
```diff
@@ -164,87 +163,92 @@ export default class ComboBox extends ComboBoxBase<ComboBoxProperties> {
...
-		// Scroll menu up so top of highlighted result aligns with top of menu container
-		if (element.offsetTop - menu.scrollTop < 0) {
-			menu.scrollTop = element.offsetTop;
-		}
-		// Scroll menu down so bottom of highlighted result aligns with bottom of menu container
-		else if ((element.offsetTop - menu.scrollTop + element.offsetHeight) > menu.clientHeight) {
-			menu.scrollTop = element.offsetTop - menu.clientHeight + element.offsetHeight;
-		}
-	}
+	// should update to use interesection meta
```
bitpshr: Why is this commented out? Is this necessary anymore at all since this looks to be handled in ListBox?

src/combobox/ComboBox.ts
```diff
@@ -266,118 +267,158 @@ export default class ComboBox extends ComboBoxBase<ComboBoxProperties> {
...
-			selectedIndex: this._activeIndex,
-			theme
-		});
+		return v('div', {
+			key: 'dropdown',
+			classes: this.classes(css.dropdown),
+			onmouseover: this._onResultHover,
+			onmousedown: this._onResultMouseDown
+		}, [
+			w(Listbox, {
```
bitpshr: Did we lose the ability to pass custom result items and menus into the ComboBox?

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,264 @@
...
+ */
+
+export interface ListboxProperties extends ThemeableProperties {
+	activeIndex?: number;
+	describedBy?: string;
+	visualFocus?: boolean;
+	id?: string;
+	multiselect?: boolean;
+	optionData?: any[];
+	tabIndex?: number;
```
bitpshr: tabIndex is missing from the TypeDocs

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,264 @@
...
+				active: activeIndex === index,
+				classes: this.getOptionClasses(activeIndex === index, disabled, selected),
+				disabled,
+				label: this.renderOptionLabel(option, index),
+				id: this._getOptionId(index),
+				key: `option-${index}`,
+				option,
+				selected,
+				theme,
+				onClick: (data: any) => {
```
bitpshr: This should probably be its own function so it's not redefined on every render.

src/listbox/example/index.ts
```diff
@@ -0,0 +1,139 @@
+import { WidgetBase } from '@dojo/widget-core/WidgetBase';
+import { WidgetProperties } from '@dojo/widget-core/interfaces';
+import { ProjectorMixin } from '@dojo/widget-core/mixins/Projector';
+import { v, w } from '@dojo/widget-core/d';
+import Listbox from '../Listbox';
+import dojoTheme from '../../themes/dojo/theme';
+
+interface CustomOption {
+	value: string;
```
bitpshr: Nit, but alphabetize

src/listbox/example/index.ts
```diff
@@ -0,0 +1,139 @@
...
+			w(Listbox, {
+				key: 'listbox2',
+				activeIndex: this._listbox2Index,
+				id: 'listbox2',
+				optionData: this._moreOptions,
+				theme: this._theme,
+				getOptionLabel: (option: CustomOption) => option.label,
+				getOptionDisabled: (option: CustomOption) => !!option.disabled,
+				getOptionSelected: (option: CustomOption) => !!option.selected,
+				onActiveIndexChange: (index: number) => {
```
bitpshr: I thought the types of the arguments for these callbacks could be inferred now, but I could be insane.

src/select/Select.ts
```diff
@@ -18,382 +17,332 @@ import * as iconCss from '../common/styles/icons.m.css';
  *
  * Properties that can be set on a Select component
  *
- * @property CustomOption   Custom widget constructor for options. Should use SelectOption as a base
```
bitpshr: Is this no longer supported?

src/combobox/ComboBox.ts
```diff
@@ -266,118 +267,158 @@ export default class ComboBox extends ComboBoxBase<ComboBoxProperties> {
...
-			index = nextIndex(index);
-		}
-
-		this._activeIndex = index;
+		this._activeIndex = nextIndex(this._activeIndex);
 		this.invalidate();
 	}
 
-	protected onElementCreated(element: HTMLElement, key: string) {
+	protected onElementUpdated(element: HTMLElement, key: string) {
```
bitpshr: Can we remove reliance on `onElement*` methods? Technically they're sort of kind of deprecated, though the message was temporarily removed. I'm wondering if we can leverage meta in a way that negates the need for these altogether.

src/select/Select.ts
```diff
@@ -18,382 +17,332 @@ import * as iconCss from '../common/styles/icons.m.css';
...
+	onBlur?(key: string | number): void;
+	onChange?(option: any, key: string | number): void;
+	onFocus?(key: string | number): void;
 }
 
 export const SelectBase = ThemeableMixin(WidgetBase);
 
 @theme(css)
 @theme(iconCss)
+@diffProperty('options', reference)
```
bitpshr: Was this working incorrectly before explicitly diffing by reference?

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,264 @@
...
+		}
+	}
+
+	protected animateScroll(element: HTMLElement, scrollValue: number) {
+		element.scrollTop = scrollValue;
+	}
+
+	@diffProperty('activeIndex', auto)
+	protected calculateScroll(previousProperties: ListboxProperties, { activeIndex = 0 }: ListboxProperties) {
+		const scrollOffset = this.meta(Dimensions).get('root').scroll.top;
```
tomdye: This should be a single call to `this.meta(Dimensions)` not multiple

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,264 @@
...
+				onActiveIndexChange && onActiveIndexChange(0, key);
+				break;
+			case Keys.End:
+				onActiveIndexChange && onActiveIndexChange(optionData.length - 1, key);
+				break;
+		}
+	}
+
+	protected animateScroll(element: HTMLElement, scrollValue: number) {
+		element.scrollTop = scrollValue;
```
tomdye: I think the correct way to do this would be to create a local `scrollTop` meta within this file that takes a `key` and a `scrollValue`. That would avoid the need to use `onElementCreated` / `onElementUpdated`

src/combobox/ComboBox.ts
smhigley: ooops, forgot I had that 😂

src/combobox/ComboBox.ts
smhigley: Yup, I think we talked about not passing around sub-widget constructors anymore, right? I think Ant or Matt G originally wanted them gone entirely, but it makes handling events more difficult. I suppose ComboBox/Select could take an extension of Listbox though, since it's a standalone widget.

src/select/Select.ts
smhigley: This briefly came up in the dojo2 room: since Listbox can essentially function as a multiselect widget and native multiselect is awful anyway, I dropped support for it in Select.

src/combobox/ComboBox.ts
bitpshr: I think that the ability to easily customize items and the menu itself is important. I'm not tied to passing in a custom class, but we need some way to accomplish the same functionality that was provided previously. I don't think the APIs should change drastically because of the Listbox change.

smhigley: I could definitely add the `CustomMenu` that takes an extension of Listbox back in. It seemed (at least to me) that we were moving in the direction of having most or all customization happen through extension. I'm definitely open to other ideas, though.

bitpshr: I can get on board with extension and no sub-widgets as long as the offloaded render methods are extremely small and simple to override.

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,264 @@
...
+
+		const disabled = getOptionDisabled(option, index);
+		const selected = getOptionSelected(option, index);
+
+		return v('div', { key: this._getOptionId(index) }, [
+			w(ListboxOption, {
+				active: activeIndex === index,
+				classes: this.getOptionClasses(activeIndex === index, disabled, selected),
+				disabled,
+				label: this.renderOptionLabel(option, index),
```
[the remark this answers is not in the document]
smhigley: Should I break out `id` as well? I originally only pulled out `disabled` and `selected` because they're also used to get the state classes whereas `label`/`id` are only used once.

src/select/Select.ts
smhigley: Listbox on its own is effectively a custom multiselect widget, and the native multiselect is so universally awful that I don't think we need to go out of our way to support it.

src/combobox/ComboBox.ts
smhigley: Would the animation meta be able to handle scroll? I suppose we could write our own within Listbox, but I'd rather use something that exists, if we can.

src/listbox/Listbox.ts
smhigley: More or less. I separated out `calculateScroll` and `animateScroll` so that the latter would be easy to override (e.g. to animate the scroll)

src/listbox/example/index.ts
smhigley: I get implicit any errors without the type ¯\_(ツ)_/¯

src/select/Select.ts
smhigley: Yup, it doesn't call a render if you update the `options` property otherwise.

src/combobox/ComboBox.ts
smhigley: Moved to meta 👍

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,264 @@
...
+import { diffProperty } from '@dojo/widget-core/decorators/diffProperty';
+import { auto, reference } from '@dojo/widget-core/diff';
+import { v, w } from '@dojo/widget-core/d';
+import Dimensions from '@dojo/widget-core/meta/Dimensions';
+import uuid from '@dojo/core/uuid';
+import { Keys } from '../common/util';
+import * as css from './styles/listbox.m.css';
+
+/* Listbox Option sub-widget */
+export interface ListboxOptionProperties extends ThemeableProperties {
```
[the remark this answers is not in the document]
smhigley: Good question 😂 . It used to be, then after talking with Ant/Matt G, the conclusion was that sub-widgets shouldn't really exist because they don't stand on their own, except that they make handling certain events much easier. I moved it out of its own file so it would seem less separate, but I agree that it seemed better organized when separate

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,270 @@
...
+	activeIndex?: number;
+	describedBy?: string;
+	getOptionDisabled?(option: any, index: number): boolean;
+	getOptionId?(option: any, index: number): string;
+	getOptionLabel?(option: any, index: number): DNode;
+	getOptionSelected?(option: any, index: number): boolean;
+	id?: string;
+	multiselect?: boolean;
+	optionData?: any[];
+	ScrollMeta?: WidgetMetaConstructor<DefaultScroll>;
```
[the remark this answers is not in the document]
smhigley: I just want to make sure it's easy to customize, e.g. to animate the scroll or something. Do you think there's anything wrong with passing it in as a property? I suppose just having `protected animateScroll` could be enough, since someone could override it.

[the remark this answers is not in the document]
tomdye: To me it feels somewhat overkill to provide both an extension point and an injection point. Should we not simply choose one pattern and stick to it?

src/listbox/Listbox.ts
```diff
@@ -0,0 +1,270 @@
...
+		const { ScrollMeta = DefaultScroll } = this.properties;
+		this.meta(ScrollMeta).scroll('root', scrollValue);
+	}
+
+	@diffProperty('activeIndex', auto)
+	protected calculateScroll(previousProperties: ListboxProperties, { activeIndex = 0 }: ListboxProperties) {
+		const menuDimensions = this.meta(Dimensions).get('root');
+		const scrollOffset = menuDimensions.scroll.top;
+		const menuHeight = menuDimensions.offset.height;
+		const optionOffset = this.meta(Dimensions).get(this._getOptionId(activeIndex)).offset;
```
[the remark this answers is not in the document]
smhigley: It returns a unique string that's used in two places, once for an `id` and once for a `key`

src/listbox/Listbox.ts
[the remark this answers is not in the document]
smhigley: Works for me. I'm fine removing the prop and leaving `protected animateScroll` as the point of extension 👍

src/listbox/Listbox.ts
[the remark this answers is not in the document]
smhigley: Moved back to its own file. We'll see what happens 😄

PR #319, opened 2017-10-17
Widget property generics

### LINKED ISSUE #313: Allow individual widget properties to be extended
Allow individual widget properties to be extended when extending the widget, so instead of:
   `export default class Button extends ButtonBase<ButtonProperties>`
   we would have:
   `export default class Button<P extends ButtonProperties = ButtonProperties> extends ButtonBase<P>`

### LINKED ISSUE #314: Extendable widget properties
**Enhancement** 

Allow each widget's properties interface to be extended, so instead of:
`export default class Button extends ButtonBase<ButtonProperties>`
we would have:
`export default class Button<P extends ButtonProperties = ButtonProperties> extends ButtonBase<P>`

Comes from the first point in https://github.com/dojo/widgets/issues/266#issuecomment-335559039

smhigley:
**Type:** feature

**Description:**

Resolves #314. This is currently blocked by https://github.com/Microsoft/TypeScript/issues/17005, which causes TS errors when narrowing optional properties.


kitsonk:
Microsoft/TypeScript#17005 will be in TypeScript 2.6 final.

src/calendar/example/index.ts
```diff
@@ -6,6 +6,24 @@ import Calendar from '../../calendar/Calendar';
 import Checkbox from '../../checkbox/Checkbox';
 import dojoTheme from '../../themes/dojo/theme';
 
+interface MyWidgetProperties {
```
bitpshr: Was this intentional, or just for testing?

smhigley: Definitely a mistake :joy:. Thanks for catching that!

bitpshr:
+1 for strictly typing children. I believe most of our applicable components do, like TabController and AccordionPane. Definitely something to stay vigilant about.

PR #171, opened 2017-05-11
Add `Label` functional tests.

lzhoucs:
**Type:** functional tests

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)
* [x] Unit or Functional tests are included in the PR

Resolves #153

PR #321, opened 2017-10-20
Label functional tests

### LINKED ISSUE #153: Label: Add Functional Tests

pottedmeat:
**Type:** feature

**Description:**

Adds basic functional tests for `Label` based on lzhoucs's PR https://github.com/dojo/widgets/pull/171 with rebasing and updates.

Resolves #153

### DIFF
--- a/src/label/example/index.ts
+++ b/src/label/example/index.ts
@@ -20,7 +20,7 @@ export class App extends WidgetBase<WidgetProperties> {
 			]),
 			v('h3', {}, ['Hidden label after the input']),
 			w(Label, {
-                extraClasses: { root: 'label2' },
+				extraClasses: { root: 'label2' },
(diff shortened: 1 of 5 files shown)


src/label/example/index.ts
```diff
@@ -10,6 +10,7 @@ export class App extends WidgetBase<WidgetProperties> {
 			v('h1', {}, ['Label Examples']),
 			v('h3', {}, ['Label assigned as string without extra options']),
 			w(Label, {
+				extraClasses: { root: 'label1' },
```
tomdye: i don't think we should ever show `extraClasses` in an example if we can get away with it as it's kind of a back door / anti pattern. In this case better to wrap in a `div` / `span` with styles

PR #324, opened 2017-10-23
Textarea functional tests

### LINKED ISSUE #160: TextArea: Add Functional Tests

pottedmeat:
**Type:** feature

**Description:**

Resolves #160

### DIFF
--- a/src/textarea/example/index.ts
+++ b/src/textarea/example/index.ts
@@ -9,7 +9,7 @@ export class App extends WidgetBase<WidgetProperties> {
 	private _theme: {};
 	private _value1: string;
 	private _value2: string;
-	private _invalid = false;
+	private _invalid: boolean;
 
 	themeChange(event: TypedTargetEvent<HTMLInputElement>) {
 		const checked = event.target.checked;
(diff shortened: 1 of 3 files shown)


src/textarea/tests/functional/Textarea.ts
```diff
@@ -0,0 +1,155 @@
...
+			.sleep(500)
+			.findByCssSelector(`.t3.${css.root}`)
+				.getProperty('className')
+				.then((className: string) => {
+					assert.notInclude(className, css.valid);
+					assert.include(className, css.invalid);
+				})
+			.end();
+	},
+	'hidden label should not be displayed'() {
```
smhigley: I'm not sure we need to add an example and test for this, since it's part of label functionality rather than textarea functionality (and I'm not a fan of hidden labels, so not big on modeling them in all the widget examples)

src/textarea/example/index.ts
```diff
@@ -52,6 +54,7 @@ export class App extends WidgetBase<WidgetProperties> {
 			}),
 			v('h3', {}, ['Validated, Required Textarea']),
 			w(Textarea, {
+				extraClasses: { root: 't3' },
```
tomdye: same comment as others re. wrapping div with class rather than `extraClasses`.

Issue #309, opened 2017-10-10
Focus Management


kitsonk:
@kitsonk commented on [Wed Nov 16 2016](https://github.com/dojo/widget-core/issues/107)

**User Story**

A developer needs an easy way to manage focus in an application and easily move that focus.

**Discussion**

Focus management is a core bit of functionality to "get right".  Focus management in a virtual DOM though can require programatically setting focus on the real DOM node during a render.  It makes sense to consider an "abstraction" of some sort, that is integrated into the projector, that allows a focus manager to "collect" the real DOM nodes and when a widget needs to have its focus set, it would be a call to the projector with something like `setFocus(widget)` or `setFocus('widget-id')`.  Possibly this would be an `app` construct too.

Also, note that in many cases, the DOM/VDom root emitted by the widget would not be the focus node.  Therefore we will need a way identify the nodes that are the focus node for a particular widget (e.g. `data-focus-node` attribute).

---

@mwistrand commented on [Tue Apr 04 2017](https://github.com/dojo/widget-core/issues/107#issuecomment-291655897)

So would applications be required to pass the projector around in order for widgets to receive focus when they need it? In this case, how might widgets that have specific requirements for keyboard navigation handle focus management without requiring that the application explicitly pass in a projector instance?

---

@kitsonk commented on [Wed Apr 05 2017](https://github.com/dojo/widget-core/issues/107#issuecomment-291764917)

I think we need to switch it around...

Why would a widget need to grab focus?  It might need to express that "if I have focus, this sub part of me should be the focus" and some sort of "if you navigated within me, here is the current ordering of my focus points".  As a widget would reveal different sub-parts, it would re-render that "meta-data".

Remember, widgets are reactive.  They respond to things.  They don't actually do anything if we can avoid it at any cost.

The management part of it would have to occur at a point where some actor has a whole view of everything going on.  Widgets shouldn't take focus, otherwise it would be chaos, like little squabbling children.

---

@mwistrand commented on [Wed Apr 05 2017](https://github.com/dojo/widget-core/issues/107#issuecomment-291837160)

[...]

PR #332, opened 2017-10-25
372 - Escape key closes dialog

### LINKED ISSUE #327: Escape key should close Dialog
**Bug**

Hitting the escape key should close the Dialog component.

Package Version: latest

tomdye:
**Type:** bug

**Description:**

Close Dialog when escape key is pressed

Resolves #327

### DIFF
--- a/src/dialog/Dialog.ts
+++ b/src/dialog/Dialog.ts
@@ -68,9 +68,7 @@ export default class Dialog extends DialogBase<DialogProperties> {
 	}
 
 	private _onKeyUp(event: KeyboardEvent) {
-		const { keyCode } = event;
-
-		if (keyCode === Keys.Escape) {
+		if (event.keyCode === Keys.Escape) {
 			this._onCloseClick();
 		}
 	}
--- a/src/dialog/Dialog.ts
+++ b/src/dialog/Dialog.ts
@@ -3,6 +3,8 @@ import { WidgetBase } from '@dojo/widget-core/WidgetBase';
 import { ThemeableMixin, ThemeableProperties, theme } from '@dojo/widget-core/mixins/Themeable';
 import { v } from '@dojo/widget-core/d';
 import uuid from '@dojo/core/uuid';
+import { Keys } from '../common/util';
+import { createHandle } from '@dojo/core/lang';
 
 import * as css from './styles/dialog.m.css';
 import * as iconCss from '../common/styles/icons.m.css';
@@ -65,6 +67,24 @@ export default class Dialog extends DialogBase<DialogProperties> {
 		!this.properties.modal && this._onCloseClick();
 	}
 
+	private _onKeyUp(event: KeyboardEvent) {
+		const { keyCode } = event;
+
+		if (keyCode === Keys.Escape) {
+			this._onCloseClick();
+		}
+	}
+
+	constructor() {
+		super();
+
+		const keyUpFunc = this._onKeyUp.bind(this);
+		document.addEventListener('onkeyup', keyUpFunc);
+		this.own(createHandle(() => {
+			document.removeEventListener('onkeyup', keyUpFunc);
+		}));
+	}
+
 	render(): DNode {
 		const {
 			closeable = true,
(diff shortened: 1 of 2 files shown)


tomdye:
Looks like CI is failing due to textinput functional test failure

src/dialog/Dialog.ts
```diff
@@ -65,6 +67,24 @@ export default class Dialog extends DialogBase<DialogProperties> {
 		!this.properties.modal && this._onCloseClick();
 	}
 
+	private _onKeyUp(event: KeyboardEvent) {
+		const { keyCode } = event;
```
bitpshr: You could either destructure the argument itself to save two lines (`private _onKeyUp({ keyCode }) {`) or you could really not even destructure at all since you only ever reference a property on `event` once, so it's equally as optimal to just use `event.keyCode` directly. That'd be my vote, not to over-destructure.

src/dialog/Dialog.ts
```diff
@@ -65,6 +67,24 @@ export default class Dialog extends DialogBase<DialogProperties> {
...
+		if (keyCode === Keys.Escape) {
+			this._onCloseClick();
+		}
+	}
+
+	constructor() {
+		super();
+
+		const keyUpFunc = this._onKeyUp.bind(this);
+		document.addEventListener('onkeyup', keyUpFunc);
```
bitpshr: I really don't like having to use a document-level handler for this. Is it possible instead to use the root node and to trap focus somehow?

src/dialog/Dialog.ts
tomdye: happy to do so.

src/dialog/Dialog.ts
tomdye: This is a similar approach to that used in `SlidePane`. I don't believe it's worthwhile adding `focus trapping` complications at this point before we have a focus management meta / mixin available. I could attach / detach the listener when the dialog is open / closed instead perhaps?

bitpshr: Fair enough.

src/dialog/Dialog.ts
```diff
@@ -65,6 +67,22 @@ export default class Dialog extends DialogBase<DialogProperties> {
 		!this.properties.modal && this._onCloseClick();
 	}
 
+	private _onKeyUp(event: KeyboardEvent) {
+		if (event.keyCode === Keys.Escape) {
```
smhigley: we generally use `event.which` in widgets. There's no difference for the escape key, but it's nice to standardize things.

src/dialog/Dialog.ts
smhigley: I think it's fine to have the handler on the root of the modal. It's not keyboard accessible now anyway without focus trapping, so I don't think it's worth the document-level handler for this.

src/dialog/Dialog.ts
```diff
@@ -78,11 +96,15 @@ export default class Dialog extends DialogBase<DialogProperties> {
 			underlay
 		} = this.properties;
 
+		document.body.onkeyup = this._onKeyUp.bind(this);
```
smhigley: Is this necessary, since you already have the `document.addEventListener`?

src/dialog/Dialog.ts
tomdye: @smhigley  i couldn't get it to trigger the callback on the root of the model, but i will give it another go to make sure i did not mess it up.

src/dialog/Dialog.ts
tomdye: I copied this from `titlepane`, happy to change as we should encourage consistency.

tomdye: raised: https://github.com/dojo/widgets/issues/341 to address

src/dialog/Dialog.ts
smhigley: @tomdye it's a little unfortunate in that the focus doesn't move into the modal when it opens right now :/. I made my stuff comments rather than requested changes b/c I would totally understand if you decide to continue with the current approach so it just works.

src/dialog/Dialog.ts
tomdye: unlikely, must have been a mistake.

src/dialog/Dialog.ts
smhigley: haha didn't even notice `titlepane` still had keyCode 😆

smhigley: I was also debating making the switch over to `event.key`, since I think the support is pretty much there now.

src/dialog/Dialog.ts
tomdye: not a problem, i guess that this is something that we can address when focus management becomes a thing.

src/dialog/Dialog.ts
tomdye: feel free to add a comment on https://github.com/dojo/widgets/issues/341

tomdye:
@smhigley @bitpshr comments addressed

bitpshr:
Looks good to merge Tom. I added [a comment](https://github.com/dojo/widgets/issues/309#issuecomment-340837295) to #309 to remove the document-level handler after focus management is sorted.

PR #336, opened 2017-10-27
Simplify widget render

### LINKED ISSUE #316: Break out render() code
Break out `render()` code into more manageable chunks, specifically in the following ways:

   - For widgets that generate more DOM, move sensible chunks into separate functions, e.g. `renderControls()` for the buttons in ComboBox, or `renderOptions` in Listbox.
   - For widgets that render children, e.g. TabController, return `this.children` in a separate function. Just as an example, this would make it much easier to add custom loading content.
   - Split "state" classes into a separate function (`getModifierClasses`?) for all widgets that could potentially support multiple visual states, e.g. `loading` or `dirty` on a tab button.
   - If a widget uses an icon, generate the icon separately. Swapping out icons or using something other than a font icon for a visual symbol seems like a fairly common modification, e.g. changing left/right month arrows in Calendar for text, or tiny month preview images.
   - Widgets that calculate inline styles (right now only Slider, I think) should do so in a separate fn. This would allow a triangular volume slider to adjust the height of the thumb as it changes, for example.

smhigley:
**Type:** feature

**Description:**

Resolves #316 

Makes widgets more easily extensible by simplifying the render code, particularly addressing these points from the issue:
- break out large chunks of rendered dom
- state classes are returned in a function
- icons are broken out
- inline styles are calculated separately

### DIFF
--- a/src/button/Button.ts
+++ b/src/button/Button.ts
@@ -70,11 +70,11 @@ export default class Button extends ButtonBase<ButtonProperties> {
 	private _onTouchEnd (event: TouchEvent) { this.properties.onTouchEnd && this.properties.onTouchEnd(event); }
 	private _onTouchCancel (event: TouchEvent) { this.properties.onTouchCancel && this.properties.onTouchCancel(event); }
 
-	protected getContent() {
+	protected getContent(): DNode[] {
 		return this.children;
 	}
 
-	protected getModifierClasses() {
+	protected getModifierClasses(): (string | null)[] {
 		const {
 			disabled,
 			popup = false,
@@ -88,7 +88,7 @@ export default class Button extends ButtonBase<ButtonProperties> {
 		];
 	}
 
-	protected renderPopupIcon() {
+	protected renderPopupIcon(): DNode {
 		return v('i', { classes: this.classes(css.addon, iconCss.icon, iconCss.downIcon),
 			role: 'presentation', 'aria-hidden': 'true'
 		});
--- a/src/button/Button.ts
+++ b/src/button/Button.ts
@@ -70,6 +70,30 @@ export default class Button extends ButtonBase<ButtonProperties> {
 	private _onTouchEnd (event: TouchEvent) { this.properties.onTouchEnd && this.properties.onTouchEnd(event); }
 	private _onTouchCancel (event: TouchEvent) { this.properties.onTouchCancel && this.properties.onTouchCancel(event); }
 
+	protected getContent() {
+		return this.children;
+	}
+
+	protected getModifierClasses() {
+		const {
+			disabled,
+			popup = false,
+			pressed
+		} = this.properties;
+
+		return [
+			disabled ? css.disabled : null,
+			popup ? css.popup : null,
+			pressed ? css.pressed : null
+		];
+	}
+
+	protected renderPopupIcon() {
+		return v('i', { classes: this.classes(css.addon, iconCss.icon, iconCss.downIcon),
+			role: 'presentation', 'aria-hidden': 'true'
+		});
+	}
+
 	render(): DNode {
 		let {
 			describedBy,
@@ -87,12 +111,7 @@ export default class Button extends ButtonBase<ButtonProperties> {
 		}
 
 		return v('button', {
-			classes: this.classes(
-				css.root,
-				disabled ? css.disabled : null,
-				popup ? css.popup : null,
-				pressed ? css.pressed : null
-			),
+			classes: this.classes(css.root, ...this.getModifierClasses()),
 			disabled,
 			id,
 			name,
@@ -115,10 +134,8 @@ export default class Button extends ButtonBase<ButtonProperties> {
 			'aria-pressed': typeof pressed === 'boolean' ? pressed.toString() : null,
 			'aria-describedby': describedBy
 		}, [
-			...this.children,
-			popup ? v('i', { classes: this.classes(css.addon, iconCss.icon, iconCss.downIcon),
-				role: 'presentation', 'aria-hidden': 'true'
-			}) : null
+			...this.getContent(),
+			popup ? this.renderPopupIcon() : null
 		]);
 	}
 }
--- a/src/calendar/Calendar.ts
+++ b/src/calendar/Calendar.ts
@@ -1,7 +1,7 @@
 import { WidgetBase } from '@dojo/widget-core/WidgetBase';
 import { ThemeableMixin, ThemeableProperties, theme } from '@dojo/widget-core/mixins/Themeable';
 import { v, w } from '@dojo/widget-core/d';
-import { DNode, Constructor } from '@dojo/widget-core/interfaces';
+import { DNode } from '@dojo/widget-core/interfaces';
 import uuid from '@dojo/core/uuid';
 import { Keys } from '../common/util';
 import { CalendarMessages } from './DatePicker';
@@ -16,7 +16,6 @@ import * as iconCss from '../common/styles/icons.m.css';
  *
  * Properties that can be set on a Calendar component
  *
- * @property CustomDateCell    Custom widget constructor for the date cell. Should use CalendarCell as a base.
  * @property labels            Customize or internationalize accessible text for the Calendar widget
  * @property month             Set the currently displayed month, 0-based
  * @property monthNames        Customize or internationalize full month names and abbreviations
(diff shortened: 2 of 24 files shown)


src/button/Button.ts
```diff
@@ -70,6 +70,30 @@ export default class Button extends ButtonBase<ButtonProperties> {
 	private _onTouchEnd (event: TouchEvent) { this.properties.onTouchEnd && this.properties.onTouchEnd(event); }
 	private _onTouchCancel (event: TouchEvent) { this.properties.onTouchCancel && this.properties.onTouchCancel(event); }
 
+	protected getContent() {
```
bitpshr: It may be nice to embrace ES getters if no arguments are passed to these:

```ts
get content() {
    return this.children;
}
```

src/calendar/tests/unit/Calendar.ts
```diff
@@ -20,11 +20,9 @@ const compareId = compareProperty((value: any) => {
 	return typeof value === 'string';
 });
 
-let dateIndex = -1;
 const expectedDateCell = function(widget: any, date: number, active: boolean) {
-	dateIndex++;
 	return w(CalendarCell, {
-		key: `date-${dateIndex}`,
+		key: `date-${date}${active ? '' : '-dimmed'}`,
```
bitpshr: Why was this change necessary?

src/calendar/Calendar.ts
```diff
@@ -16,7 +16,6 @@ import * as iconCss from '../common/styles/icons.m.css';
  *
  * Properties that can be set on a Calendar component
  *
- * @property CustomDateCell    Custom widget constructor for the date cell. Should use CalendarCell as a base.
```
bitpshr: Did you remove sub-widgets as part of this PR as well?

src/button/Button.ts
tomdye: i'd say this is overkill @bitpshr especially if the `getter` ends up doing more than simply getting.

src/calendar/Calendar.ts
```diff
@@ -321,38 +320,65 @@ export default class Calendar extends CalendarBase<CalendarProperties> {
...
+			onRequestMonthChange: (requestMonth: number) => {
+				onMonthChange && onMonthChange(requestMonth);
+			},
+			onRequestYearChange: (requestYear: number) => {
+				onYearChange && onYearChange(requestYear);
+			}
+		});
+	}
+
+	protected renderPagingButtonContent(type: 'next' | 'previous') {
```
tomdye: `next` / `previous` would be better as a string enum

src/calendar/DatePicker.ts
```diff
@@ -191,7 +191,38 @@ export default class DatePicker extends DatePickerBase<DatePickerProperties> {
 		onPopupChange && onPopupChange(this._getPopupState());
 	}
 
-	private _renderMonthRadios() {
+	protected renderControlsTrigger(type: 'month' | 'year'): DNode {
```
tomdye: string enum

src/calendar/DatePicker.ts
```diff
@@ -215,7 +246,20 @@ export default class DatePicker extends DatePickerBase<DatePickerProperties> {
 		]));
 	}
 
-	private _renderYearRadios() {
+	protected renderPagingButtonContent(type: 'next' | 'previous') {
```
tomdye: string enum

src/button/Button.ts
bitpshr: Fair.

src/calendar/Calendar.ts
smhigley: The problem is sub-widgets make it much easier to handle data binding on event handlers (e.g. `onClick` on a specific date cell returns the correct date and disabled state). Based on talking with Ant, I think the compromise reached was to leave in sub-widgets, but try not to expose them to the user

src/calendar/tests/unit/Calendar.ts
smhigley: when I moved rendering the actual date cell into its own function (`renderDateCell`), I didn't have access to the index any more, and it seemed overkill to pass it in just for the key, when I could also generate a unique key based on the date and currentMonth.

src/calendar/Calendar.ts
smhigley: I've been wondering if it makes sense to remove the typedoc for CalendarCell. It's marginally useful to developers wanting to extend CalendarCell, but I definitely don't want it showing up in Dojo documentation

src/timepicker/TimePicker.ts
```diff
@@ -166,41 +166,50 @@ export const TimePickerBase = ThemeableMixin(WidgetBase);
...
+	}
+
+	private _getOptionLabel(value: TimeUnits) {
+		const { getOptionLabel } = this.properties;
+		const units = parseUnits(value);
+		return getOptionLabel ? getOptionLabel(units) : this._formatUnits(units);
+	}
+
+	private _onNativeBlur(event: FocusEvent) {
+		this.properties.onBlur && this.properties.onBlur((<HTMLInputElement> event.target).value);
```
pottedmeat: You could add this interface and use this instead of `FocusEvent` to clean things up a little.

```ts
interface FocusInputEvent extends FocusEvent {
	target: HTMLInputElement;
}
```

src/titlepane/TitlePane.ts
```diff
@@ -108,36 +132,19 @@ export default class TitlePane extends TitlePaneBase<TitlePaneProperties> {
...
-					closeable ? css.closeableFixed : null,
-					css.titleFixed
-				),
-				onclick: this._onTitleClick,
-				onkeyup: this._onTitleKeyUp,
+				classes: this.classes(css.title, ...this.getModifierClasses()).fixed(css.titleFixed, ...this.getFixedModifierClasses()),
 				role: 'heading'
 			}, [
-				v('div', {
+				v('button', {
```
pottedmeat: Why did this change to a button?

src/titlepane/TitlePane.ts
```diff
@@ -108,36 +132,19 @@ export default class TitlePane extends TitlePaneBase<TitlePaneProperties> {
...
-					'aria-disabled': closeable ? null : 'true',
 					'aria-expanded': String(open),
+					disabled: !closeable,
 					classes: this.classes(css.titleButton),
 					id: this._titleId,
-					role: 'button',
-					tabIndex: closeable ? 0 : -1
+					onclick: this._onTitleClick
 				}, [
-					v('i', {
```
pottedmeat: Why isn't this displayed anymore?

src/calendar/DatePicker.ts
```diff
@@ -191,7 +207,38 @@ export default class DatePicker extends DatePickerBase<DatePickerProperties> {
...
+		const onclick = type === Controls.month ? this._onMonthButtonClick : this._onYearButtonClick;
+
+		return v('button', {
+			key: `${type}-button`,
+			'aria-controls': `${this._idBase}_${type}_dialog`,
+			'aria-expanded': `${open}`,
+			'aria-haspopup': 'true',
+			id: `${this._idBase}_${type}_button`,
+			classes: this.classes(
+				(css as any)[`${type}Trigger`],
```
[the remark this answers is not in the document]
smhigley: yeah, I chose concise code over css types in this particular case, but I was kinda ambivalent about it.

src/titlepane/TitlePane.ts
smhigley: I always thought it should be a button, since one of the basic a11y principles in our widget design is to use native elements whenever possible. I just took this widget pattern PR as a good place to change it. It also simplifies the code a ton, since we no longer need to handle focus and key events manually.

src/titlepane/TitlePane.ts
smhigley: it is, I just moved it to `this.renderExpandIcon()`

src/calendar/DatePicker.ts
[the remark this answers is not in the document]
pottedmeat: The closest thing I found is:

```ts
const ControlTrigger: { [key: string]: 'monthTrigger' | 'yearTrigger' } = {
	[Controls.month]: 'monthTrigger',
	[Controls.year]: 'yearTrigger'
};

const ControlTriggerActive: { [key: string]: 'monthTriggerActive' | 'yearTriggerActive' } = {
	[Controls.month]: 'monthTriggerActive',
	[Controls.year]: 'yearTriggerActive'
};
```

```ts
css[ControlTrigger[type]],
open ? css[ControlTriggerActive[type]] : null
```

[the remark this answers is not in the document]
pottedmeat: ^^ This still misses type safety on the key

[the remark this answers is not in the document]
pottedmeat: Something like this adds type safety and minimal code but does require a lot of manually typed strings.

```ts
function controlClass(control: Controls, suffix: 'Trigger' | 'TriggerActive'): 'monthTrigger' | 'monthTriggerActive' | 'yearTrigger' | 'yearTriggerActive' {
	return `${control}${suffix}` as any;
}
```

PR #340, opened 2017-10-30
SlidePane: Add top / bottom support

### LINKED ISSUE #323: SlidePane - Support top/bottom?
**Enhancement**

Would it make sense to support top/bottom anchoring for a SlidePane?  I have a use case where I would want to slide out some diagnostic information from the bottom of a UI.

bitpshr:
**Type:** feature

**Description:**

This PR adds top and bottom alignment support to the SlidePane component.

Resolves #323


src/slidepane/SlidePane.ts
```diff
@@ -11,16 +11,18 @@ import uuid from '@dojo/core/uuid';
  * Enum for left / right alignment
  */
 export const enum Align {
-	left,
-	right
+	bottom = 'bottom',
```
bitpshr: I was hoping to get clever and use string enum values as indexes when accessing CSS classes, but it didn't work out as nicely as I'd hoped.

src/slidepane/SlidePane.ts
```diff
@@ -50,19 +52,46 @@ const DEFAULT_WIDTH = 320;
...
 
+	get plane() {
+		const { align = Align.left } = this.properties;
+		return align === Align.left || align === Align.right ? Plane.x : Plane.y;
+	}
+
+	private _getDelta(event: MouseEvent & TouchEvent, eventType: string) {
+		const { align = Align.left } = this.properties;
+
+		if (this.plane === Plane.x) {
```
bitpshr: Originally I got cleverer about how heavily I refactored all these `Plane.x` vs. `Plane.y` logic branches, but readability was pure despite saving many LOC.

src/slidepane/SlidePane.ts
```diff
@@ -156,7 +189,10 @@ export default class SlidePane extends SlidePaneBase<SlidePaneProperties> {
 
 		const contentClasses = [
 			css.pane,
-			align === Align.left ? css.left : css.right,
+			align === Align.left ? css.left : null,
```
bitpshr: This is where I originally tried to get clever using string enum values above for `Align`. I couldn't get it to work because of typings and indexes. Happy to try a less verbose approach though.

src/slidepane/SlidePane.ts
kitsonk: You would need it to be [something like this](https://www.typescriptlang.org/play/#src=const%20enum%20Align%20%7B%0A%20%20bottom%20%3D%20'bottom'%2C%0A%20%20top%20%3D%20'top'%2C%0A%20%20left%20%3D%20'left'%2C%0A%20%20right%20%3D%20'right'%0A%7D%0A%0Adeclare%20const%20css%3A%20%7B%20%5BK%20in%20Align%5D%3A%20string%3B%20%7D%3B%0A%0Alet%20direction%20%3D%20Align.bottom%3B%0A%0Aconst%20classes%20%3D%20%5B%20css%5Bdirection%5D%20%5D%3B%0A) I suspect:

```ts
const enum Align {
  bottom = 'bottom',
  top = 'top',
  left = 'left',
  right = 'right'
}

declare const css: { [K in Align]: string; };

let direction = Align.bottom;

const classes = [ css[direction] ];
```

src/slidepane/example/index.ts
```diff
@@ -94,4 +94,4 @@ export class App extends WidgetBase<WidgetProperties> {
 const Projector = ProjectorMixin(App);
 const projector = new Projector();
 
-projector.append();
+projector.append();
```
kitsonk: Suspect this is a mistake...  should end with a newline

bitpshr:
Thanks @kitsonk. Ready for another glance if you have a chance.

src/slidepane/SlidePane.ts
```diff
@@ -50,19 +52,42 @@ const DEFAULT_WIDTH = 320;
...
 	private _content: HTMLElement;
-	private _initialX: number;
+	private _initialPosition: number;
 	private _slideIn: boolean;
 	private _swiping: boolean;
 	private _titleId = uuid();
 	private _transform: number;
 	private _wasOpen: boolean;
 
+	get plane() {
```
tomdye: protected / private?

src/slidepane/styles/slidePane.m.css
```diff
@@ -72,19 +100,22 @@
...
 }
 
-.left.slideIn {
-	transition: transform ease-in-out var(--short-animation-duration);
+.bottomFixed.slideInFixed {
+	transform: translateY(0%) !important;
 }
 
-.right.slideIn {
+.left.slideOut,
```
tomdye: do you really need `.right` / `.top` etc prefixes here? Looks like you've added all of them

bitpshr:
@tomdye ready for another review, thanks!

PR #346, opened 2017-10-31
Fix broken functional tests

bryanforbes:
**Type:** bug

**Description:**

* Run Firefox on Windows 10 to always get latest version
* Change AccordionPane test to click on heading (Edge and Safari drivers have broken click)
* Change TextInput example to use onInput (Edge and Safari drivers don't fire onChange correctly)
* Change TextInput test to compare active element instead of CSS classes


tomdye:
Are the extra `sleep(500)` calls really necessary ?

bryanforbes:
> Are the extra sleep(500) calls really necessary ?

Actually, since the example is using `onInput`, the whole switching focus code is unnecessary. I'll rework the test.

PR #347, opened 2017-11-01
Slider dragging bug

### LINKED ISSUE #328: Slider issue in FireFox
**Bug**

The Slider sometimes doesn't slide when dragged using the mouse in FireFox on OSX.

Package Version: latest

smhigley:
**Type:** bug

**Description:**

Resolves #328 

Updated theme styles to make the invisible input controls larger. I also added `onChange` in the example page to fix a VO + Safari issue where the visual UI didn't update while interacting with the input.

### DIFF
--- a/src/slider/example/index.ts
+++ b/src/slider/example/index.ts
@@ -60,6 +60,7 @@ export class App extends WidgetBase<WidgetProperties> {
 				},
 				step: 1,
 				value: tribbleValue,
+				onChange: this.onTribbleInput,
 				onInput: this.onTribbleInput,
 				theme: this._theme
 			}),
@@ -87,6 +88,7 @@ export class App extends WidgetBase<WidgetProperties> {
 					});
 				},
 				outputIsTooltip: true,
+				onChange: this.onTribbleInput,
 				onInput: this.onVerticalInput,
 				theme: this._theme
 			})
--- a/src/themes/dojo/slider.m.css
+++ b/src/themes/dojo/slider.m.css
@@ -39,20 +39,21 @@
 }
 
 .input {
-	height: var(--border-width);
+	height: 15px;
 	outline: none;
+	top: -7px;
 }
 .input::-ms-thumb {
-	height: calc(var(--grid-base) * 2);
-	width: calc(var(--grid-base) * 2);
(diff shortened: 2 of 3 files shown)


src/slider/example/index.ts
```diff
@@ -60,6 +60,7 @@ export class App extends WidgetBase<WidgetProperties> {
 				},
 				step: 1,
 				value: tribbleValue,
+				onChange: this.onTribbleInput,
```
bitpshr: Does this mean that both onChange and onInput are required when using this component in certain browsers?

src/themes/dojo/slider.m.css
```diff
@@ -39,20 +39,21 @@
 }
 
 .input {
-	height: var(--border-width);
+	height: 15px;
```
bitpshr: Is this component-specific?

src/themes/dojo/variables.css
```diff
@@ -68,6 +68,7 @@
 	--grid-base: 8px;
 	--spacing-regular: var(--grid-base);
 	--spacing-large: calc(var(--grid-base) * 2);
+	--touch-target-min: 48px;
```
bitpshr: You could base this on `--grid-base` like other derived variables.

src/themes/dojo/slider.m.css
smhigley: yeah, I just picked a number that seemed big enough. I'm not sure what variable it would be / if it should even be a variable

src/themes/dojo/variables.css
smhigley: I considered it, but I think it's actually an independent pixel value in its own right, based on recommendations for minimum touch target sizes. If it were based on `--grid-size`, that would imply it could change if you changed the grid base, whereas I think it should always stay the same

src/slider/example/index.ts
bitpshr: (we chatted about this, only for VO + Safari)

bitpshr:
🚢

src/slider/example/index.ts
smhigley: I need to remember to file a bug report with VoiceOver :P

PR #356, opened 2017-11-02
Use output of tcm and update unit tests

### LINKED ISSUE #307: tcm should be part of the build process
**Enhancement**

`tcm` should be added to the grunt `dev` and `dist` process to ensure that our styles are aligned properly to the CSS classes in the style files.

maier49:
**Type:** bug introfuced by a feature

**Description:**
This updates the existing types to the output of the new `tcm` task in grunt, based on the current css files, and fixes some related issues in unit tests where they are asserting the presence of non existent classes.

Along with a new release of grunt-dojo2, this will resolve #307


maier49:
The churn is purely because of using `tcm`. This was the only file that had extra classes that weren't in the actual css file apparently, so it was the only one that changed.

PR #359, opened 2017-11-06
Z-index organization

### LINKED ISSUE #335: Organize z-index values
**Bug**

We already have some issues with conflicting z-index values (see https://dojo.github.io/widgets, and try opening the SlidePane or Dialog over the TitlePane). We should document and implement a standard approach to z-index values, and probably put it in the widgets readme and perhaps also in an article or tutorial.

Initial thoughts on z-index ranges:

0 - 100: Any specific component layering (e.g. social icons that overlap an image)
100 - 200: Tooltips
200 - 300: Dropdowns (e.g. menu or combobox)
300 - 400: Fixed position elements
400 - 500: Modals / Dialogs (including underlays)
500 - 600: Notifications and special cases

smhigley:
**Type:** feature

**Description:**

Resolves #335.

Some styles were modified to reduce the use of tabIndex, particularly in TabController. Z-index styles were also removed from fixed classes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ import Button, { ButtonProperties } from '@dojo/widgets/button/Button';
 ```
 
 Because each widget is a separate module, when you create a release build of your application, you will only include the widgets that you have explicitly imported.
-This allows our [`dojo cli`](https://github.com/dojo/cli) build tooling to make sure that the production build of your application only includes the widgets you use and is as small as possible. 
+This allows our [`dojo cli`](https://github.com/dojo/cli) build tooling to make sure that the production build of your application only includes the widgets you use and is as small as possible.
 
 ## Features
 
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ When creating a widget, spacing (margin, padding) should be specified using `px`
 
 ### Z-index layering
 
-Widgets adhere to a basic convention for blocking off ranges of z-index values based on function and visual context. This is present in both individual widget CSS and in the Dojo theme styles. These values can be overridden in a custom theme if necessary, since no `z-index` values are set in fixed styles.
+Widgets adhere to a basic convention for using specific ranges of z-index values based on function and visual context. This convention is followed in both individual widget CSS and in the Dojo theme styles. These values can be overridden in a custom theme if necessary, since no `z-index` values are set in fixed styles.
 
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -115,6 +115,19 @@ Where a theme requires specific icons that are not part of the Font Awesome set,
 `px vs. em` - we specify font sizes in `px`.
 When creating a widget, spacing (margin, padding) should be specified using `px` unless the design calls for proportional spacing, in which case `em` can be used.
 
+### Z-index layering
+
+Widgets adhere to a basic convention for blocking off ranges of z-index values based on function and visual context. This is present in both individual widget CSS and in the Dojo theme styles. These values can be overridden in a custom theme if necessary, since no `z-index` values are set in fixed styles.
```
bitpshr: "blocking off" -> "using specific"
"This is present" -> "This convention is followed"

src/dialog/styles/dialog.m.css
```diff
@@ -11,7 +11,7 @@
 	width: 450px;
 	overflow: hidden;
 	max-width: calc(100% - 30px);
-	z-index: 2;
+	z-index: calc(var(--zindex-dialog) + 1);
```
bitpshr: Why is the `+ 1` necessary? Shouldn't the dialog use the `--zindex-dailog` CSS variable directly?

src/slidepane/styles/slidePane.m.css
```diff
@@ -8,17 +8,16 @@
 	left: 0;
 	width: 100%;
 	height: 100%;
-	z-index: 1;
 }
 
 .underlayVisible {
 	background: var(--underlay-background);
+	z-index: var(--zindex-dialog);
```
bitpshr: I see, so it looks like you're using `--zindex-dialog` for underlays, and then adding one for the content itself. For some reason that seems odd to me, maybe it's just a matter of variable naming.

smhigley: I think technically they could both be `--zindex-dialog`, since the dialog is after the underlay in the dom. I'd be fine with that if you are, although the `+ 1` is a little nice in that it makes the layering explicit.

bitpshr: I'm fine with everything being explicit.

PR #361, opened 2017-11-06
Select placeholder

### LINKED ISSUE #298: Add placeholder support to Select
**Enhancement**

It would be nice to be able to add a placeholder to Select.

Package Version: latest

**Expected behavior:**

```ts
w(Select, {
	placeholder: 'Choose an option',
	options: dataSmall,
	value: selectValue,
	onChange: (option: OptionData) => {
		this.setState({ selectValue: option.value });
	}
})
```

**Actual behavior:**

```ts
w(Select, {
	options: dataSmall,
	value: selectValue,
	onChange: (option: OptionData) => {
		this.setState({ selectValue: option.value });
	}
})
```

smhigley:
**Type:** feature

**Description:**

Resolves #298. Adds Select placeholder. Also adds `::placeholder` styles to TextInput so the two will match.


src/select/Select.ts
```diff
@@ -70,6 +72,15 @@ export default class Select extends SelectBase<SelectProperties> {
 	private _open = false;
 	private _baseId = uuid();
 
+	private _getOptionLabel(option: any) {
```
pottedmeat: Shouldn't this be protected without the `_`?

pottedmeat: Okay, I see that `getOptionLabel` is still passed as a property. Should that be moved to this method instead?

src/select/Select.ts
```diff
@@ -285,18 +295,30 @@ export default class Select extends SelectBase<SelectProperties> {
...
+
 		const selectedOption = find(options, (option: any, index: number) => {
 			return getOptionSelected ? getOptionSelected(option, index) : false;
-		}) || options[0];
+		});
+
+		if (selectedOption) {
+			label = this._getOptionLabel(selectedOption);
+		}
+
```
pottedmeat: extra linebreak?

src/select/Select.ts
smhigley: @pottedmeat I think the property should stay for consistency with Listbox and ComboBox. I added `_getOptionLabel` because I realized I was doing `getOptionLabel ? getOptionLabel(option) : fallback` a number of times, and wanted to consolidate. I made it private because I don't think it needs to be overridden in addition to having `getOptionLabel`

smhigley: But perhaps it should be public for consistency as well? It's pretty similar to other methods that are public /shrug

src/combobox/example/index.ts
```diff
@@ -75,8 +75,8 @@ export class App extends WidgetBase<WidgetProperties> {
 		this.invalidate();
 	}
 
-	onChange(value: string, key: string) {
-		(this as any)[`_value${key}`] = value;
+	onChange(value: string, key?: string) {
+		(this as any)[`_value${key!}`] = value;
```
tomdye: Can you get this working without using `!` please?

src/select/Select.ts
```diff
@@ -70,6 +72,15 @@ export default class Select extends SelectBase<SelectProperties> {
 	private _open = false;
 	private _baseId = uuid();
 
+	private _getOptionLabel(option: any) {
+		const { getOptionLabel } = this.properties;
+		if (!option) {
```
tomdye: Should this not be passed to the `getOptionLabel` anyway, the custom handler may wish to do something with empty labels.

src/select/styles/select.m.css
```diff
@@ -12,6 +12,7 @@
 	background: var(--component-background);
 	border: 1px solid var(--component-color);
 	font-size: inherit;
+	min-height: 25px;
```
tomdye: do we have this in a variable somewhere for consistency ?

src/themes/dojo/textinput.m.css
```diff
@@ -35,6 +35,9 @@
 	box-shadow: var(--box-shadow-dimensions-small) color(var(--color-box-shadow));
 }
 
+.input::placeholder { color: var(--color-text-faded); }
+.input:-ms-input-placeholder { color: var(--color-text-faded); }
```
tomdye: is the prefixed placeholder not handled by our autoprefixer plugin?

src/select/styles/select.m.css
smhigley: Not really, since we aren't setting the heights of inputs in any standardized way. Perhaps we should go through the widgets styling and do a variable setup similar to the one we use in the dojo theme? I think it's outside the scope of this PR though.

src/select/Select.ts
smhigley: good point 👍

PR #362, opened 2017-11-06
Tooltip

### LINKED ISSUE #92: Tooltip widget
# Specification
This widget provides tooltip behaviour.
 
## Features
- You specify the content for a tooltip. 
- Tooltip content can contain arbitrary vDOM nodes and widgets, e.g. 
screen shot 2017-03-30 at 2 30 03 pm
- The tooltip can be positioned to open in different directions, _relative_ to the element that causes the tooltip to be shown, e.g. `{position: TooltipPosition.TopCenter}` would open a tooltip above the element.
screen shot 2017-03-30 at 2 32 50 pm

- You can specify the event that will launch a tooltip. Usually this will be the `hover` of an element, but it could also be a click event etc.

Ability to provide advanced styling and animation.

`onHide`
`onShow`

Depends on the content of the tooltip. If the tooltip contained widgets, then would you need to be able to tab into this content - or is this _horrendous_ UX 🤔 This affects the acceptance criteria.

None

Controlled

**List of any icons needed**

Look and feel of the tooltip, pointing arrow.

**Any other considerations**:

## Acceptance criteria
When I perform an action that should make a tooltip appear, the tooltip appears, without obscuring the element that I was interacting with.
When I perform an action that should make the tooltip disappear, then when the tooltip is no longer shown, the focus should be on the element that had focus before the tooltip was shown.

bitpshr:
**Type:** feature

**Description:**

Tooltip

This PR introduces a new Tooltip component that can serve as a basis for other tooltip-related functionality like form component validation. The properties interface for Tooltip is intentionally minimal. Initially, children were wrapped in a single parent node, and specific events could be used to show or hide the tooltip (`click`, `mouseenter`, etc.) This seemed overly restrictive, especially when used in conjunction with other Dojo 2 widgets like a TextInput. Without mandating an API that all Tooltip targets / children must implement, and to support regular DOM, Dojo 2 components, and MyAmazingFramework widgets, the ability to show the tooltip must be fully inverted. Luckily for us, that keeps the code fairly clean.

Resolves #92


src/accordionpane/createAccordionPaneElement.ts
```diff
@@ -8,8 +8,20 @@ export default function createAccordionPaneElement(): CustomElementDescriptor {
 	return {
 		tagName: 'dojo-accordionpane',
 		widgetConstructor: AccordionPane,
-		attributes: [],
-		properties: [],
-		events: []
+		properties: [
```
bitpshr: This was an omission on my part with the AccordionPane PR.

kitsonk: Why is it part of a tooltip PR?

src/slidepane/SlidePane.ts
```diff
@@ -203,6 +203,8 @@ export default class SlidePane extends SlidePaneBase<SlidePaneProperties> {
 			align = Align.left,
 			open = false
 		} = this.properties;
+		// https://github.com/Microsoft/TypeScript/issues/13042
+		// const alignCss: {[Key in Align]: string} = css;
```
kitsonk: Even if there is an issue, an uncomment line of code is ugly... if there is further action here when the issue is resolved, we should have an issue created in widgets, referencing this issue, and detailing what we need to do so we don't forget about it.

kitsonk: Also again, seems wholly unrelated to this PR.

src/accordionpane/createAccordionPaneElement.ts
bitpshr: Removed. Opened #364 to track adding this and the AccordionPane README.

src/slidepane/SlidePane.ts
bitpshr: Agreed. I removed this comment altogether from SlidePane. I also created #365 to track the updating of component code once TS 2.7 is released and fixes the respective issue.

src/tooltip/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Dojo 2's `Tooltip` component can be used to show or hide content around a child component based on specific events.
+
+## Features
+
+- Agnostic to an eventing implementation
+- Supports top, right, bottom, and left orientations
+- Optionally persistent
+
+![Image of basic dialog](http://placekitten.com/450/300)
```
tomdye: not a dialog, remove placekitten

src/tooltip/README.md
```diff
@@ -0,0 +1,49 @@
...
+*Basic Example*
+```typescript
+import TextInput from '@dojo/widgets/textinput/TextInput';
+import Tooltip from '@dojo/widgets/tooltip/Tooltip';
+import { w } from '@dojo/widget-core/d';
+
+w(Tooltip, {
+	key: 'foo',
+	content: 'This is a tooltip!',
+	showing: this._tooltipShowing
```
tomdye: do we use `showing` or `show` / `open` / `visible` elsewhere? not a fan of the `showing` naming here

tomdye: looks like `Dialog` and `SlidePane` use `open` ?

bitpshr: A tooltip doesn't open or close, but I'll go with `open` for consistency.

Issue #357, opened 2017-11-03
Updates to latest widget-core and test-extras breaking changes


kitsonk:
**Bug**

There are several breaking changes in coming releases of `widget-core` and `test-extras` that will require updates to `widgets`.

PR #363, opened 2017-11-07
Widget Core and Test Extras Updates

kitsonk:
**Type:** bug

* [X] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)
* [X] Unit or Functional tests are included in the PR

**Description:**

This PR updates widgets with the breaking changes from `widget-core` and `test-extras` that are either in master, or soon to be in master.

A couple notes:

- Several of the `ComboBox` unit tests were beyond repair.  They were never converted to the widget harness and the changes to how widget-core.  I spent a good 3-4 hours trying to fix it or convert it to harness, but it is complicated and so for now I have commented out the tests and suggest someone familiar with that widget convert it
- At least one functional test for ComboBox is now failing.
- I will rebase off of master and deal with any merge conflicts for this

Fixes: #357


kitsonk:
Ok, ComboBox is throwing an error that I can't figure out and I tried to update the ScrollMeta test is failing though I have tried my best to update it to the latest widget-core.  Going to need some help on both of those.  @agubler any ideas?

kitsonk:
Ok, with the publish of test-extras this is essentially ready for review with the exception of the one failure of the `ScrollMeta` test... It appears the node isn't there when rendered, and therefore never gets it scroll position adjusted.  I am not sure it is a valid test under the new widget core though.

kitsonk:
Ok, I think I fixed the `ScrollMeta` test... Hopefully CI will prove me right.

kitsonk:
The one test failure, validated TextInput on IE11, is not a problem with the functional tests, it appears to be a legitimate regression somehow.  Opened #367 to track it.

PR #366, opened 2017-11-07
Add AccordionPane README and custom element descriptor

### LINKED ISSUE #364: Add missing AccordionPane README and custom element descriptor
**Bug**

The AccordionPane component doesn't contain a README or a custom element descriptor.

Package Version: latest

bitpshr:
**Type:** bug

* [ ] ~~Unit or Functional tests are included in the PR~~

**Description:**

Adds a README for AccordionPane (mostly placeholder, like other components) and a proper custom element descriptor.

Resolves #364

### DIFF
--- a/src/accordionpane/README.md
+++ b/src/accordionpane/README.md
@@ -1,0 +1,53 @@
+# @dojo/widgets/accordionpane/AccordionPane widget
+
+Dojo 2's `AccordionPane` component can be used to show multiple pieces of content inside collapsible panes within a common parent component. It provides responsive styling and a mechanism of control over multiple child `TitlePane` components.
+
+## Features
+
+- Optionally supports multiple panes open at once
+
+![Image of basic dialog](http://placekitten.com/450/300)
+
+### Accessibility Features
+
+An `AccordionPane` expects all children to be `TitlePane` components, and as such, supports standard keyboard navigation for toggling child content open or closed.
+
+**TitleBar Events**:
+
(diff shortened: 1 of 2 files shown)


src/accordionpane/README.md
```diff
@@ -0,0 +1,53 @@
+# @dojo/widgets/accordionpane/AccordionPane widget
```
tomdye: have we used the full path in naming these elsewhere? I think just `# AccordionPane` will suffice

src/accordionpane/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Dojo 2's `AccordionPane` component can be used to show multiple pieces of content inside collapsible panes within a common parent component. It provides responsive styling and a mechanism of control over multiple child `TitlePane` components.
+
+## Features
+
+- Optionally supports multiple panes open at once
+
+![Image of basic dialog](http://placekitten.com/450/300)
```
tomdye: Not a dialog, this is accordion pane and this should not be a place kitten

src/accordionpane/README.md
bitpshr: Yes, we used full paths elsewhere. I'm going to stick with it here to be consistent. If we want to change that, we should change it throughout all components.

src/accordionpane/README.md
bitpshr: I removed the picture entirely since our other components either have no picture or a placekitten picture. These READMEs are not thorough or entirely standardized yet.

PR #368, opened 2017-11-09
Add TimePicker functional tests

### LINKED ISSUE #162: TimePicker: Add Functional Tests

pottedmeat:
**Type:** feature

**Description:**

Wraps example tests with identifiers. Adds functional tests.

Resolves #162

### DIFF
--- a/src/timepicker/tests/functional/TimePicker.ts
+++ b/src/timepicker/tests/functional/TimePicker.ts
@@ -1,0 +1,194 @@
+const { registerSuite } = intern.getInterface('object');
+const { assert } = intern.getPlugin('chai');
+
+import { Remote } from 'intern/lib/executors/Node';
+import Command from '@theintern/leadfoot/Command';
+import Element from '@theintern/leadfoot/Element';
+import keys from '@theintern/leadfoot/keys';
+import * as comboBoxCss from '../../../combobox/styles/comboBox.m.css';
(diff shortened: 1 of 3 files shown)


src/timepicker/tests/functional/TimePicker.ts
```diff
@@ -0,0 +1,217 @@
...
+
+		return testDisabledPicker(this.remote, 'example-disabled');
+	},
+	'readonly timepickers cannot be opened'() {
+		return testDisabledPicker(this.remote, 'example-readonly');
+	},
+	'validated inputs update on input'() {
+		const { browserName } = this.remote.session.capabilities;
+		if (browserName === 'internet explorer') {
+			this.skip('Test does not work on Internet Explorer');
```
bitpshr: Should we be more descriptive with the messages we pass when skipping tests? Do we even know the level at which this test fails? Intern? Leadfoot? A driver? This is largely rhetorical, just trying to think of ways to make the skips less ugly.

This may be as good as we can do for now.

pottedmeat: When I'm able to easily find a reason in one of the other skips, I use it. Otherwise, trying to figure out the underlying cause is what had been holding up the tests, so I haven't been bothering anymore. I think the plan was to just skip with abandon then document and fix later.

PR #372, opened 2017-11-14
360 - Label Changes - using for and id instead of wrapping label

### LINKED ISSUE #360: Investigate normalising Label usage
@tomdye commented on [Mon Nov 06 2017](https://github.com/dojo/widget-core/issues/749)

# Enhancement

Currently when using a `Label` it is set as the outer node of a widget and is passed classes by the creating widget to enable theming and styling. We should investigate using an outer div instead to avoid needing to pass classes and other such fields to the label.

Plan is to investigate different approaches considering the pros and cons including passing classes vs properties and utilising `for` / `id` properties to allow accessibility. 

## Current
```ts
const rootClasses = [
   css.root,
  disabled ? css.disabled : null
];

w('Label', { extraClasses: { root: rootClasses.join('') }, labelText: 'Email Address' }, [
	v('input', { type: 'text' })
]);
```

## Proposed 1
```ts
v('div', { classes: this.theme(css.root) }, [
	w('Label', { for: 'inputId', labelText: 'Email Address', disabled: true }),
	v('input', { id: 'inputId', type: 'text' })
]);
```

## Proposed 2
```ts
v('div', { classes: this.theme(css.root) }, [
	w('Label', { labelText: 'Email Address', disabled: true }, [
		v('input', { type: 'text' })
	])	
]);
```

tomdye:
**Type:** feature

**Description:**

Changes our widgets to use a `Label` as a standard widget. Passing it properties to determine it's own theming classes etc. 
In usage, Labels are created when a `label` string is passed to a widget that accepts `LabeledProperties`. The `Label` is created alongside the widget within it's `root` node.
Modifiers `labelHidden` and `labelAfter` can be used to control the visibility and position of the label.

**TODO:**
* [x] Raise issue to theme labels. 
* [x] Raise issue to remove individual label theming from each widget

Resolves #360 
Raises: 
- https://github.com/dojo/widgets/issues/373
- https://github.com/dojo/widgets/issues/374
- https://github.com/dojo/widgets/issues/375
- https://github.com/dojo/widgets/issues/376


src/common/interfaces.d.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @property onKeyPress     Called on the input's keypress event
+ * @property onKeyUp        Called on the input's keyup event
+ * @property onMouseDown    Called on the input's mousedown event
+ * @property onMouseUp      Called on the input's mouseup event
+ * @property onTouchStart   Called on the input's touchstart event
+ * @property onTouchEnd     Called on the input's touchend event
+ * @property onTouchCancel  Called on the input's touchcancel event
+ */
+export interface InputProperties extends InputEventProperties {
+	describedBy?: string;
```
bitpshr: The `controls` property is missing despite being documented.

src/common/interfaces.d.ts
```diff
@@ -0,0 +1,65 @@
...
+	disabled?: boolean;
+	invalid?: boolean;
+	maxLength?: number | string;
+	minLength?: number | string;
+	name?: string;
+	placeholder?: string;
+	readOnly?: boolean;
+	required?: boolean;
+	value?: string;
```
bitpshr: Same with `type`.

src/common/interfaces.d.ts
```diff
@@ -0,0 +1,65 @@
...
+	onInput?(event: Event): void;
+	onKeyDown?(event: KeyboardEvent): void;
+	onKeyPress?(event: KeyboardEvent): void;
+	onKeyUp?(event: KeyboardEvent): void;
+}
+
+export interface InputEventProperties {
```
bitpshr: Should the documentation for this interface be separate than the other interface?

src/label/Label.ts
```diff
@@ -1,85 +1,56 @@
...
-}
-
-/**
- * This is a helper function for using `extraClasses` with Label.
- * It can be used as follows:
- * extraClasses: { root: parseLabelClasses(this.theme([ css.class1, css.class2 ])) }
- */
-export function parseLabelClasses(classes: SupportedClassName | SupportedClassName[]): string {
-	return Array.isArray(classes) ? classes.filter((str) => str).join(' ') : classes || '';
+	disabled?: boolean;
```
bitpshr: I think we need documentation for these properties.

src/radio/Radio.ts
```diff
@@ -29,32 +31,23 @@ import * as css from './styles/radio.m.css';
...
-export interface RadioProperties extends ThemedProperties {
+export interface RadioProperties extends ThemedProperties, LabeledProperties, InputEventProperties {
 	checked?: boolean;
 	describedBy?: string;
 	disabled?: boolean;
 	invalid?: boolean;
-	label?: string | LabelOptions;
 	name?: string;
 	readOnly?: boolean;
 	required?: boolean;
```
bitpshr: Why aren't all InputProperties removed from this interface?

tomdye: the interfaces still need some more work to split out better

tomdye:
Ready for another review @bitpshr

src/textinput/TextInput.ts
```diff
@@ -149,23 +117,11 @@ export default class TextInput extends TextInputBase<TextInputProperties> {
 				ontouchend: this._onTouchEnd,
 				ontouchcancel: this._onTouchCancel
 			})
```
smhigley: I think we should actually keep `inputWrapper`. It's pretty necessary for anything but basic CSS styles since you can't add pseudo elements to an `<input>`. TextInput addons would need it, for example.

src/combobox/ComboBox.ts
```diff
@@ -38,15 +39,14 @@ import * as iconCss from '../common/styles/icons.m.css';
...
  * @property value              Value to set on the input
  */
-export interface ComboBoxProperties extends ThemedProperties {
+export interface ComboBoxProperties extends ThemedProperties, LabeledProperties {
 	clearable?: boolean;
 	disabled?: boolean;
 	getResultLabel?(result: any): string;
 	id?: string;
-	inputProperties?: TextInputProperties;
+	inputProperties?: InputProperties;
```
smhigley: doesn't this still need to be `TextInputProperties`? It should be valid to add `type` to the input.

src/combobox/ComboBox.ts
```diff
@@ -38,15 +39,14 @@ import * as iconCss from '../common/styles/icons.m.css';
  * @property results            Results for the current search term; should be set in response to `onRequestResults`
  * @property value              Value to set on the input
  */
-export interface ComboBoxProperties extends ThemedProperties {
+export interface ComboBoxProperties extends ThemedProperties, LabeledProperties {
 	clearable?: boolean;
 	disabled?: boolean;
 	getResultLabel?(result: any): string;
 	id?: string;
-	inputProperties?: TextInputProperties;
```
bitpshr: This should probably still be `TextInputProperties`.

src/combobox/ComboBox.ts
```diff
@@ -390,20 +389,17 @@ export default class ComboBox extends ThemedBase<ComboBoxProperties> {
...
-		]);
-
-		if (label) {
-			controls = w(Label, {
-				label,
-				theme
-			}, [ controls ]);
-		}
+		const controls = [
+			label ? w(Label, { key: 'label', theme, disabled, invalid, readOnly, required, hidden: labelHidden, forId: id }, [ label ]) : null,
```
bitpshr: Nit, but can we break this line up somehow?

src/common/interfaces.d.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @property onBlur         Called when the input loses focus
+ * @property onChange       Called when the node's 'change' event is fired
+ * @property onFocus        Called when the input is focused
+ */
+export interface InputEventProperties {
+	onBlur?(event: FocusEvent): void;
+	onChange?(event: Event): void;
+	onFocus?(event: FocusEvent): void;
+	onInput?(event: Event): void;
+}
```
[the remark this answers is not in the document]
bitpshr: +1

src/label/Label.ts
```diff
@@ -1,85 +1,61 @@
...
-};
-
 /**
  * @type LabelProperties
  *
  * Properties that can be set on a Label component
  *
  * @property forId     ID to explicitly associate the label with an input element
- * @property label      Label settings for form label text, position, and visibility
+ * @property disabled
```
bitpshr: Were description for these properties intentionally omitted? It's probably fine if so, just making sure.

src/radio/Radio.ts
bitpshr: Circling back on this. Why can't we remove all of these properties yet? What additional work will be required, and will it be part of this PR?

src/textinput/TextInput.ts
```diff
@@ -109,18 +71,24 @@ export default class TextInput extends TextInputBase<TextInputProperties> {
...
 			placeholder,
 			readOnly,
 			required,
 			type = 'text',
-			value
+			value,
+			theme
 		} = this.properties;
 
-		const textinput = v('div', { classes: this.theme(css.inputWrapper) }, [
```
bitpshr: We may want to keep this for styling. @smhigley may have more insight.

src/checkbox/Checkbox.ts
```diff
@@ -173,24 +163,13 @@ export default class Checkbox extends CheckboxBase<CheckboxProperties> {
 					ontouchend: this._onTouchEnd,
 					ontouchcancel: this._onTouchCancel
 				})
-			])
+			]),
+			label ? w(Label, { theme, disabled, invalid, readOnly, required, hidden: labelHidden, forId: this._uuid }, [ label ]) : null
```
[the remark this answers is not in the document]
tomdye: We'll need to extend `Label` -> `CheckBoxLabel` such that it gets it own style. I can raise a follow up PR for this should we *really* want to do this.

src/label/Label.ts
tomdye: they seemed kinda obvious to me. Can add them in if we feel it's important

src/slider/Slider.ts
```diff
@@ -149,7 +111,7 @@ export default class Slider extends SliderBase<SliderProperties> {
 
 		return v('output', {
 			classes: [ this.theme(css.output), outputIsTooltip ? css.outputTooltip : null ],
-			for: `${this._inputId}`,
+			for: this._inputId,
```
bitpshr: Is this string conversion not necessary?

tomdye: didn't seem so. let me double check

tomdye: `this._inputId` comes from `uuid()` thus is inferred as a string.

src/common/interfaces.d.ts
```diff
@@ -0,0 +1,70 @@
+export interface LabeledProperties {
+	labelAfter?: boolean;
+	labelHidden?: boolean;
+	label?: string;
```
smhigley: Could this be a `DNode`? That would give a little more flexibility in usage (e.g. adding an icon after the text, or a red "*", etc)

src/common/interfaces.d.ts
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * @type PointerEventProperties
+ * @property onClick        Called when the input is clicked
+ * @property onMouseDown    Called on the input's mousedown event
+ * @property onMouseUp      Called on the input's mouseup event
+ * @property onTouchStart   Called on the input's touchstart event
+ * @property onTouchEnd     Called on the input's touchend event
+ * @property onTouchCancel  Called on the input's touchcancel event
+ */
+export interface PointerEventProperties {
```
smhigley: 👍

src/common/interfaces.d.ts
tomdye: The icon / * would be achieved via the appropriate modifier classes on `label` itself. If the user wishes to use a more complicated `Label` type the can do this by manually creating and wrapping.
ie.

```ts
w(MySpecialLabel, {}, [
   w(TextInput, {})
])
```

tomdye:
@smhigley will address your checkbox / span concerns in the morning

tomdye:
@smhigley there were two options for the `faded` version of the label.

1) create `checkRadioLabel` which extends `Label`
2) add a `faded` property to label.

I went for the latter as I believe this gives more flexibility in future usage and avoids the scenario where people may end up using a `checkRadioLabel` to wrap something thats not a check of a radio to achieve a faded label.

src/common/interfaces.d.ts
smhigley: It's an easy non-breaking change to make in the future if we think it's necessary, so I'm willing to go with `string` for now and see if anyone complains

PR #378, opened 2017-11-16
Proposal: Update widget examples

agubler:
**Type:** feature

**Description:**

 * Use a single projector in the index instead and switch widgets using the registry instead of changing the window location and pealing the hash for the module
 * Add the theme switch to the common example index.ts instead of in every example

### DIFF
--- a/src/common/example/index.html
+++ b/src/common/example/index.html
@@ -38,25 +38,7 @@
 			}
 		}));
 
-		(function () {
-			var query = window.location.search.substring(1).split('&');
-			var module;
-
-			for (var i = 0; i < query.length; i++) {
-				var pair = query[i].split('=');
-				if (pair[0].toLowerCase() === 'module') {
-					module = pair[1].toLowerCase();
-					break;
-				}
-			}
-
-			require(['@dojo/shim/main'], function () {
-				require([ 'src/common/example/index' ], function () {});
-				if (module) {
-					require([ 'src/' + module + '/example/index' ], function () {});
-				}
-			});
-		})();
+		require([ 'src/common/example/index' ], function () {});
 	</script>
 </body>
 </html>
--- a/src/common/example/index.ts
+++ b/src/common/example/index.ts
@@ -1,8 +1,31 @@
 import { WidgetBase } from '@dojo/widget-core/WidgetBase';
-import { WidgetProperties } from '@dojo/widget-core/interfaces';
 import { ProjectorMixin } from '@dojo/widget-core/mixins/Projector';
 import Select from '../../select/Select';
 import { v, w } from '@dojo/widget-core/d';
+import { Registry } from '@dojo/widget-core/Registry';
+import dojoTheme from '../../themes/dojo/theme';
+
+import accordionpane from './../../accordionpane/example/index';
+import button from './../../button/example/index';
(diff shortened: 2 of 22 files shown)


src/common/example/index.html
```diff
@@ -38,25 +38,7 @@
 			}
 		}));
 
-		(function () {
-			var query = window.location.search.substring(1).split('&');
```
pottedmeat: Can we make this continue to work with a query string? Or is there a workaround I'm not thinking of to run the functional tests?

agubler: I didn't realise the functionals used the same index.html file as the examples... we'd need to resolve this.

src/common/example/index.html
```diff
@@ -38,25 +38,7 @@
...
-				}
-			}
-
-			require(['@dojo/shim/main'], function () {
-				require([ 'src/common/example/index' ], function () {});
-				if (module) {
-					require([ 'src/' + module + '/example/index' ], function () {});
-				}
-			});
-		})();
```
bitpshr: This query string functionality is used by functional tests. We'll need to add it back in or figure out a different way to run them. I'm guessing CI fails because of this.

src/common/example/index.ts
```diff
@@ -1,8 +1,31 @@
 import { WidgetBase } from '@dojo/widget-core/WidgetBase';
-import { WidgetProperties } from '@dojo/widget-core/interfaces';
 import { ProjectorMixin } from '@dojo/widget-core/mixins/Projector';
 import Select from '../../select/Select';
 import { v, w } from '@dojo/widget-core/d';
+import { Registry } from '@dojo/widget-core/Registry';
+import dojoTheme from '../../themes/dojo/theme';
+
+import accordionpane from './../../accordionpane/example/index';
```
bitpshr: I wish we had a convenience index module so this isn't so verbose. We could technically have one that just isn't built. This is rhetorical, just thinking out loud.

src/common/example/index.ts
```diff
@@ -42,12 +94,24 @@ export class App extends WidgetBase<WidgetProperties> {
...
+				'Use Dojo Theme ',
+				v('input', {
+					styles: {
+						'margin-top': '20px'
+					},
+					type: 'checkbox',
+					onchange: this.themeChange
+				})
+			]),
+			this._module ? w<any>(this._module, { theme: this._theme }) : null
```
bitpshr: Nit, but can we do better than the `<any>` cast?

src/common/example/index.html
agubler: Yeah @pottedmeat mentioned, this (seems strange that it would use the same index.html for both examples and tests imo) - we do need to think about the best way to deal with this, if we want to make the examples works as a real single projector example.

src/common/example/index.ts
agubler: Perhaps could be `ThemedMixin(WidgetBase)` but I assume I would have tried that!

Maybe need to actually make a "fake" class for it to be able to be used as a type.

```ts
class WidgetExample extends ThemedMixin(WidgetBase) {}
```

agubler:
This is waaaay out of date now so closing, but still think it's a good idea for the examples. We'd need to separate the examples and functionals as part of this.

PR #379, opened 2017-11-16
Upgrade to TypeScript 2.6.1, turn on strict, and bump minor version

bryanforbes:
**Type:** enhancement

**Description:**

Relates to dojo/meta#210

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dojo/widgets",
-  "version": "0.2.3-pre",
+  "version": "0.3.0-pre",
   "description": "Dojo 2 Widgets",
   "private": true,
   "homepage": "https://dojo.io",
@@ -21,16 +21,16 @@
     "test": "grunt test"
   },
   "peerDependencies": {
-    "@dojo/core": "~0.2.1",
-    "@dojo/has": "~0.1.0",
-    "@dojo/i18n": "~0.2.0",
-    "@dojo/shim": "~0.2.2",
-    "@dojo/widget-core": "~0.3.0"
+    "@dojo/core": "~0.3.0",
+    "@dojo/has": "~0.1.1",
+    "@dojo/i18n": "~0.3.0",
+    "@dojo/shim": "~0.2.3",
+    "@dojo/widget-core": "~0.4.0"
   },
   "devDependencies": {
-    "@dojo/interfaces": "~0.1.0",
-    "@dojo/loader": "~0.1.0",
-    "@dojo/test-extras": "~0.2.0",
+    "@dojo/interfaces": "~0.2.0",
+    "@dojo/loader": "~0.1.1",
+    "@dojo/test-extras": "~0.3.0",
     "@types/chai": "3.4.*",
     "@types/glob": "5.0.*",
     "@types/grunt": "0.4.*",
@@ -38,11 +38,10 @@
     "@types/sinon": "^1.16.31",
     "grunt": "^1.0.1",
     "grunt-dojo2": "latest",
-    "grunt-dojo2-extras": "^2.0.0-beta1.3",
     "intern": "~4.1.0",
     "load-grunt-tasks": "^3.5.2",
     "sinon": "^1.17.6",
     "ts-node": "^3.3.0",
-    "typescript": "~2.5.0"
+    "typescript": "~2.6.1"
   }
 }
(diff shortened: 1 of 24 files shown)


src/accordionpane/example/index.ts
```diff
@@ -15,8 +15,9 @@ export class App extends WidgetBase<WidgetProperties> {
 	private _openKeys = new Set<string>();
 	private _theme: {};
 
-	themeChange(event: TypedTargetEvent<HTMLInputElement>) {
```
pottedmeat: Why was this changed?

src/accordionpane/example/index.ts
```diff
@@ -15,8 +15,9 @@ export class App extends WidgetBase<WidgetProperties> {
...
-		this._theme = event.target.checked ? dojoTheme : {};
+	themeChange(event: Event) {
+		const target: HTMLInputElement = event.target as any;
+		this._theme = target.checked ? dojoTheme : {};
```
pottedmeat: This looks functionally equivalent to the button index change, but uses slightly different code.

src/textinput/example/index.ts
```diff
@@ -39,8 +39,9 @@ export class App extends WidgetBase<WidgetProperties> {
 					type: 'text',
 					placeholder: 'Hello, World',
 					value: this._value1,
-					onChange: (event: TypedTargetEvent<HTMLInputElement>) => {
-						this._value1 = event.target.value;
+					onChange: (event: Event) => {
+						const target: HTMLInputElement = event.target as any;
```
pottedmeat: This differs from the `as HTMLInputElement` we seem to be preferring elsewhere in the code

tsconfig.json
```diff
@@ -14,8 +14,6 @@
 		],
 		"module": "umd",
 		"moduleResolution": "node",
-		"noImplicitAny": true,
```
pottedmeat: Has this been replaced with a different compiler option somewhere?

src/accordionpane/example/index.ts
pottedmeat: I mean I understand it's the function covariance stuff but I'm curious if this is our only option.

package.json
```diff
@@ -21,16 +21,16 @@
     "test": "grunt test"
   },
   "peerDependencies": {
-    "@dojo/core": "~0.2.1",
-    "@dojo/has": "~0.1.0",
-    "@dojo/i18n": "~0.2.0",
-    "@dojo/shim": "~0.2.2",
-    "@dojo/widget-core": "~0.3.0"
+    "@dojo/core": "./tarballs/dojo-core-0.3.0-beta.1.tgz",
```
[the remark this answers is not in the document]
bryanforbes: No. This is only to update the widgets to use the unpublished API changes to the dependencies.

src/accordionpane/example/index.ts
bryanforbes: At this time, yes, it's our only option. @agubler and @matt-gadd talked about changing to non-bubbling events in `widget-core/vdom` (which would mean we could rely on the type of the target), so I wasn't willing to go in and update those signatures to get this type of thing to work.

src/accordionpane/example/index.ts
bryanforbes: Yeah, I wasn't consistent and should have been. I can clean that up next week.

src/textinput/example/index.ts
bryanforbes: See above.

tsconfig.json
bryanforbes: Yes, `strict` covers `noImplicitAny`, `noImplicitThis`, `alwaysStrict`, `strictNullChecks`, and `strictFunctionTypes`.

bitpshr:
A few minor conflicts. Thanks for doing this work.

agubler:
I think this needs to wait until the dependencies have had their TS2.6 upgrade merged and been released.

src/textinput/example/index.ts
```diff
@@ -39,8 +39,9 @@ export class App extends WidgetBase<WidgetProperties> {
 					type: 'text',
 					placeholder: 'Hello, World',
 					value: this._value1,
-					onChange: (event: TypedTargetEvent<HTMLInputElement>) => {
-						this._value1 = event.target.value;
+					onChange: (event: Event) => {
+						const target = event.target as HTMLInputElement;
+						this._value1 = target.value;
```
smhigley: any reason to define `target` here instead of the `this._value1 = (event.target as HTMLInputElement).value` approach used elsewhere

bryanforbes: I'll make that change

PR #388, opened 2017-11-27
Increase font size, decrease padding for dropdown arrow

### LINKED ISSUE #354: Down chevrons on select / combobox are different
**Bug** 

screenshot 2017-11-01 17 27 33

I would expect `Select` to be the same as the others

pottedmeat:
**Type:** bug

**Description:**

ComboBox now uses similar styles to Select. Resolves #354


tomdye:
Surely would be better still if the two widgets shared the same css?

pottedmeat:
Yeah, I made a note about sharing CSS in the issue. Can talk about it next widget meeting.

PR #391, opened 2017-11-28
i18n

### LINKED ISSUE #213: i18n in widgets
**Enhancement**

We need to figure out how to handle nls files for widgets that have internal string. I can see a couple different possibilities:

1. Keep everything within each individual widget
2. Put nls files in `widgets/common`, which may result in less code duplication for things like months or weekday names that could appear in multiple widgets
3. Leave it up to widget users to provide translations and pass them in via properties. This seems like a little too much work for end users, though.

pottedmeat:
**Type:** feature

**Description:**

Adds i18n to widgets and updates unit tests. Used [strings.php.zip](https://github.com/dojo/widgets/files/1511350/strings.php.zip), run from the `src` directory, to try to detect any and all strings that appear in widgets.

Resolves #213


src/calendar/Calendar.ts
```diff
@@ -90,6 +91,14 @@ export default class Calendar extends CalendarBase<CalendarProperties> {
 		return lastDate.getDate();
 	}
 
+	private _getMonths() {
+		const messages = this.localizeBundle(commonBundle);
```
bitpshr: Can this call (and the one below in `_getWeekdays`) be moved to a member variable so it doesn't need to be called every render? This may not be possible depending on how components are notified of i18n data changes...

src/calendar/Calendar.ts
```diff
@@ -307,8 +324,8 @@ export default class Calendar extends CalendarBase<CalendarProperties> {
 
 	protected renderDatePicker(): DNode {
 		const {
-			labels = DEFAULT_LABELS,
-			monthNames = DEFAULT_MONTHS,
+			labels = this.localizeBundle(calendarBundle),
```
bitpshr: Same comment re: potentially caching this call.

src/dialog/tests/unit/Dialog.ts
```diff
@@ -32,7 +32,11 @@ const expectedCloseButton = function() {
 };
 
 const expected = function(widget: Harness<Dialog>, open = false, closeable = false, children: any[] = []) {
-	return v('div', { classes: css.root }, open ? [
+	return v('div', {
+		classes: css.root,
+		dir: null,
```
bitpshr: Are these added by the `I18NMixin`?

mwistrand: Yes, the `I18nMixin` adds both `lang` and `dir` attributes.

src/calendar/Calendar.ts
pottedmeat: I think it's already cached in the mixin. I'm going to pull Matt W in.

mwistrand: After they've been loaded, messages are cached, so subsequent calls to `localizeBundle` will immediately return the cached messages without requiring an additional resource request. That said, with a future release message loading will be [handled by the bundles themselves](https://github.com/dojo/i18n/issues/106), so it might be best to make that request once and then pass in the result to private methods like `_getMonths` to avoid making multiple requests for the same unloaded bundle.

src/combobox/nls/ComboBox.ts
```diff
@@ -0,0 +1,9 @@
+const bundlePath = 'src/combobox/nls/common';
+const locales: string[] = [];
+
+const messages = {
+	clear: 'clear combo box',
```
tomdye: possibly a question for @smhigley . Do these need to explicitly say `combo box` or could they be moved to common and simply read: `clear`, `open`, `close` etc...? Would save repetition.

smhigley:
Closing via #405

PR #393, opened 2017-11-28
TextInput Addons

### LINKED ISSUE #91: Addon support for TextInput
# Specification
This widget builds on the existing `textInput` widget and allow the user to specify an 'addon'.

Below is an example of `addons` from bootstrap. It shows leading, trailing and multiple addons:

screen shot 2017-03-29 at 2 05 26 pm
 
## Features
- You can have both leading and trailing addons on a single `textInput`.
- You can specify text, icon, or text + icon for either leading or trailing addons.
- You can include form widgets such as a `checkbox` or `radioCheckbox` in an addon.

builds on the existing `textInput` and offers visual prompts to aid completion, e.g. type of values accepted 

none

possibility to show different tooltips when hovering over leading and trailing addons

none

as per `textInput`

**List of any icons needed**
any icon in font-awesome that we ship should be usable as either a leading or trailing icon.

**Design input required**
Need to consider how we style an invalid or mandatory input that has addons. 
Need to consider using addons and putting a `textInput` into a wait state.
Need to consider how we style addons if the `textInput` is disabled.

**Any other considerations**:
Initial support will not include changing the icon based on validity of the `textInput`.
With an input of type `search`, will Addon provide all we need to effectively offer a `Search` widget?

## Questions
Need to consider if the icons should be `spoken` as they are included for informative purposes.
Need to consider what happens if long text is specified for an addon. Do we ellid the text, always show, ..

## Acceptance criteria
When I specify an addon with text and an invalid icon, then I expect the text to be rendered without the icon.
When I specify an addon with unicode text, then I expect it to render correctly (including adjusting for height of non-latin chars).

smhigley:
**Type:** feature

**Description:**

Resolves #91 

Depends on #319 for the build to pass


src/decoratedtextinput/DecoratedTextInput.ts
```diff
@@ -0,0 +1,33 @@
...
+import TextInput, { TextInputProperties } from '../textinput/TextInput';
+import * as css from './styles/decoratedtextinput.m.css';
+
+export interface DecoratedTextInputProperties extends TextInputProperties {
+	addonAfter: DNode[];
+	addonBefore: DNode[];
+}
+
+@theme(css)
+export default class DecoratedTextInput extends TextInput<DecoratedTextInputProperties> {
```
bitpshr: I hate to be that guy, but the name `DecoratedTextInput` feels weird to me since we're overloading the term "decorated" in the context of ES.

src/decoratedtextinput/styles/decoratedtextinput.m.css
```diff
@@ -0,0 +1,18 @@
+@import '../../common/styles/variables.css';
+
+.addon {
```
bitpshr: Should we be making any classes used in the base styles `fixed` classes? At least ones that the user wouldn't want to override. Maybe none of these are applicable.

src/decoratedtextinput/README.md
```diff
@@ -0,0 +1,54 @@
...
+## Features
+
+- Allows any number of addons to be inserted before or after the input
+- Includes all TextInput functionality
+
+### Accessibility Features
+
+The `DecoratedTextInput` does not add any built-in accessibility features above what is available in `TextInput`. If this widget is used, it should be done with careful consideration of what instructional text is needed to convey the information visually implied by the addon or addons. This should be done in the label text, and optionally supplemented by custom descriptive text and `aria-describedby`.
+
+After careful consideration we decided not to automatically point the input's `aria-describedby` attribute to the `id` or `id`s of the addons, as Bootstrap does in their examples. In all the cases we examined, the text of the addon was insufficient to describe the input required of the user, and in many cases it was actually more confusing.
```
bitpshr: SHOTS FIRED

src/decoratedtextinput/styles/decoratedtextinput.m.css
smhigley: I thought about it, since the flexbox styles could be seen as integral to the display. But then I thought another possibility might be someone who wants the addon text to display within the text input box with absolute positioning and padding.

Maybe that's not enough of a common use case to cater to it (or any other weird style variation). I'm not sure, and I'm open to other opinions.

src/decoratedtextinput/DecoratedTextInput.ts
smhigley: I changed it to `EnhancedTextInput`. DEBATE AWAY

src/decoratedtextinput/README.md
smhigley: Hopefully the updated readme text isn't as full of bullets 😂

src/enhancedtextinput/README.md
```diff
@@ -0,0 +1,54 @@
+# @dojo/widgets/enhancedtextinput/EnhancedTextInput widget
+
+Dojo 2's `EnhancedTextInput` widget extends `TextInput` to provide optional addons before or after the input, similar to those in (bootstrap)[https://v4-alpha.getbootstrap.com/components/input-group/#basic-example].
```
pottedmeat: The rounded and square parens need to be swapped to make this link work.

src/enhancedtextinput/README.md
```diff
@@ -0,0 +1,54 @@
...
+	value: this.state.price,
+	onChange: (event: TypedTargetEvent<HTMLInputElement>) => {
+		this.setState({ price: event.target.value });
+	}
+});
+```
+
+## Theming
+
+The following CSS classes are available on the `EnhancedTextInput` widget for use with custom themes:
```
pottedmeat: Also `addonBefore` and `addonAfter`.

PR #394, opened 2017-11-29
385 - Progress Widget

### LINKED ISSUE #385: Progress Component
**Enhancement**

A Progress component should be created that can be used to indicate the progress of a given operation based on a property. Styling should be extrapolated based on other existing components for the initial version.

Package Version: beta5

tomdye:
**Type:** bug / feature

**Description:**
Creates progress bar widget

screenshot 2017-11-29 17 22 34

Resolves #385

### DIFF
--- a/src/common/tests/unit/all.ts
+++ b/src/common/tests/unit/all.ts
@@ -18,6 +18,7 @@ import '../../../dialog/tests/unit/Dialog';
 import '../../../label/tests/unit/Label';
 import '../../../listbox/tests/unit/Listbox';
 import '../../../listbox/tests/unit/ListboxOption';
+import '../../../progress/tests/unit/Progress';
 import '../../../radio/tests/unit/Radio';
 import '../../../select/tests/unit/Select';
 import '../../../slidepane/tests/unit/SlidePane';
--- a/src/progress/Progress.ts
+++ b/src/progress/Progress.ts
@@ -2,6 +2,7 @@ import { WidgetBase } from '@dojo/widget-core/WidgetBase';
 import { ThemedMixin, ThemedProperties, theme } from '@dojo/widget-core/mixins/Themed';
 import { v } from '@dojo/widget-core/d';
 import * as css from './styles/progress.m.css';
+import { HNode } from '@dojo/widget-core/interfaces';
 
 /**
  * @type ProgressProperties
@@ -31,7 +32,7 @@ export default class Progress extends ProgressBase<ProgressProperties> {
 		return output ? output(value, percent) : `${percent}%`;
 	}
 
-	protected renderProgress(percent: number) {
+	protected renderProgress(percent: number): HNode[] {
 		return [
 			v('div', {
 				classes: this.theme(css.progress),
--- a/src/progress/Progress.ts
+++ b/src/progress/Progress.ts
@@ -9,11 +9,11 @@ import { HNode } from '@dojo/widget-core/interfaces';
  *
  * Properties that can be set on a Progress component
  *
- * @property value            The current value
- * @property output...........A function used to determine the output display
- * @property showOutput.......Toggles visibility of progess bar output
- * @property max..............Value used to calculate percent width
- * @property min..............Value used to calculate percent width
+ * @property value          The current value
+ * @property output         A function used to determine the output display
+ * @property showOutput     Toggles visibility of progess bar output
+ * @property max            Value used to calculate percent width
+ * @property min            Value used to calculate percent width
  */
 export interface ProgressProperties extends ThemedProperties {
 	value: number;
--- a/src/progress/Progress.ts
+++ b/src/progress/Progress.ts
@@ -1,5 +1,4 @@
 import { WidgetBase } from '@dojo/widget-core/WidgetBase';
-import { DNode } from '@dojo/widget-core/interfaces';
 import { ThemedMixin, ThemedProperties, theme } from '@dojo/widget-core/mixins/Themed';
 import { v } from '@dojo/widget-core/d';
 import * as css from './styles/progress.m.css';
@@ -43,7 +42,7 @@ export default class Progress extends ProgressBase<ProgressProperties> {
 		];
 	}
 
-	render(): DNode {
+	render() {
 		const {
 			value,
 			showOutput = true,
--- a/src/progress/Progress.ts
+++ b/src/progress/Progress.ts
@@ -63,7 +63,7 @@ export default class Progress extends ProgressBase<ProgressProperties> {
 				'aria-valuenow': `${value}`,
 				'aria-valuetext': output
 			}, this.renderProgress(percent)),
-			showOutput ? v('span', [ output ]) : null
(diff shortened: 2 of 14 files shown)


src/progress/Progress.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @property value            The current value
+ * @property output...........A function used to determine the output display
+ * @property showOutput.......Toggles visibility of progess bar output
+ * @property max..............Value used to calculate percent width
+ */
+export interface ProgressProperties extends ThemedProperties {
+	value: number;
+	output?(value: number, percent: number): string;
+	showOutput?: boolean;
+	max?: number;
```
smhigley: Although I expect `min` will almost always be 0, we should probably still support `min`

src/progress/Progress.ts
```diff
@@ -0,0 +1,54 @@
...
+		const {
+			value,
+			showOutput = true,
+			max = 100
+		} = this.properties;
+
+		const percent = Math.round((value / max) * 100);
+
+		return v('div', { classes: this.theme(css.root) }, [
+			v('div', { classes: this.theme(css.bar) }, [
```
smhigley: Additional properties on the bar:
```
role: 'progressbar',
'aria-valuemin': `${min}`,
'aria-valuemax': `${max}`,
'aria-valuenow': `${value}`,
'aria-valuetext`: output ? `${output(value, percent)}` : null
```

There are some other attributes that would need to be added and controlled if the `aria-*` thing falls through, like `aria-labelledby`, `aria-live` (probably also aria-atomic). It's possible we should add them now, so we don't forget and so it isn't lacking in the meantime.

smhigley: Also the bit with `aria-valuetext` doesn't draw from `this._output` because it should only be present if custom text is supplied

src/progress/Progress.ts
```diff
@@ -0,0 +1,54 @@
...
+				v('div', {
+					classes: this.theme(css.progress),
+					styles: {
+						width: `${percent}%`
+					}
+				})
+			]),
+			showOutput ? v('output', [ this._output(value, percent) ]) : null
```
smhigley: It's a little ambiguous, but I don't think this actually needs to be an `output` element, sine it's debatable whether it's showing a calculation from a form, and the same information is presented as part of the progressbar itself with `aria-valuenow` and (optionally) `aria-valuetext`. We might potentially even add `aria-hidden="true"` to prevent the value from being read twice. I'm going to double check to see if it's necessary first though.

smhigley:
One other thought: would it make sense to split out a `renderProgress(percentComplete: number)` function so it's easy to extend and add ticks or discrete steps or something?

src/progress/Progress.ts
tomdye: ok, not a problem

src/progress/Progress.ts
tomdye: Thanks for these

tomdye: re. `aria-valuetext`, i think that it *should* draw from the `this._output` function because it will at least always add `%` to the value so it will be different to the value itself.

src/common/tests/unit/all.ts
```diff
@@ -18,6 +18,7 @@ import '../../../dialog/tests/unit/Dialog';
 import '../../../label/tests/unit/Label';
 import '../../../listbox/tests/unit/Listbox';
 import '../../../listbox/tests/unit/ListboxOption';
+import '../../../progress/tests/unit/Progress';
```
bitpshr: I think we need a functional test for this component, even if it's just verifying it renders something visible on the page. Otherwise someone will have to retroactively (and painfully) add one.

src/progress/Progress.ts
```diff
@@ -0,0 +1,54 @@
...
+			showOutput = true,
+			max = 100
+		} = this.properties;
+
+		const percent = Math.round((value / max) * 100);
+
+		return v('div', { classes: this.theme(css.root) }, [
+			v('div', { classes: this.theme(css.bar) }, [
+				v('div', {
+					classes: this.theme(css.progress),
```
bitpshr: Did we create an actual theme for this? I know we don't have mock ups, but I think we need to take the styling as far as we can. We should create applicable theme CSS files and extrapolate a theme that looks decent, per Scott's advice. We should at least get the infrastructure and files in place rather than only relying on structural styles.

src/common/tests/unit/all.ts
tomdye: Is there really any point doing a functional for this widget given that its not interactive and is fully covered by unit tests. All we'd be doing is checking that widget-core is working?

smhigley: I didn't think so, as I didn't do one for the addons either (same thing, no interactivity). Definitely open to being convinced, though

src/progress/Progress.ts
smhigley: I think the `aria-valuetext` thing should be OK with `this._output`. Screenreaders do actually add "percent" when no value text is given, and I usually tend towards adding the least aria possible. On the other hand, always using `this._output` is cleaner and it should be the same experience. I'll double check w/ NVDA and if that reads fine then I'm good leaving off the extra conditional.

src/progress/Progress.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as css from './styles/progress.m.css';
+import { HNode } from '@dojo/widget-core/interfaces';
+
+/**
+ * @type ProgressProperties
+ *
+ * Properties that can be set on a Progress component
+ *
+ * @property value            The current value
+ * @property output...........A function used to determine the output display
```
bitpshr: What's up with the periods?

tomdye: thats a bit odd...

tomdye:
@smhigley comments all addressed

src/progress/Progress.ts
```diff
@@ -0,0 +1,69 @@
...
+		const output = this._output(value, percent);
+
+		return v('div', { classes: this.theme(css.root) }, [
+			v('div', {
+				classes: this.theme(css.bar),
+				role: 'progressbar',
+				'aria-valuemin': `${min}`,
+				'aria-valuemax': `${max}`,
+				'aria-valuenow': `${value}`,
+				'aria-valuetext': output
```
smhigley: One last thing I missed before: we should add an `id` prop as well. A common use case will probably be a progress bar for a container while loading, in which case the container should have `aria-describedby` pointing at the `id` of the progress bar.

tomdye: ok, fair enough. good catch

src/progress/createProgressElement.ts
```diff
@@ -0,0 +1,37 @@
...
+			{
+				attributeName: 'max',
+				value: value => Number(value)
+			},
+			{
+				attributeName: 'min',
+				value: value => Number(value)
+			},
+			{
+				attributeName: 'min',
```
smhigley: Duplicate `min` attributes. I think one should be `id`, but just wasn't changed?

tomdye: ah oops, thanks

PR #397, opened 2017-11-30
Toolbar

### LINKED ISSUE #386: Toolbar Component
**Enhancement**

A Toolbar component should be created that eases the development of standard navigation bars. It should conditionally show children inside of a SlidePane component based on a property and support an opinionated but themable default style. We should look to the Monster card application to see what features (if any) were used in that toolbar implementation. For this initial version, styling should be extrapolated based on other existing components.

Package Version: beta5

bitpshr:
**Type:** feature

**Description:**

This PR implements an opinionated Toolbar component that can configured to use either fixed or static positioning. Users can provide a `title` element and `action` elements. Action elements will optionally be rendered in a `SlidePane` when the component width is less than the `collapseWidth`. Both the title and actions can be any `DNode`, which means events can be attached directly to those elements instead of being propagated up through the Toolbar itself.

Example

Resolves #386


src/common/example/index.ts
```diff
@@ -32,7 +33,7 @@ export class App extends WidgetBase<WidgetProperties> {
 	}
 
 	render() {
-		return v('div', [
+		return v('div', { id: 'module-select' }, [
```
bitpshr: The `Toolbar` uses a flexbox-based layout wherein page content should be a child of the toolbar itself. This negates the need for any `padding-top` on the `<body>` element to account for a fixed-positioned toolbar, and allows the toolbar to be an arbitrary height while still allowing page content to scroll independently.

Unfortunately, because our example pages render this module select UI _before_ any component is rendered, it's impossible to place this inside the `Toolbar` example. To get around this, the toolbar example page adds a single CSS rule to the example page to style the select in a more sane manner, hence this added ID.

src/toolbar/Toolbar.ts
```diff
@@ -0,0 +1,158 @@
...
+import * as css from './styles/toolbar.m.css';
+import * as iconCss from '../common/styles/icons.m.css';
+
+/**
+ * @type ToolbarProperties
+ *
+ * Properties that can be set on a Toolbar component
+ *
+ * @property actions           Action elements to show in the toolbar
+ * @property collapseWidth     Width at which to collapse actions into a SlidePane
```
bitpshr: I feel strongly that we shouldn't _always_ be inverting control to the parent for every piece of UI functionality. It's both tedious and common enough to hide action items in a SlidePane that our opinionated component should provide this, while also allowing easy extension. In practice, using a property like `collapseWidth` instead of `collapsed` and internalizing window resizing logic felt much cleaner.

src/toolbar/Toolbar.ts
```diff
@@ -0,0 +1,158 @@
...
+@theme(iconCss)
+export default class Toolbar extends ThemedBase<ToolbarProperties> {
+	private _collapsed = false;
+	private _onResize: () => void;
+	private _open = false;
+
+	constructor() {
+		super();
+		this._onResize = debounce(this._collapseIfNecessary, 250);
+		window.addEventListener('resize', this._onResize);
```
bitpshr: This is stopgap until dojo/widget-core#618 lands. For now, we do the best we can by debouncing a resize handler that's explicitly removed on detachment.

src/toolbar/example/index.ts
```diff
@@ -0,0 +1,51 @@
...
+export class App extends WidgetBase<WidgetProperties> {
+	private _theme: {};
+
+	themeChange(event: TypedTargetEvent<HTMLInputElement>) {
+		this._theme = event.target.checked ? dojoTheme : {};
+		this.invalidate();
+	}
+
+	onAttach() {
+		const style = document.createElement('style');
```
bitpshr: This method is purely for demo styling purposes. See above comment re: Toolbar and its children.

src/toolbar/styles/toolbar.m.css
```diff
@@ -0,0 +1,59 @@
...
+	flex-grow: 0;
+	flex-shrink: 0;
+}
+
+.stickyFixed .contentFixed {
+	flex: 1;
+	overflow-y: auto;
+}
+
+.action { }
```
bitpshr: We may not need all these classes to be themeable, but it felt right to expose as many of them as possible.

src/toolbar/Toolbar.ts
```diff
@@ -0,0 +1,158 @@
...
+		}, [ action ]));
+
+		if (actionsElements.length === 0) {
+			return null;
+		}
+
+		return this._collapsed ? w(SlidePane, {
+			align: Align.right,
+			closeText: 'close menu',
+			key: 'menu',
```
tomdye: not sure you should have two elements with the same `key`

src/toolbar/Toolbar.ts
```diff
@@ -0,0 +1,158 @@
...
+				classes: this.theme([ iconCss.icon, iconCss.barsIcon ]),
+				role: 'presentation', 'aria-hidden': 'true'
+			})
+		]) : null;
+	}
+
+	protected renderTitle(): DNode {
+		const { title } = this.properties;
+
+		return title ? v('div', {
```
tomdye: no `h1` ?

src/toolbar/Toolbar.ts
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	protected getFixedModifierClasses(): (string | null)[] {
+		return [
+			css.rootFixed,
+			this.properties.fixed ? css.stickyFixed : null
+		];
+	}
+
+	protected getModifierClasses(): (string | null)[] {
```
tomdye: renamed these to `getRootClasses` elsewhere

src/themes/dojo/toolbar.m.css
```diff
@@ -0,0 +1,38 @@
...
+
+.menuButton {
+	background: none;
+	border: none;
+	cursor: pointer;
+	font-size: 0;
+	padding: var(--grid-base);
+	position: absolute;
+	right: var(--grid-base);
+	top: 50%;
```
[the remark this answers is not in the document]
bitpshr: Sort of, but it would require one extra element. Since I'm already absolutely positioning the button to the right it was a bit cleaner (and more uniform) to just use `top` and `transform`.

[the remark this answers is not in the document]
tomdye: what about when our users want to use flexbox? If adding an extra element to make this simpler and removes absolute positioning etc is it not worth it?

src/toolbar/styles/toolbar.m.css
```diff
@@ -0,0 +1,59 @@
...
+	right: 0;
+	top: 50%;
+	transform: translateY(-50%);
+}
+
+.rootFixed {
+	height: 100%;
+	left: 0;
+	overflow: hidden;
+	position: fixed;
```
smhigley: Does this need to be fixed? It also works with `position: absolute`, and then it isn't constrained to only being document-level (although a toolbar on anything else seems weird, granted).

src/toolbar/Toolbar.ts
```diff
@@ -0,0 +1,159 @@
...
+		window.removeEventListener('resize', this._collapseIfNecessary);
+	}
+
+	protected renderActions(): DNode {
+		const { actions = [], theme } = this.properties;
+
+		const actionsElements = actions.map((action, index) => v('div', {
+			classes: [ this.theme(css.action) ],
+			key: index
+		}, [ action ]));
```
smhigley: I don't really understand why these need to be wrapped in a div. Is it just so their css/theming is within the toolbar CSS as opposed to in parent-widget-level CSS?

src/toolbar/Toolbar.ts
```diff
@@ -0,0 +1,159 @@
...
+			})
+		]) : null;
+	}
+
+	protected renderTitle(): DNode {
+		const { title } = this.properties;
+
+		return title ? v('h1', {
+			classes: [ this.theme(css.title), css.titleFixed ]
+		}, [ title ]) : null;
```
smhigley: I wonder if this should really be an `h1`. Usually that should go to a content-specific title, not to a company name/logo, which is more common for toolbars. Since `title` is already a `DNode`, could we just return that, wrapped in a `div` for flex styling?

src/toolbar/tests/functional/Toolbar.ts
```diff
@@ -0,0 +1,47 @@
...
+			.findByCssSelector('body > div:last-child > div:first-child > div:nth-child(2) > div button')
+				.click()
+			.end()
+			.sleep(DELAY)
+			.findByCssSelector('body > div:last-child > div:first-child > div:nth-child(2) > div')
+				.isDisplayed()
+				.then(displayed => {
+					assert.isFalse(displayed);
+				});
+	}
```
smhigley: I know functional tests are a pain, but maybe one to make sure the slidepane is shown or not shown based on window width would be good

src/tooltip/Tooltip.ts
```diff
@@ -12,7 +12,7 @@ import * as css from './styles/tooltip.m.css';
  *
  * @property content           Information to show within the tooltip
  * @property orientation       Where this tooltip should render relative to its child
- * @property open           Determines if this tooltip is visible
+ * @property open              Determines if this tooltip is visible
```
smhigley: lol

src/toolbar/Toolbar.ts
```diff
@@ -0,0 +1,159 @@
...
+			v('div', {
+				classes: [ this.theme(css.toolbar), css.toolbarFixed ]
+			}, [
+				this.renderTitle(),
+				this.renderActions(),
+				this.renderButton()
+			]),
+			v('div', {
+				classes: [ this.theme(css.content), css.contentFixed ]
+			}, this.children)
```
smhigley: Not sure if this needs to be added now, but reversing the order of the toolbar and content based on a `position` property would be easy, and having the option of a bottom toolbar is nice. Works with the styling as-is, too.

src/toolbar/Toolbar.ts
smhigley: p.s.: really like the breakdown of the render fn :)

src/toolbar/Toolbar.ts
```diff
@@ -0,0 +1,159 @@
...
+			key: 'menu'
+		}, actionsElements);
+	}
+
+	protected renderButton(): DNode {
+		return this._collapsed ? v('button', {
+			classes: [ this.theme(css.menuButton), css.menuButtonFixed ],
+			onclick: this._toggleMenu
+		}, [
+			'open menu',
```
smhigley: You probably already know this, but depending on the order of merging of this and Neil's i18n PR, the 'open menu'/'close menu' text should either be changed here, or Neil should know that you added some more strings for his PR

src/toolbar/tests/functional/Toolbar.ts
smhigley: ^ ignore that if you want, or if you already tried and removed it in frustration

src/toolbar/Toolbar.ts
bitpshr: Yes, this is really just for easier theming. Most situations I thought of were pretty difficult to accomplish without actions being grouped in a common parent. I'm open to removing it if necessary though.

src/toolbar/styles/toolbar.m.css
bitpshr: Good catch. The `fixed` styling was left over from the approach that didn't rely on content being a child of the Toolbar.

src/toolbar/tests/functional/Toolbar.ts
bitpshr: I think this is a good idea, but I had the hardest time getting functional tests to run locally. I can open a new issue for this if that works for you.

bitpshr:
@smhigley This is ready for another review when you have a chance.

src/toolbar/Toolbar.ts
bitpshr: Added!

src/toolbar/Toolbar.ts
smhigley: Hm, I kind of imagine most of the time the actions will be passed in already wrapped in a parent, for semantics and styling. E.g.:
```
w(Toolbar, {
   actions: [
      v('div', { id: 'menu', role: 'menubar' }, [
         v('a', { href: '/#home' }, [ 'Home' ]),
         v('a', { href: '/#about' }, [ 'About' ]),
         v('a', { href: '/#contact' }, [ 'Contact' ])
      ])
   ],
   title: 'Foobar'
}, [
   content
])
```
In that case the extra wrapping div seems a little unnecessary. I don't really dislike it, I just think it introduces an extra wrapping element where none is needed.

src/toolbar/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Features
+
+- Capable of auto-collapsing action elements into `SlidePane`
+- Capable of rendering either a fixed or statically positioned root element
+- Customizable `title` element
+
+### Accessibility Features
+
+- The menu button displays screen-readable text that's visually hidden and its icon has an `aria-hidden` attribute
```
smhigley: Thinking about the menu wrapping, it might also be good to add some text like this:
> Toolbar makes no assumptions about what content will be inserted through `actions`. If you wish to create a site-wide navigation section, it would be best to wrap links in an element with the appropriate semantics (e.g. `<nav>` or `role="navigation"`)

Issue #223, opened 2017-06-06
Add nls files to Calendar


smhigley:
**Enhancement**

Calendar should handle its own nls files. Implementation will depend on the resolution of #213.

smhigley:
Resolved in #405

PR #405, opened 2017-12-14
Enhancement i18n

### LINKED ISSUE #213: i18n in widgets

smhigley:
**Type:** bug / feature

**Description:**

from @pottedmeat's PR:
> Description:
> 
> Adds i18n to widgets and updates unit tests. Used strings.php.zip, run from the src directory, to try to detect any and all strings that appear in widgets.

Resolves #213 and #223


src/combobox/nls/ComboBox.ts
```diff
@@ -0,0 +1,9 @@
+const bundlePath = 'src/combobox/nls/common';
+const locales: string[] = [];
+
+const messages = {
+	clear: 'clear combo box',
```
tomdye: Same question I asked Neil, do we really need messages of this granularity, we could replace all the `open` / `clear` / `toggle` etc messages with a single `commonBundle`. Seems that we have a lot of repetition otherwise. If people want a custom message then I believe they can pass it as a widget property no?

mwistrand: I tend to agree that messages should be injected into any component where the developer cannot directly modify the messages bundle. It would still be possible to set custom locale-specific messages, but doing so requires using `@dojo/i18n/i18n.setLocaleMessages` which isn't ideal, or users will try modifying the `locales` property of the bundle itself, which we also don't want.

smhigley: @tomdye I actually think it needs more than `open`/`clear`/`toggle`. If you're tabbing through without visual guidance, it's not always clear what controls are related to what widget. For the dialog, I played with the idea of something like `open ${title} dialog`, or even maybe `open ${title}`. I don't think that would get us to a single common bundle because ComboBox doesn't have a user-created title, and dialog's title is optional.

I'm open to thinking more about the organization though, if you think it's warranted. We could at least perhaps consolidate the words "open" and "close"

src/calendar/Calendar.ts
```diff
@@ -359,9 +374,12 @@ export default class Calendar<P extends CalendarProperties = CalendarProperties>
 	}
 
 	protected render(): DNode {
+		const calendarMessages = this.localizeBundle(calendarBundle);
+		const commonMessages = this.localizeBundle(commonBundle);
```
mwistrand: It would be ideal if these were only loaded when their corresponding properties are unavailable. As it stands now, locale-specific messages will always be loaded, possibly via another HTTP request, regardless of whether they are actually needed.

src/calendar/nls/Calendar.ts
```diff
@@ -0,0 +1,13 @@
...
+const messages = {
+	chooseMonth: 'Choose Month',
+	chooseYear: 'Choose Year',
+	previousMonth: 'Previous Month',
+	nextMonth: 'Next Month',
+	previousYears: 'Earlier years',
+	nextYears: 'Later years'
+};
+
+export default { bundlePath, locales, messages };
```
mwistrand: The most recent releases of `widget-core` (0.5.0) and `i18n` (0.4.0) change the way bundles are formatted. The `bundlePath` is no longer necessary, and `locales` is now a map of loaders:

```typescript
export default {
    locales: {
        es: () => import('./es/common') // or `require('./es/common')` where tslint dislikes `import()`
    },
    messages: { /* ... */ }
}
```

(Sorry 😖)

src/common/nls/common.ts
```diff
@@ -0,0 +1,48 @@
...
+	september: 'September',
+	october: 'October',
+	november: 'November',
+	december: 'December',
+	clear: 'clear',
+	close: 'close',
+	open: 'open'
+};
+
+export default { bundlePath, locales, messages };
```
mwistrand: Same comment as above.

src/calendar/nls/Calendar.ts
smhigley: Leaving this for when we update to `widget-core: 0.5.0`, as `test-extras` needs a new release first.

smhigley:
Received approval from Matt W via chat

Issue #376, opened 2017-11-15
Extend textinput for combobox


tomdye:
The current Implementation of ComboBox needs to pass extra fields to it's input widget to facilitate accessibility.
This never actually worked and was masked by the use of an `<any>` typing. If this still needs to happen we should create an extended `ComboBoxTextInput` which accepts these extra properties.

Code below shows commented out fields that are passed currently:

```ts
protected renderInput(): DNode {
		const {
			// clearable,
			disabled,
			inputProperties = {},
			invalid,
			readOnly,
			required,
			// results = [],
			value = '',
			theme
		} = this.properties;

		// TODO: Likely need to extend TextInput here to add in these items
		return w(TextInput, {
			...inputProperties,
			key: 'textinput',
			// 'aria-activedescendant': this._getResultId(results[this._activeIndex], this._activeIndex),
			// 'aria-owns': this._getMenuId(),
			// classes: this.theme(clearable ? css.clearable : null),
			controls: this._getMenuId(),
			disabled,
			invalid,
			onBlur: this._onInputBlur,
			onFocus: this._onInputFocus,
			onInput: this._onInput,
			onKeyDown: this._onInputKeyDown,
			readOnly,
			required,
			theme,
			value
		});
	}
```

smhigley:
Resolved via #410

PR #410, opened 2017-12-28
Custom aria interface and implementation

### LINKED ISSUE #392: Ability to add custom aria to widgets
**Enhancement**

Given the wide range of possibilities for valid ARIA semantics on any given widget, it seems useful to  provide a way to allow any `aria-*` attribute to dojo widgets. One example of the need for this is #376. Needing to add `aria-activedescendant`, `aria-controls`, and `aria-owns` to a TextInput is not a common enough use case to justify adding them to the properties interface, but it's also not uncommon to need some sort of specialized ARIA on a basic widget.

An example of a possible solution that adds a common `CustomAriaProperties` interface and leaves the implementation up to widget authors is here: https://github.com/dojo/widgets/compare/master...smhigley:aria-properties (this POC only covers TextInput/Combobox).

If [this Typescript issue](https://github.com/Microsoft/TypeScript/issues/6579) ever lands, the interface could be simplified to allow `aria-*` properties without explicitly naming all of them.

smhigley:
**Type:** feature

**Description:**

Resolves #392 and #376. Adds an interface for a custom `aria` bucket, and a util to format properties passed through it. Leaves implementation up to widget authors.

### DIFF
--- a/src/button/Button.ts
+++ b/src/button/Button.ts
@@ -4,7 +4,8 @@ import { ThemedMixin, ThemedProperties, theme } from '@dojo/widget-core/mixins/T
 import { v } from '@dojo/widget-core/d';
 import * as css from '../theme/button/button.m.css';
 import * as iconCss from '../theme/common/icons.m.css';
-import { InputEventProperties, PointerEventProperties, KeyEventProperties } from '../common/interfaces';
+import { CustomAriaProperties, InputEventProperties, PointerEventProperties, KeyEventProperties } from '../common/interfaces';
+import { formatAriaProperties } from '../common/util';
 
 export type ButtonType = 'submit' | 'reset' | 'button' | 'menu';
 
@@ -13,7 +14,6 @@ export type ButtonType = 'submit' | 'reset' | 'button' | 'menu';
  *
  * Properties that can be set on a Button component
  *
- * @property describedBy    ID of element with descriptive text
  * @property disabled       Whether the button is disabled or clickable
  * @property popup       		Controls aria-haspopup, aria-expanded, and aria-controls for popup buttons
  * @property name           The button's name attribute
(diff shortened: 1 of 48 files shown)


mwistrand:
Does `Label` really not need `CustomAriaProperties`?

src/button/Button.ts
```diff
@@ -75,7 +74,7 @@ export default class Button<P extends ButtonProperties = ButtonProperties> exten
 
 	render(): DNode {
 		let {
-			describedBy,
+			aria = {},
```
tomdye: Are passed in `aria` properties meant to override defaults? 
Should passed in `aria` not be merged with default and *then* passed to `formatAriaProperties` ?

ie. 

``` ts
render() {
	const defaultAria = {
		haspopup: popup ? 'true' : null,
		controls: popup ? popup.id : null
	}

	const { aria = {} } = this.properties

	...formatAriaProperties({...defaultAria, ...aria})
}
```

tomdye: just a thought.

src/combobox/tests/unit/ComboBox.ts
```diff
@@ -52,7 +52,11 @@ const getExpectedControls = function(widget: Harness<ComboBox>, useTestPropertie
 	}, [
 		w(TextInput, <any> {
 			key: 'textinput',
-			controls: <any> compareId,
+			aria: {
+				activedescendant: <any> compareId,
```
tomdye: are these `<any>` needed?

src/common/util.ts
```diff
@@ -12,3 +12,11 @@ export const enum Keys {
 	Tab = 9,
 	Up = 38
 }
+
+export function formatAriaProperties(aria: { [key: string]: string }): { [key: string]: string } {
```
tomdye: this should use the same typing at the widget properties

src/common/util.ts
```diff
@@ -12,3 +12,11 @@ export const enum Keys {
...
+	const formattedAria: { [key: string]: string } = {};
+	Object.keys(aria).forEach((key: string) => {
```
tomdye: might be better as a `reduce`

src/combobox/tests/unit/ComboBox.ts
smhigley: unfortunately yeah, but should probably be `as any`

src/button/Button.ts
smhigley: I was actually thinking passed-in aria should not override the in-widget stuff, but if you can think of a good use case for overriding, I'm open to changing it

PR #431, opened 2018-01-17
Add `id` property to form widgets

### LINKED ISSUE #430: Form widgets lack the ability to specify `id`
If someone wished to point a custom `<label>` at an existing form field widget, they would be unable to do so since we don't provide an `id` property on most of them.

smhigley:
**Type:** feature

**Description:**

Resolves #430, adds an `id` prop to form widgets so anyone wishing to use their own label may do so

### DIFF
--- a/src/checkbox/Checkbox.ts
+++ b/src/checkbox/Checkbox.ts
@@ -21,6 +21,7 @@ import * as css from '../theme/checkbox/checkbox.m.css';
  */
 export interface CheckboxProperties extends ThemedProperties, InputProperties, LabeledProperties, InputEventProperties, KeyEventProperties, PointerEventProperties, CustomAriaProperties {
 	checked?: boolean;
+	id?: string;
 	mode?: Mode;
 	offLabel?: DNode;
 	onLabel?: DNode;
@@ -114,6 +115,7 @@ export class CheckboxBase<P extends CheckboxProperties = CheckboxProperties> ext
 			aria = {},
 			checked = false,
 			disabled,
+			id = this._uuid,
 			invalid,
 			label,
 			labelAfter = true,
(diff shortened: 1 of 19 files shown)


src/checkbox/Checkbox.ts
```diff
@@ -21,6 +21,7 @@ import * as css from '../theme/checkbox/checkbox.m.css';
  */
 export interface CheckboxProperties extends ThemedProperties, InputProperties, LabeledProperties, InputEventProperties, KeyEventProperties, PointerEventProperties, CustomAriaProperties {
 	checked?: boolean;
+	id?: string;
```
tomdye: Could this be added to `InputProperties` or other type in the `interfaces` file instead? Otherwise we're back to repeated properties.

src/splitpane/tests/functional/SplitPane.ts
```diff
@@ -7,6 +7,7 @@ import Test from 'intern/lib/Test';
 import * as css from '../../../theme/splitpane/splitPane.m.css';
 
 const DELAY = 300;
+const ERROR_MARGIN = 5;
```
tomdye: Glad this has been moved to a const, but why has it been changed from `1` to `5` within this issue?

src/checkbox/Checkbox.ts
smhigley: I'm good with it in `InputProperties` if you are 👍. I left it out initially because it didn't seem an input-specific thing, but we can say they're being added for input-related reasons.

src/splitpane/tests/functional/SplitPane.ts
smhigley: Just so the functional tests would actually pass (at least more frequently) :P

PR #437, opened 2018-01-25
Refactor usages of `onElementCreated` and `onElementUpdated`

agubler:
**Type:** feature

**Description:**

Refactored usages of `onElementCreated` and `onElementUpdated`in widgets as the hooks are being removed from [`@dojo/widget-core/WidgetBase`](https://github.com/dojo/widget-core/pull/840).

### DIFF
--- a/src/dialog/Dialog.ts
+++ b/src/dialog/Dialog.ts
@@ -3,6 +3,7 @@ import { WidgetBase } from '@dojo/widget-core/WidgetBase';
 import { I18nMixin } from '@dojo/widget-core/mixins/I18n';
 import { ThemedMixin, ThemedProperties, theme } from '@dojo/widget-core/mixins/Themed';
 import { v } from '@dojo/widget-core/d';
+import Focus from '@dojo/widget-core/meta/Focus';
 import uuid from '@dojo/core/uuid';
 import { CustomAriaProperties } from '../common/interfaces';
 import { formatAriaProperties, Keys } from '../common/util';
@@ -81,11 +82,9 @@ export class DialogBase<P extends DialogProperties = DialogProperties> extends T
 		document.addEventListener('keyup', this._onKeyUp);
 	}
 
-	protected onElementCreated(element: HTMLElement, key: string) {
-		if (key === 'main') {
-			const { open = false } = this.properties;
-			open && element.focus();
-		}
+	protected onAttach() {
+		const { open = false } = this.properties;
+		open && this.meta(Focus).set('main');
 	}
 
 	protected onDetach(): void {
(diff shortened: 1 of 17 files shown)


src/dialog/Dialog.ts
```diff
@@ -81,11 +82,9 @@ export class DialogBase<P extends DialogProperties = DialogProperties> extends T
...
 	}
 
-	protected onElementCreated(element: HTMLElement, key: string) {
-		if (key === 'main') {
-			const { open = false } = this.properties;
-			open && element.focus();
-		}
+	protected onAttach() {
+		const { open = false } = this.properties;
+		open && this.meta(Focus).set('main');
```
smhigley: this isn't working, `onAttach` is only called when the widget as a whole is added to the DOM, not when the dialog element is created and attached.

smhigley: I forgot to add functional focus tests when I added focus handling to the dialog, but those would've been useful 😝

src/splitpane/SplitPane.ts
```diff
@@ -39,11 +40,11 @@ const DEFAULT_SIZE = 100;
 
 @theme(css)
 export class SplitPaneBase<P extends SplitPaneProperties = SplitPaneProperties> extends ThemedBase<P, null> {
-	private _divider: HTMLElement;
+	// private _divider: HTMLElement;
 	private _dragging: boolean;
 	private _lastSize?: number;
 	private _position: number;
-	private _root: HTMLElement;
+	// private _root: HTMLElement;
```
smhigley: Can we remove these?

src/dialog/Dialog.ts
agubler: Ha that would have let me know it wasn't working!

src/select/Select.ts
```diff
@@ -159,14 +159,6 @@ export class SelectBase<P extends SelectProperties = SelectProperties> extends T
...
 	}
 
-	protected onElementUpdated(element: HTMLElement, key: string) {
-		if (key === 'root' && this._callListboxFocus) {
-			this._callListboxFocus = false;
-			const listbox = <HTMLElement> element.querySelector('[role="listbox"]');
-			listbox && listbox.focus();
-		}
-	}
-
```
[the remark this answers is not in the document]
agubler: Think we're going to have to raise an issue to address this retrospectively (again no functional test failures)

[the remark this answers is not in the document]
smhigley: It probably needs the same treatment as Combobox/TextInput, where Listbox would need a `focus` property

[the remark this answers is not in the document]
smhigley: I'm hesitant to merge the PR with broken keyboard functionality for Select, though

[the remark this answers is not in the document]
agubler: Should be sending focus in the listbox now.

src/dialog/Dialog.ts
agubler: Have moved this to be called when the `open` property is `true` and the dialog was not open previously (already had this state in the widget)

Issue #458, opened 2018-02-02
Hover effect of enhanced text box does not include before / after addons


tomdye:
**Bug** 

The onHover shadow / highlight on the text input does not extend to the addons. This makes them look like they are not a single input.

![screenshot 2018-02-02 14 54 37](https://user-images.githubusercontent.com/814453/35739330-28e9b66c-0829-11e8-87ad-7d0772a5a7d6.png)

smhigley:
Other libraries that include input addons don't appear to include the addon in the hover/focus styles (e.g. https://getbootstrap.com/docs/4.0/components/input-group/). Might be good to get Scott's opinon.

Either way, I've updated form widgets to always append a `.focused` class to the root node so more advanced focus styles are at least possible, even if we don't decide to implement them.

tomdye:
I agree that bootstrap does not include the add-on in it's focus ring, but it's the drop shadow that I found most troubling. It made it look like it was not part of the same component.

smhigley:
I've actually forgotten what was in the original comp, but either way I was planning to send this (and a couple others ) to Scott for his opinion when he has availability.

PR #469, opened 2018-02-08
Focused class on form widgets

smhigley:
**Type:** bug / feature

**Description:**

Adds a `css.focused` class to the root node of form widgets when the widget contains focus. This makes it possible to make the styling change referenced in #458


tomdye:
@smhigley do you think that `focus` property should be passed to the `label` as well?

smhigley:
@tomdye I would lean towards not providing that to `Label`. Generally the focus outline/highlight should stick pretty close to the actual thing in focus, and I'm having a hard time thinking of a good use case for the label styles needing to change when the input is focused.

smhigley:
I actually think it makes more DOM sense to have the `.focused` class on `.inputWrapper` in most cases, but organizationally we already have `getModifierClasses` and some widgets like `Listbox` lack `inputWrapper`.

tomdye:
@smhigley I understand what you're saying re. label, however, I am thinking about scenarios such as the floating labels for material input labels. Without the ability to set the label as `focused`, I believe you would find it harder to add the appropriate effect via a theme.

screenshot 2018-02-15 10 08 02

PR #475, opened 2018-02-15
Titlepane bugfixes

### LINKED ISSUE #448: Titlepane animates to closed on page load
**Bug** 

When closed titlepanes are placed on a page they start open and animate to closed. This causes an animation on page load and does not provide a good user experience.
They should likely be changed to use the WebAnimation meta as well.

### LINKED ISSUE #449: Closed Titlepane contents shows through the title
**Bug** 

Due to the nature of the `margin-top` animation for hiding a titlepane's contents, it is visible through the titlepane itself. The contents appears to be at a level above that of the title and thus, it is clickable. This can render the title toggle itself to be unusable.

screenshot 2018-02-02 14 36 01

### LINKED ISSUE #450: Closed titlepane contents shows on page resize
**Bug** 

When a page / split pane is resized that contains a closed titlepane, it's contents when being re-flowed can show out form underneath the title.

This is likely due to the the nature of the `margin-top` used rather than `height: 0` to hide the contents.

screenshot 2018-02-02 14 38 20

smhigley:
**Type:** bug

**Description:**
Resolves #448 
Resolves #449 
Resolves #450

### DIFF
--- a/intern.json
+++ b/intern.json
@@ -43,7 +43,11 @@
 
 	"node": {
 		"plugins": [
-			"_build/common/tests/node-css-plugin.js"
+			"_build/common/tests/node-css-plugin.js",
+			{
+				"script": "./_build/common/tests/support/loadJsdom.js",
+				"useLoader": true
+			}
 		]
 	},
 
--- a/src/titlepane/TitlePane.ts
+++ b/src/titlepane/TitlePane.ts
@@ -1,7 +1,7 @@
 import uuid from '@dojo/core/uuid';
 import { DNode } from '@dojo/widget-core/interfaces';
 import { theme, ThemedMixin, ThemedProperties } from '@dojo/widget-core/mixins/Themed';
-import { v } from '@dojo/widget-core/d';
+import { v, w } from '@dojo/widget-core/d';
 import { WidgetBase } from '@dojo/widget-core/WidgetBase';
 
 import * as fixedCss from './styles/titlePane.m.css';
@@ -9,6 +9,7 @@ import * as css from '../theme/titlepane/titlePane.m.css';
 import * as iconCss from '../theme/common/icons.m.css';
 import { Dimensions } from '@dojo/widget-core/meta/Dimensions';
 import { customElement } from '@dojo/widget-core/decorators/customElement';
+import { GlobalEvent } from '../global-event/GlobalEvent';
 
 /**
  * @type TitlePaneProperties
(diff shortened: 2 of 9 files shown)


src/titlepane/TitlePane.ts
```diff
@@ -73,6 +79,30 @@ export class TitlePaneBase<P extends TitlePaneProperties = TitlePaneProperties>
 		}
 	}
 
+	protected animate(effects: any[]) {
+		const { open = true } = this.properties;
+		this.meta(WebAnimation).animate('content', {
```
tomdye: i'd imagine you'll need to add an `onAnimationEnd` callback to set `play` to false btw. Could be wrong but i had issues with this elsewhere when other things in the page re-render.

src/titlepane/TitlePane.ts
```diff
@@ -73,6 +79,30 @@ export class TitlePaneBase<P extends TitlePaneProperties = TitlePaneProperties>
 		}
 	}
 
+	protected animate(effects: any[]) {
```
tomdye: Is this `protected` for the purposes of extensibility?
I have raised a PR to type the `effects` array in the `meta`, https://github.com/dojo/widget-core/pull/865/files#diff-2b4688de89332ae4dfc392a34512c92aR36

src/titlepane/TitlePane.ts
```diff
@@ -73,6 +79,30 @@ export class TitlePaneBase<P extends TitlePaneProperties = TitlePaneProperties>
...
+		const { open = true } = this.properties;
+		this.meta(WebAnimation).animate('content', {
+			id: this._id,
+			effects,
```
tomdye: you should be able to pass `this.getAnimationKeyframes` here and change it to `private`. No need to call it before hand and pass the `effects[]` in.

smhigley: This was so I would only need to call `this.getAnimationKeyframes` once. It's also used to set initial styles here: https://github.com/dojo/widgets/pull/475/files/a0b4c8fca97cc8b9dd4c778e7b4a0b68f1aa25c9#diff-6c2953e8f663b684ca1cfaf618e9b404R185

src/titlepane/TitlePane.ts
smhigley: Yup, since things like duration and easing are no longer easily changed with CSS variables

intern.json
```diff
@@ -43,7 +43,11 @@
 
 	"node": {
 		"plugins": [
-			"_build/common/tests/node-css-plugin.js"
+			"_build/common/tests/node-css-plugin.js",
+			{
+				"script": "./_build/common/tests/support/loadJsdom.js",
```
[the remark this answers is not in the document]
smhigley: Stubbed the meta, but the web animation polyfill in `widget-core` throws an error without jsdom

tomdye:
Raised https://github.com/dojo/widgets/pull/493 to fix conflicts / assist with updates.

Issue #402, opened 2017-12-08
Slider `<output>` receives focus in Edge


mwistrand:
**Bug** 

When navigating with the `tab` key through an application that contains the `Slider` widget on MS Edge, the `<output>` tag receives focus. As far as I can tell, this is unexpected behavior, but it still causes focus to appear to vanish. It would be preferable if this element were explicitly excluded from receiving focus.

Package Version: rc

**Expected behavior:**

The slider's `<output>` cannot receive focus through keyboard navigation.

**Actual behavior:**

The slider's `<output>` can and does receive focus through keyboard navigation.

rorticus:
PR, https://github.com/dojo/widgets/pull/476

agubler:
Fixed by #476

PR #476, opened 2018-02-16
Explicitly disabling focus from output tag for Edge

rorticus:
Massive change to disable focus from `<output>` tag so Edge doesn't jump to it when navigating with Tab.

Issue #402


rorticus:
Oopsie! Moved onto something and didn't notice the CI failures. Added that comment and those tests should be 💯

PR #477, opened 2018-02-16
Add icon widget

### LINKED ISSUE #463: Create an icon widget
**Enhancement** 

We've circled around this one a few times, but when trying to create a Button with an icon it for the theme tester, I have realised that without an icon widget, it is not straight forward for a user to use the dojo 2 icons in their own apps.

```ts
// with an icon widget
w(Button, {}, [
   'search',
   w(Icon, { type: 'search' })
])

// without an icon widget
import * as iconCss from '@dojo/widgets/theme/common/icons.m.css'

@theme(iconCss)
w(Button, {}, [
    'search ',
    v('i', { classes: this.theme([ iconCss.icon, iconCss.searchIcon ])})
])
```

maier49:
**Type:** feature

**Description:**

Resolves #463


src/icon/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Example Usage
+
+*Basic Example*
+```typescript
+import Icon from '@dojo/widgets/icon/Icon';
+import { w } from '@dojo/widget-core/d';
+
+w(Icon, {
+	type: 'downIcon'
```
tomdye: can you add `aria` usage to the example please

src/icon/icon.ts
```diff
@@ -0,0 +1,64 @@
...
+	]
+})
+export class IconBase<P extends IconProperties = IconProperties> extends ThemedBase<P, null> {
+	private _onClick (event: MouseEvent) {
+		event && event.stopPropagation();
+		this.properties.onClick && this.properties.onClick();
+	}
+
+	constructor() {
+		super();
```
tomdye: don't think you need this

src/icon/icon.ts
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	protected render(): DNode {
+		const {
+			aria = {},
+			type = 'downIcon'
```
tomdye: I don't think we should have a default for `type`

maier49: Should the property be required on the props then or just render without an icon type if it's not provided?

src/icon/Icon.ts
```diff
@@ -0,0 +1,59 @@
...
+	],
+	attributes: [ 'type' ],
+	events: [
+		'onClick'
+	]
+})
+export class IconBase<P extends IconProperties = IconProperties> extends ThemedBase<P, null> {
+	private _onClick (event: MouseEvent) {
+		event && event.stopPropagation();
+		this.properties.onClick && this.properties.onClick();
```
smhigley: We shouldn't be providing an `onClick` event, since icons aren't interactive -- they're not buttons, nothing indicates they can be clicked, and they are neither focusable nor keyboard-operable. If someone wants an icon to do something, they need to wrap it in a button or link.

src/icon/Icon.ts
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	protected render(): DNode {
+		const {
+			aria = {},
+			type
+		} = this.properties;
+
+		return v('i', {
+			...formatAriaProperties(aria),
```
smhigley: It's good to include the aria props, but it might be worthwhile to also have an explicit property for hidden descriptive text, maybe like `altText`? If it's provided, we can leave off `role="presentation"` and include a visually hidden span with the text. If it's not provided, we can include `role="presentation"` and `aria-hidden="true"`

src/icon/Icon.ts
tomdye: Good point, missed that :)

src/icon/Icon.ts
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * @type IconProperties
+ *
+ * Properties that can be set on an Icon component
+ *
+ * @property type           Icon type, e.g. downIcon, searchIcon, etc.
+ */
+export interface IconProperties extends ThemedProperties, CustomAriaProperties {
+	type: IconType;
+	altText?: string;
```
smhigley: Yay! Could you just add it to typedoc and the `customElement` decorator?

src/button/Button.ts
```diff
@@ -134,7 +127,7 @@ export class ButtonBase<P extends ButtonProperties = ButtonProperties> extends T
 			'aria-pressed': typeof pressed === 'boolean' ? pressed.toString() : null
 		}, [
 			...this.getContent(),
-			popup ? this.renderPopupIcon() : null
+			popup ? w(Icon, { extraClasses: { root: css.addon }, type: 'downIcon' }) : null
```
tomdye: this won't work as `extraClasses` is not themeable. You'll likely need to wrap it in a `span` or a `div` and pass the `css.addon` class to that instead.

src/button/tests/unit/Button.ts
```diff
@@ -1,12 +1,13 @@
+
```
tomdye: why the new line?

src/button/tests/unit/Button.ts
```diff
@@ -80,11 +81,7 @@ registerSuite('Button', {
 				value: 'value'
 			}, [
 				'foo',
-				v('i', {
-					classes: [ css.addon, iconCss.icon, iconCss.downIcon ],
-					role: 'presentation',
-					'aria-hidden': 'true'
-				})
+				w(Icon, { type: 'downIcon', extraClasses: { root: css.addon } })
```
tomdye: same comment again re. wrapping the icon with something for the sake of `css.addon`

src/icon/Icon.ts
```diff
@@ -0,0 +1,69 @@
...
+			classes: this.theme([ css.icon, css[type] ]),
+			role
+		});
+	}
+
+	protected render(): DNode {
+		const {
+			aria = {
+				hidden: 'true'
+			},
```
smhigley: I don't think we want to default to `aria-hidden="true"`, because that will hide the alt text if provided. It should probably only be present if `altText` is missing.

src/icon/README.md
```diff
@@ -0,0 +1,62 @@
...
+*Basic Example*
+```typescript
+import Icon from '@dojo/widgets/icon/Icon';
+import { w } from '@dojo/widget-core/d';
+
+w(Icon, { type: 'downIcon', });
+```
+
+By default the icon will be rendered with `role="presentation"` and
+`aria-hidden="true"`. Custom aria attributes can be specified to
```
smhigley: Really quick word update to something like "The icon will be rendered with `role="presentation"` and
 +`aria-hidden="true"` if `altText` is not provided"

src/icon/Icon.ts
```diff
@@ -0,0 +1,69 @@
...
+	properties: [
+		'theme',
+		'aria',
+		'extraClasses'
+	],
+	attributes: [ 'type' ]
+})
+export class IconBase<P extends IconProperties = IconProperties> extends ThemedBase<P, null> {
+	protected renderAltText(altText?: string) {
+		return altText && v('span', { classes: [ baseCss.visuallyHidden ] }, [ altText ]);
```
tomdye: could the `altText` not just be a `title` attribute on the `i` tag? @smhigley ??

src/icon/Icon.ts
```diff
@@ -0,0 +1,69 @@
...
+		const icon = this.renderIcon(type, altTextNode ? undefined : 'presentation', aria);
+
+		return v('span', { classes: this.theme(css.root) }, [
+			icon,
+			altTextNode
+		]);
+	}
+}
+
+export default class Icon extends IconBase<IconProperties> {}
```
tomdye: just `export default class Icon`

maier49: Not sure I follow. Do you want me to get rid of the base class?

src/icon/Icon.ts
maier49: Now it's always setting `aria-hidden="true"` on the icon itself (unless explicitly overridden), never setting `role="presentation"` on anything, and includes a visually hidden with `altText` if provided.

src/icon/Icon.ts
maier49: Added

src/icon/Icon.ts
maier49: As we talked about, the label is a sibling of the icon so we can safely always hide the icon.

src/icon/README.md
maier49: Updated this, but with slightly different wording since we'll always have `aria-hidden="true"` regardless of whether `altText` is provided.

src/button/Button.ts
maier49: Wrapped them in a `span`

src/button/tests/unit/Button.ts
maier49: Formatting failure. Removed these from all the files they were in.

src/icon/Icon.ts
maier49: I think having a separate `span` is preferred.

src/icon/Icon.ts
smhigley: @tomdye I'm not entirely sure if this is what you meant, but this pattern was adopted because of typing weirdness around lack of type narrowing when providing the generic for properties.

src/icon/Icon.ts
```diff
@@ -0,0 +1,69 @@
...
+import { customElement } from '@dojo/widget-core/decorators/customElement';
+
+export type IconType = keyof typeof css;
+
+/**
+ * @type IconProperties
+ *
+ * Properties that can be set on an Icon component
+ *
+ * @property type           Icon type, e.g. downIcon, searchIcon, etc.
```
smhigley: I promise this is the very last thing: could you add `altText` after this?

smhigley: nevermind! My view was out of date :)

src/icon/Icon.ts
tomdye: Wasn't aware of that pattern being used, I assumed that the class being defined up above with the `extends` and `IconBase` should allow this to simply be `export default class Icon` ?

src/calendar/Calendar.ts
```diff
@@ -375,13 +374,11 @@ export class CalendarBase<P extends CalendarProperties = CalendarProperties> ext
 	}
 
 	protected renderPagingButtonContent(type: Paging, labels: CalendarMessages): DNode[] {
-		const iconClass = type === Paging.next ? iconCss.rightIcon : iconCss.leftIcon;
+		const iconType = type === Paging.next ? 'rightIcon' : 'leftIcon';
```
tomdye: could this be `IconType.rightIcon` ? Might be better using the exported type. There's a few instances like this within this PR.

tomdye:
@maier49 this will require a change in `@dojo/themes` too as the icons filename has changed.

tomdye:
@maier49 I raised https://github.com/dojo/themes/pull/9 to go with this change.

src/calendar/Calendar.ts
maier49: Right now `IconType` is just a type, so there is no `IconType.rightIcon`.

tomdye:
@maier49 looks like this is failing CI right now

src/icon/README.md
```diff
@@ -0,0 +1,76 @@
+# @dojo/widgets/icon/Icon widget
+
+Dojo 2's `Icon` widget renders an icon.
+
+
+## Example Usage
+
+*Basic Example*
+```typescript
+import Icon from '@dojo/widgets/icon/index';
```
tomdye: When people use our widgets in their own apps they don't need to use `/index`, this is just for our AMD test loaders sadly. I'm happy to let this go as it though and will update all readmes in a follow up PR

src/icon/index.ts
```diff
@@ -0,0 +1,69 @@
...
+				hidden: 'true'
+			},
+			type,
+			altText
+		} = this.properties;
+
+		const altTextNode = this.renderAltText(altText);
+		const icon = this.renderIcon(type, aria);
+
+		return v('span', { classes: this.theme(css.root) }, [
```
tomdye: Do you think this really needs a `renderIcon` method? I can't see anyone ever extending it, it's simply an icon.
As for `renderAltText`, I think that it should only be called if there is `altText`.

```ts
return v('span', { classes: this.theme(css.root) }, [
   v('i', {
      ...formatAriaProperties(aria),
      classes: this.theme([ css.icon, css[type] ])
   }),
   altText ? this.renderAltText(altText) : null
]);
```

src/theme/icon.m.css
```diff
@@ -1,13 +1,15 @@
 @font-face {
 	font-family: 'dojo2';
 	src:
-	url('./fonts/dojo2.ttf?fkcsox') format('truetype'),
-	url('./fonts/dojo2.woff?fkcsox') format('woff'),
-	url('./fonts/dojo2.svg?fkcsox#dojo2') format('svg');
+	url('../common/fonts/dojo2.ttf?fkcsox') format('truetype'),
```
tomdye: I believe this path was correct before, the font files are in `./fonts`

tomdye:
Raised https://github.com/dojo/widgets/pull/494 to replace this PR

PR #480, opened 2018-02-16
Extract Toggle from Checkbox

### LINKED ISSUE #455: Checkbox / Toggle widgets should be split
**Enhancement** 

The checkbox and toggle widgets should likely be split into their own widgets.
The css file for these would be made much simpler and it will be clearer to users how to use and style them.

jason0x43:
Toggle is a separate component that extends Checkbox.

references #455

**Type:** feature

**Description:**

This PR splits Toggle out of Checkbox.

Resolves #455

### DIFF
--- a/src/checkbox/index.ts
+++ b/src/checkbox/index.ts
@@ -16,62 +16,54 @@ import { customElement } from '@dojo/widget-core/decorators/customElement';
  * Properties that can be set on a Checkbox component
  *
  * @property checked        Checked/unchecked property of the radio
- * @property mode           The type of user interface to show for this Checkbox
- * @property offLabel       Label to show in the "off" positin of a toggle
- * @property onLabel        Label to show in the "on" positin of a toggle
  * @property value           The current value
  */
 export interface CheckboxProperties extends ThemedProperties, InputProperties, LabeledProperties, KeyEventProperties, PointerEventProperties, CustomAriaProperties, CheckboxRadioEventProperties {
 	checked?: boolean;
-	mode?: Mode;
-	offLabel?: DNode;
-	onLabel?: DNode;
 	value?: string;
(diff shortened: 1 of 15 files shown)


src/checkbox/Checkbox.ts
```diff
@@ -98,29 +82,7 @@ export class CheckboxBase<P extends CheckboxProperties = CheckboxProperties> ext
 	}
 
 	protected renderToggle(): DNode[] {
```
tomdye: `renderToggle` can be removed from here

tomdye:
Also, don't forget to raise a change in `@dojo/themes` when this is complete

src/checkbox/Checkbox.ts
jason0x43: It's necessary so it can be called in `render` and overridden in Toggle (which keeps Toggle from having to duplicate all the render code).

jason0x43:
I reworked this a bit to better share functionality between Checkbox and Toggle. Almost everything is shared; the only thing added by toggle is two element attributes, `onLabel` and `offLabel`, and an override of `renderInput`.

Speaking of `renderInput`, I updated Checkbox to follow TextInput's model. The input element is now rendered in `renderInput`, which is called by `render`. Toggle overrides `renderInput` to add its own elements, then concats the output of Checkbox's `renderInput`.

I'm working on an update for @dojo/themes.

jason0x43:
Hmmm...I'm not sure how theme inheritance works, but I may need to re-work it a bit.

tomdye:
Closing this as discussed with @jason0x43 as it has fallen stale due to RC changes and have raised issues around theme inheritance. @jason0x43 are you able to raise issues for what you discovered?

PR #481, opened 2018-02-16
Fix: Toolbar should not wrap content

### LINKED ISSUE #474: Toolbar breaks scroll on History API
**Bug** 

Toolbar has internal scroll, so scroll position isn't maintained when navigating out and back. We should remove the wrapping behavior and have it simply be an element at the top of the page.

mwistrand:
**Type:** bug

**Description:**

Resolves #474.

Updates the Toolbar widget to sit alongside content widgets instead of wrapping them, which allows the current scroll position to be maintained when navigating away from the screen and then returning.

Also removes the Toolbar's content `div` and associated classes; places the title, actions, and button directly beneath the root node; and removes the toolbar node wrapper and associated classes. 

Finally, disables setting the position to `Position.bottom` when the `fixed` property is falsy, as
this functionality is better served by a footer component.


dylans:
Also for some reason the CI on this is using TypeScript 2.2.1 and failing. And a test is failing on iOS 9 Safari.

umaar:
Hey @smhigley I first rebased with master to make sure the branch is up to date and in turn resolved a few conflicts, then applied the changes we spoke about! Do you mind checking? Have also checked the widget visually and can confirm if we add items to a page, the toolbar will not overlap them.

tomdye:
taking a look @umaar thanks for picking this up

tomdye:
closed via: https://github.com/dojo/widgets/pull/504

PR #482, opened 2018-02-19
Remove dom event from widget callbacks

### LINKED ISSUE #302: Standardize widget events
**Enhancement** 

The approach to widget events is fragmented across existing widgets. Basic form widgets have handlers for all interaction events (mouse, touch, and keyboard), and pass back the dom event object as the argument. Most other widgets only include widget-specific events (e.g. `onOpen` for Dialog), and pass some part of requested state back as the argument (e.g. `year` for `onYearChange` in Calendar).

Widgets could be improved by standardizing event handlers in two ways:
- More logical approach to which events are exposed
- Consistent approach to which arguments are passed to the event handler

tomdye:
**Type:** feature

**Description:**

Removes the dom events from widget callbacks.
Widgets not callback with data more appropriate to the widget type. ie. 

- inputs return `value:string`
- checkbox returns `value: string` and `checked: boolean`.

Resolves #302

### DIFF
--- a/src/checkbox/Checkbox.ts
+++ b/src/checkbox/Checkbox.ts
@@ -61,20 +61,20 @@ export const ThemedBase = ThemedMixin(WidgetBase);
 })
 export class CheckboxBase<P extends CheckboxProperties = CheckboxProperties> extends ThemedBase<P, null> {
 	private _onBlur (event: FocusEvent) {
-		const checkbox = (event.target as HTMLInputElement);
+		const checkbox = event.target as HTMLInputElement;
 		this.properties.onBlur && this.properties.onBlur(checkbox.value, checkbox.checked);
 	}
 	private _onChange (event: Event) {
-		const checkbox = (event.target as HTMLInputElement);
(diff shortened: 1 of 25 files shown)


src/checkbox/Checkbox.ts
```diff
@@ -62,38 +61,41 @@ export const ThemedBase = ThemedMixin(WidgetBase);
 })
 export class CheckboxBase<P extends CheckboxProperties = CheckboxProperties> extends ThemedBase<P, null> {
 	private _onBlur (event: FocusEvent) {
-		this.properties.onBlur && this.properties.onBlur(event);
+		const checkbox = (event.target as HTMLInputElement);
+		this.properties.onBlur && this.properties.onBlur(checkbox.value, checkbox.checked);
 	}
 	private _onChange (event: Event) {
-		this.properties.onChange && this.properties.onChange(event);
+		const checkbox = (event.target as HTMLInputElement);
```
agubler: Don’t need the parentheses here

tomdye: good spot

PR #495, opened 2018-02-28
Build and distribute all widgets as custom elements 

agubler:
**Type:** feature

**Description:**

Adds mechanism to build and publish all @dojo widgets as custom elements. Includes a couple of changes within widgets required to be usable as a custom element:

 * `TabController` now renders the content for all panes passing a flag to indicate if it is the content that is "shown" based on the tab index, this is because otherwise it custom elements cannot deal with a custom element child correctly.
 * `SplitPane` now uses children over properties for the `leading` and `trailing` panels - this is because custom elements cannot deal with widgets as properties currently.


src/enhanced-text-input/index.ts
```diff
@@ -21,7 +21,7 @@ export interface EnhancedTextInputProperties extends TextInputProperties {
 		'addonAfter',
 		'addonBefore',
 		'aria',
-		'focus',
+		'shouldFocus',
```
agubler: `focus` does not blend well when used with custom elements so renamed to `shouldFocus`

PR #504, opened 2018-03-02
Fix Toolbar fixed height, dont wrap content, hide button when no actions

### LINKED ISSUE #474: Toolbar breaks scroll on History API

tomdye:
**Type:** bug & feature

**Description:**

Replaces: https://github.com/dojo/widgets/pull/481

- Stops toolbar wrapping content
- Hides button when no actions
- Adds sized container styles around placeholder to push page content down

Resolves #474


smhigley:
Should we make the actions just normal children of `Toolbar` now that it doesn't need to wrap the rest of page content? We don't really need to individually wrap them, I think styling those can be up to the user/parent component.

PR #505, opened 2018-03-02
Add split-pane collapse and flip row / column

tomdye:
**Type:** feature

**Description:**

- For mobile / small screens this PR stacks the split-pane upon resize beyond a threshold using the same approach as `toolbar` when the direction is set to columns.
- This PR also swaps row/column usage as they were the wrong way around.
- Sets a min size for a the headless browser

Related to: https://github.com/dojo/examples/issues/293


src/split-pane/index.ts
```diff
@@ -111,28 +114,59 @@ export class SplitPaneBase<P extends SplitPaneProperties = SplitPaneProperties>
...
+		styles[direction === Direction.column ? 'width' : 'height'] = `${computedSize}px`;
+
 		return styles;
 	}
 
+	protected onAttach() {
+		this._collapseIfNecessary();
+	}
+
+	private _collapseIfNecessary = () => {
```
agubler: @tomdye Do you think we should react to a change to the `collapseWidth` via properties and set the `this._collapsed` flag and call `onCollapse` appropriately? Otherwise if a different `collapseWidth` is passed the component won't react until it is is resized?

agubler: Guess the same applies if the `Direction` is change also...

PR #509, opened 2018-03-05
Fix toolbar class usage

tomdye:
**Type:** bug

**Description:**

Recent toolbar changes removed the `toolbar` class and added a root class below the top level. This broke themes and our class name guidelines. This pr corrects those mistakes.


tomdye:
Tested and confirmed working in the showcase with current theme implementation

PR #526, opened 2018-03-22
TS2.7 forwards compatibility

agubler:
**Type:** feature

**Description:**

Start to prepare for compatibility with TS2.7 (not quite there as there is an outstanding issue with defaulting optional properties) and upgrade with dojo RC2 changes.

Requires release of i18n, widget-core, test-extras

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,9 +25,9 @@
   "peerDependencies": {
     "@dojo/core": "~0.5.0",
     "@dojo/has": "~0.1.1",
-    "@dojo/i18n": "~0.5.1",
-    "@dojo/shim": "~0.2.6",
-    "@dojo/widget-core": "./../widget-core/dist/dojo-widget-core-0.9.7-rc.1.tgz"
+    "@dojo/i18n": "~0.6.0",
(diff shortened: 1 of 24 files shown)


package.json
```diff
@@ -23,11 +23,11 @@
     "dist-custom-elements": "npm run build-custom-elements && shx rm -rf dist/all/dist && shx cp -r output/dist dist/all/"
   },
   "peerDependencies": {
-    "@dojo/core": "~0.4.0",
+    "@dojo/core": "~0.5.0",
     "@dojo/has": "~0.1.1",
     "@dojo/i18n": "~0.5.1",
     "@dojo/shim": "~0.2.6",
-    "@dojo/widget-core": "~0.9.5"
+    "@dojo/widget-core": "./../widget-core/dist/dojo-widget-core-0.9.7-rc.1.tgz"
```
dylans: Just remember to update this with the real release version before merging this.

agubler: Of course, I do put it in the description that it requires releases and wouldn't merge if it doesn't pass CI 😄

PR #530, opened 2018-03-27
Rename widget id properties

### LINKED ISSUE #529: Widget `id` properties cause issues for custom element attributes
**Bug**

`id` is a valid html attribute on custom elements, so it fails as a widget property name.

smhigley:
**Type:** bug

**Description:**

Resolves #529 

Renames `id` properties, primarily to `inputId` and `rootId`

### DIFF
--- a/src/enhanced-text-input/tests/unit/EnhancedTextInput.ts
+++ b/src/enhanced-text-input/tests/unit/EnhancedTextInput.ts
@@ -143,6 +143,7 @@ registerSuite('EnhancedTextInput', {
 						controls: 'foo',
 						describedBy: 'bar'
 					},
+					inputId: 'textInputId',
 					maxLength: 50,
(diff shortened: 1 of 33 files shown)


src/enhanced-text-input/tests/unit/EnhancedTextInput.ts
```diff
@@ -143,6 +143,7 @@ registerSuite('EnhancedTextInput', {
 						controls: 'foo',
 						describedBy: 'bar'
 					},
+					widgetId: 'textwidgetId',
```
dylans: Should this be textWidgetId?

smhigley: oh, hah, I think this was a find and replace relic 😆

PR #534, opened 2018-04-09
Add API docs

maier49:
**Type:** feature
 
Adds API docs and docviewer config for dojo/meta#239


agubler:
Quick question about this? Did we decide to put back API docs for widgets then? I thought it was one of the packages that didn't make sense to do the standard api docs?

dylans:
I think we decided not to include this for @dojo/widgets. So we should probably close this @maier49?

PR #546, opened 2018-04-27
Readme updates

### LINKED ISSUE #471: Readme updates
**Enhancement** 

## Main readme:
- the wording for the difference between `onRequest*` and `on*` event handlers is a little confusing
- Icons should probably be updated to say we're planning on creating an icon widget
- In "Extending Widgets" we can add more words around what extension points most widgets have (e.g. `getRootClasses`), and provide an example
- A11y section could mention that a number of widgets allow `aria: {}` as a way to add custom ARIA semantics to the applicable node (this could also go within individual widget readmes)
- Could use section on focus

## Individual widget readmes:
- Button readme has root classes listed, but misses themeable addon classes
- Some individual widget readmes still use placekittens 😹 
- ComboBox references `inputProperties.describedBy`, which has been removed in favor of `inputProperties.aria`
- ComboBox mentions `CustomResultItem`, which no longer exists
- Label widget -- would be a good addition to mention that it's already included in all form widgets and controlled by `label`, `labelAfter`, and `labelHidden`
- Label needs classes section, and examples are wrong (no `label` property within the Label widget)
- Listbox needs a readme
- Form widget readme text still says "optionally wrapped in a label" (e.g. on the Radio widget readme)
- Select readme is empty
- Textarea label example and `describedBy` are outdated
- TextInput uses the wrong way to add aria props
- Timepicker should say that it wraps ComboBox
- AccordionPane says it provides "responsive styling" over multiple TitlePanes, which is confusing
- SplitPane readme is empty
- Dialog: not sure if we should say "will respond to different screen sizes responsively" as it's up to the theme styles, not the widget functionality
- Tooltip should have a11y section

Some things will change when dealing with open issues. Typing of input events in readme examples, the icon widget,

smhigley:
**Type:** feature

**Description:**

Resolves #471


src/listbox/README.md
```diff
@@ -0,0 +1,117 @@
+# @dojo/widgets/listbox widget
+
+Dojo 2's `Listbox` provides a base widget that can be used as an options menu within a Select or ComboBox component, or as a standalone multi-select component
```
dylans: Select or ComboBox -> `Select` or `ComboBox`

src/listbox/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Features
+
+- Compatible with any underlying data provider, data format, or store
+- Allows customization of option state and vdom through reactive patterns
+- Handles all keyboard interactions if given focus
+- Can be controlled through `activeIndex` and `visualFocus`, e.g. in a ComboBox where focus stays on the text input
```
dylans: ComboxBox -> `ComboBox`

src/listbox/README.md
```diff
@@ -0,0 +1,117 @@
...
+		optionData[index].selected = !optionData[index].selected;
+		optionData = [ ...optionData ];
+		this.invalidate();
+	}
+});
+```
+
+## Theming
+
+The following CSS classes are used to style the `Select` widget and should be provided by custom themes:
```
dylans: Did you mean `Listbox` instead of `Select`?

PR #560, opened 2018-06-15
showOutput code and tests

### LINKED ISSUE #559: Slider has no option to hide output
**Bug / Enhancement** 

`Slider` should allow you to decide whether you want to show or hide output, similar to `Progress`.

Package Version: 2.0.0

**Sample**

```typescript
w(Slider, {
   /* . . . */
   showOutput: false // same property name as in Progress bar
});
```

**Expected behavior:**

The `output` node is not rendered

schontz:
**Type:** bug / feature

* [x] There is a related issue #559 
* [x] Unit or Functional tests are included in the PR (I did not include a functional test because I could not find a way to test for the node NOT existing with Intern.)

**Description:**

Adds `showOutput` option to `Slider` so you can remove the output node from the rendering. Same property as `Progress`.

Resolves #559

### DIFF
--- a/src/slider/example/index.ts
+++ b/src/slider/example/index.ts
@@ -8,6 +8,7 @@ export class App extends WidgetBase<WidgetProperties> {
 	private _tribbleValue: number | undefined;
 	private _verticalValue: number | undefined;
 	private _verticalInvalid: boolean | undefined;
+	private _outputlessValue: number | undefined;
 
 	onTribbleInput(value: number) {
 		this._tribbleValue = value;
@@ -17,6 +18,11 @@ export class App extends WidgetBase<WidgetProperties> {
 	onVerticalInput(value: number) {
 		this._verticalValue = value;
 		this._verticalInvalid = value > 50;
+		this.invalidate();
+	}
+
+	onOutputlessInput(value: number) {
+		this._outputlessValue = value;
 		this.invalidate();
 	}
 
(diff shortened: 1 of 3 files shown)


schontz:
I'm not sure what the errors in Chrome have to do with my code. It looks like unrelated.

schontz:
Would it make sense to make `output: boolean | (value: number) => DNode` (similar to things like `focus`)? Then if you simply do not want any output you can set `output=false` and be done with it. 

Also, the reason I added `showOutput` was to follow the design pattern of `Progress`. Does it make sense to unify this property regardless of the route we go? (Also, `Progress` uses a `span` for output instead of `output`. Is that by design?)

smhigley:
You have a good point about standardizing `output` properties between `Slider` and `Progress`. Let's leave `showOutput: boolean` for now since that's what `Progress` already uses and it involves the least change to existing APIs.

Regarding `<span>` vs. `<output>`, that is actually on purpose. since `<output>` semantically indicates the result of a calculation based on user input and belongs inside a form. `Slider` uses it since it's a form element, whereas `Progress` is not. The `progressbar` role also has its own semantics to indicate current value (i.e. `aria-valuenow`), so something with `output`-like semantics would be a duplicate of that. Good question though!

smhigley:
The tests pass now, so I'm going to merge it. Thanks for adding this!

PR #563, opened 2018-07-12
Change comobobox getResultLabel type to match select and listbox

smhigley:
**Type:** feature

**Description:**

Both `Select` and `Listbox` allow a `DNode` return type for `getOptionLabel`. It makes sense to update `ComboBox` to allow the same.

### DIFF
--- a/src/combobox/index.ts
+++ b/src/combobox/index.ts
@@ -28,8 +28,9 @@ import { customElement } from '@dojo/widget-core/decorators/customElement';
  * @property clearable          Determines whether the input should be able to be cleared
  * @property disabled           Prevents user interaction and styles content accordingly
  * @property getResultLabel     Can be used to get the text label of a result based on the underlying result object
- * @property getResultSelected  Can be used to highlight the selected result. Defaults to checking the result label.
- * @property widgetId            Optional id string for the combobox, set on the text input
+ * @property getResultSelected  Can be used to highlight the selected result. Defaults to checking the result label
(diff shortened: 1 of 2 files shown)


src/combobox/index.ts
```diff
@@ -48,14 +48,14 @@ import { customElement } from '@dojo/widget-core/decorators/customElement';
...
-	getResultLabel?(result: any): string;
+	getResultLabel?(result: any): DNode;
 	getResultSelected?(result: any): boolean;
 	widgetId?: string;
 	inputProperties?: TextInputProperties;
 	invalid?: boolean;
 	isResultDisabled?(result: any): boolean;
 	onBlur?(value: string, key?: string | number): void;
-	onChange?(value: string, key?: string | number): void;
+	onChange?(value: DNode, key?: string | number): void;
```
tomdye: is this right? all the other `on{thing}` callbacks return the value as a string

smhigley: @tomdye the other `on{thing}` callbacks use the `value` prop, while `onChange` is fired when changing options. It would probably originally have been better to do `onChange?(result: any, index: number, key?: string | number)` like `Listbox` does, but that would be a breaking change 😝

smhigley: The other, possibly easier, option is to add `getResultValue` and return that in `onChange`. That might be better anyway, since `onChange` now is used for both typing in the `TextInput` and selecting a result.

PR #582, opened 2018-08-31
Add onResultSelect Event to ComboBox

### LINKED ISSUE #485: Add onSelect/onClear callbacks to ComboBox
**Enhancement**

Please add onResultSelect callback to the ComboBox widget, and pass the raw value to it (similar to the getResultLabel). This makes easier to distinguish from filter text and selected value, especially when value  isn't raw text but an object.

Please add onClear callback/event so there's a distinction from input and clear (not that important).

pottedmeat:
**Type:** feature

**Description:**

Resolves #485 

Adds the onResultSelect event, called when enter/select is pressed and via the nested Listbox `onOptionSelect` event.

### DIFF
--- a/src/combobox/index.ts
+++ b/src/combobox/index.ts
@@ -40,7 +40,7 @@ import { customElement } from '@dojo/framework/widget-core/decorators/customElem
  * @property onFocus            Called when the input is focused
  * @property onMenuChange       Called when menu visibility changes
  * @property onRequestResults   Called when results are shown; should be used to set `results`
- * @property onResultSelect		Called when result is selected
+ * @property onResultSelect     Called when result is selected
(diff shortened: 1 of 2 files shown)


src/combobox/index.ts
```diff
@@ -40,6 +40,7 @@ import { customElement } from '@dojo/framework/widget-core/decorators/customElem
  * @property onFocus            Called when the input is focused
  * @property onMenuChange       Called when menu visibility changes
  * @property onRequestResults   Called when results are shown; should be used to set `results`
+ * @property onResultSelect		Called when result is selected
```
smhigley: I'm such a nitpicker, but could you format this with spaces so it lines up? 🙇‍♀️

Issue #586, opened 2018-09-05
Accessible cell editing for grid widget


smhigley:
**Enhancement**

It should be possible to initiate editing a cell with the keyboard, and to communicate that a cell is editable to assistive tech users (also possibly to all users).

Changes required:
- When a cell is not in editing mode, there should be an "Edit" button. This should ideally be visible to all users, but perhaps with an option to show it only on focus. One potential issue with only showing the button to keyboard users is that some people may have issues with time-dependent interactions like double clicking
- The edit button needs an accessible label that is better than "edit." It should likely be "Edit {{ current value of cell }}"
- Focus should be managed into and out of the input when switching to and from edit mode
- The input should be a child of the cell
- The input needs a label (also explore the possibility of a custom input type, e.g. number or email)

dylans:
I would probably advise against an edit button visible for all users... data grids with editable fields tend to be the type of thing where users do a lot of rapid editing via keyboard shortcuts. With dgrid1, the modes to enable editing were something like always an editor, turn into an editor on focus, or on click, or double click into a field. Saving of fields could be set to be done automatically, or with an explicit button per grid, or per editor (save or cancel). The most common use cases in my experiences were to either use auto save, or to think of the entire grid as a form to be submitted as a single transaction.

dylans:
In some grids editable fields are always editable. Other grids implement editing on one row at a time. Others just switch to editable controls on focus or click, eg you select a cell and it switches from text to an input field. In grids that are editable the average user is more often a power user.

PR #588, opened 2018-09-10
Increase keyboard support for cell editing

smhigley:
**Type:** feature

**Description:**

Addresses #586. Adds an edit button that is visible on focus, and handles focus between the input and button when switching into and out of edit mode.

This is primarily a quick fix, and ideally should be updated so that interactive grids use the ARIA grid spec for cell navigation and interaction.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@dojo/widgets",
-  "version": "3.0.1-rc.1",
+  "version": "3.0.1-pre",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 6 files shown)


agubler:
@smhigley could you possibly put some images showing the proposed changes?

smhigley:
@agubler here's an animated gif of keyboard interaction with the edit button and escaping out of the input. When I go through and separate fixed vs. theme styles, I plan to make the base styles ever so slightly nicer (e.g. position the edit button to the right of the cell in the theme css).
grid-interaction

Issue #584, opened 2018-09-05
Add grid theme and simplify internal grid styles


smhigley:
**Enhancement** 

This task includes:
- Split existing styles into structural/fixed classes and themeable classes
- Simplify styles remaining in the `@dojo/widgets/grid` themeable classes
- Removed hard-coded height values used to calculate body dimensions
- Create a theme for grid in `@dojo/themes` (if done in tandem with this update, some of the removed styles could potentially be helpful in writing the theme)

PR #594, opened 2018-09-20
Grid height property

smhigley:
**Type:** bug

**Description:**

Addresses #584, removes hard-coded height values for header and footer in grid row calculations in favor of an explicit height property.


smhigley:
Ooo, good call!

PR #595, opened 2018-09-20
Accessible grid sorting and filtering

### LINKED ISSUE #587: Accessible grid filtering and sorting
**Enhancement** 

Filtering:
- Inputs need a label
- Live region should announce the result of filtering (i.e. how many rows are updated as a result)
- Explore filtering on `input` vs. `submit`. If the former, live region updates should likely only happen on blur
- grid body should have `aria-busy` set when loading (this is also true in general, not only for filtering)

Sorting:
- sorter should be triggered by a button, with a label more specific than "sort." Accessible text should include both the column it will sort and the order (i.e. ascending or descending)

smhigley:
**Type:** bug / feature

**Description:**

Resolves #587 

Things to follow up on:
- is `search` the correct input type for a filtering input?
- the filter input's label text is included in the column header text read for each cell in its column, and does not seem to be overridable by `aria-label` or `aria-labelledby` on the column header element.

### DIFF
--- a/src/grid/tests/unit/widgets/Header.ts
+++ b/src/grid/tests/unit/widgets/Header.ts
@@ -1,8 +1,11 @@
 const { describe, it } = intern.getInterface('bdd');
+const { assert } = intern.getPlugin('chai');
 
 import { stub } from 'sinon';
 import harness from '@dojo/framework/testing/harness';
 import { v, w } from '@dojo/framework/widget-core/d';
+import TextInput from '../../../../text-input/index';
+import Icon from '../../../../icon/index';
 
 import * as css from '../../../widgets/styles/Header.m.css';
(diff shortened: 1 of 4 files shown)


src/grid/tests/unit/widgets/Header.ts
```diff
@@ -48,7 +51,7 @@ describe('Header', () => {
 		);
 		h.expect(() =>
 			v('div', { scrollLeft: 0, classes: [css.root, null], row: 'rowgroup' }, [
-				v('div', { classes: css.row, role: 'role' }, [
+				v('div', { classes: css.row, role: 'row' }, [
```
agubler: Ooops that was my mistake 💃

PR #600, opened 2018-10-01
Use TextInput and Button in Grid

smhigley:
**Type:** feature

**Description:**

- Updates the grid cell to use TextInput and Button widgets
- Improves focus handling for cell editing
- Improves focus handling and focusability for TextInput and Button widgets
- updates grid filtering to be case-insensitive

### DIFF
--- a/src/button/index.ts
+++ b/src/button/index.ts
@@ -143,7 +143,7 @@ export class ButtonBase<P extends ButtonProperties = ButtonProperties> extends T
 			classes: this.theme([ css.root, ...this.getModifierClasses() ]),
 			disabled,
 			id: widgetId,
-			focus: this.shouldFocus(),
+			focus: this.shouldFocus,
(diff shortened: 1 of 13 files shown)


src/button/index.ts
```diff
@@ -35,7 +36,7 @@ export interface ButtonProperties extends ThemedProperties, InputEventProperties
 	onClick?(): void;
 }
 
-export const ThemedBase = ThemedMixin(WidgetBase);
+export const ThemedBase = ThemedMixin(FocusMixin(WidgetBase));
```
agubler: Not really `ThemedBase` anymore 😃

agubler:
You got a failing test.

src/button/index.ts
smhigley: Hah, I never even noticed it was called that 😂

PR #601, opened 2018-10-02
Add pattern and autocomplete properties

### LINKED ISSUE #553: TextInput is missing the pattern property
**Bug** 

TextInput has no [`pattern` property](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-pattern) which allows regex validation.
Also a better default for `type="email"` could be nice.
Currently e.g. "example@example" is allowed.

Using

```TypeScript
// please upvote https://github.com/Microsoft/TypeScript/issues/17592
export enum TextInputPatternType {
  'text' = 'text',
  'email' = 'email',
  'search' = 'search',
  'tel' = 'tel',
  'url' = 'url'
}
export enum TextInputType {
  'text' = 'text',
  'email' = 'email',
  'number' = 'number',
  'password' = 'password',
  'search' = 'search',
  'tel' = 'tel',
  'url' = 'url'
}

// do NOT allow e.g. example@example, could be even better but would be longer
export const emailRegexStr = [
  "[a-zA-Z0-9_]+(?:\\.[A-Za-z0-9!#$%&'*+/=?^_`{|}~-]+)",
  "*@(?!([a-zA-Z0-9]*\\.[a-zA-Z0-9]*\.[a-zA-Z0-9]*\\.))",
  "(?:[A-Za-z0-9](?:[a-zA-Z0-9-]*[A-Za-z0-9])?\\.)",
  "+[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?"
].join('');

// IN THE PROPERTIES :
pattern?: string | RegExp;
// <--

// AND IN renderInput
let { pattern } = this.properties;
if (pattern instanceof RegExp) {
  // allowing RegExp
  pattern = pattern.source;
}
if (!(type in TextInputPatternType)) {
  // would be ignored anyway
  pattern = void 0;
} else if (type === 'email' && typeof pattern !== 'string') {
  // better email default
  pattern = emailRegexStr;
}
```

bryanforbes:
**Type:** feature

**Description:**

Adds `pattern` and `autocomplete` properties and passes them through to the `<input>` node

Resolves #553

### DIFF
--- a/src/text-input/index.ts
+++ b/src/text-input/index.ts
@@ -33,6 +33,8 @@ export interface TextInputProperties extends ThemedProperties, InputProperties, 
 	minLength?: number | string;
 	placeholder?: string;
 	value?: string;
+	pattern?: string | RegExp;
+	autocomplete?: string;
 	onClick?(value: string): void;
 }
 
@@ -51,7 +53,7 @@ export const ThemedBase = ThemedMixin(FocusMixin(WidgetBase));
 		'labelAfter',
 		'labelHidden'
 	],
-	attributes: [ 'widgetId', 'label', 'placeholder', 'controls', 'type', 'minLength', 'maxLength', 'value', 'name' ],
+	attributes: [ 'widgetId', 'label', 'placeholder', 'controls', 'type', 'minLength', 'maxLength', 'value', 'name', 'pattern', 'autocomplete' ],
 	events: [
 		'onBlur',
 		'onChange',
@@ -159,12 +161,15 @@ export class TextInputBase<P extends TextInputProperties = TextInputProperties> 
 			readOnly,
 			required,
 			type = 'text',
-			value
+			value,
+			pattern,
+			autocomplete
 		} = this.properties;
 
 		return v('input', {
 			...formatAriaProperties(aria),
 			'aria-invalid': invalid ? 'true' : null,
+			autocomplete,
 			classes: this.theme(css.input),
 			disabled,
 			id: widgetId,
@@ -173,6 +178,7 @@ export class TextInputBase<P extends TextInputProperties = TextInputProperties> 
 			maxlength: maxLength ? `${maxLength}` : null,
 			minlength: minLength ? `${minLength}` : null,
 			name,
+			pattern: pattern instanceof RegExp ? pattern.source : pattern,
 			placeholder,
 			readOnly,
 			'aria-readonly': readOnly ? 'true' : null,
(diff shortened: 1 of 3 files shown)


src/text-input/index.ts
```diff
@@ -52,7 +54,7 @@ export const ThemedBase = ThemedMixin(WidgetBase);
 		'labelAfter',
 		'labelHidden'
 	],
-	attributes: [ 'widgetId', 'label', 'placeholder', 'controls', 'type', 'minLength', 'maxLength', 'value', 'name' ],
+	attributes: [ 'widgetId', 'label', 'placeholder', 'controls', 'type', 'minLength', 'maxLength', 'value', 'name', 'pattern', 'autocomplete' ],
```
agubler: pattern will have to be a property if it's not just a string.

src/text-input/index.ts
```diff
@@ -33,6 +33,8 @@ export interface TextInputProperties extends ThemedProperties, InputProperties,
 	placeholder?: string;
 	value?: string;
 	shouldFocus?: boolean;
+	pattern?: string | RegExp;
+	autocomplete?: string;
```
agubler: we should make this a `boolean` and then reflect the right value depending on `true` or false (which would mean it needs to be in `properties` list for the custom element def too.

src/text-input/index.ts
```diff
@@ -33,6 +33,8 @@ export interface TextInputProperties extends ThemedProperties, InputProperties,
 	placeholder?: string;
 	value?: string;
 	shouldFocus?: boolean;
+	pattern?: string | RegExp;
```
agubler: Do you think we should a regular expression on the interface?

src/text-input/index.ts
bryanforbes: `autocomplete` can be more than `"on"` and `"off"`: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

src/text-input/index.ts
bryanforbes: I do because it's easier to catch RegExp errors in editors using RegExp syntax than in strings

src/text-input/index.ts
agubler: Urgh

agubler: `boolean | string`?

src/text-input/index.ts
agubler: It does I guess, but then it has to be a property for the custom elements, but it would be awesome if it was just an attribute. If you keep `RegExp` on the interface, you are going to have to implement a custom diff. Because the instance's expression can be changed but it would not trigger a re-render.

src/text-input/index.ts
bryanforbes: I talked with @matt-gadd today and that doesn't seem to be the case

PR #604, opened 2018-10-05
Themable Dialog animations

### LINKED ISSUE #578: Dialog animations are hard coded and cannot be changed using theming.
**Enhancement**

Dialog animations classes are hardcoded and cannot be changed using theming. It would be helpfull to have a class in the dialog.m.css to change default animations.

Package Version: 3.0.0

bryanforbes:
**Type:** bug / feature

**Description:**

Allows customization of animations via CSS

Resolves #578


src/dialog/index.ts
```diff
@@ -160,8 +159,8 @@ export class DialogBase<P extends DialogProperties = DialogProperties> extends T
 		const { underlay } = this.properties;
 		return v('div', {
 			classes: [ this.theme(underlay ? css.underlayVisible : null), fixedCss.underlay ],
-			enterAnimation: animations.fadeIn,
-			exitAnimation: animations.fadeOut,
+			enterAnimation: this.theme(css.underlayEnter) || undefined,
```
smhigley: We should probably update `SupportedClassName` typings in `framework/widget-core` should be updated so this isn't necessary 😝

src/dialog/index.ts
```diff
@@ -172,8 +171,8 @@ export class DialogBase<P extends DialogProperties = DialogProperties> extends T
 			aria = {},
 			closeable = true,
 			closeText,
-			enterAnimation = animations.fadeIn,
-			exitAnimation = animations.fadeOut,
+			enterAnimation = this.theme(css.enter) || undefined,
```
smhigley: Another thought -- would it make sense to add `enterAnimationActive` and `exitAnimationActive` properties to support both transitions and animations for increased styling flexibility? Using keyframes for a simple opacity fade in/out seems overkill, although I suppose any given transition could be made into an animation. Your call.

PR #607, opened 2018-10-09
Select: support keyboard selecting by partially typing label

### LINKED ISSUE #606: Select: support keyboard selecting by partially typing label
**Enhancement** 

Simulate some of the selecting behaviour from the native select box:

* typing characters in to the select when it is closed should select the best matching item
* typing characters in to the select when it is open should highlight the best matching item

matt-gadd:
**Type:** feature

**Description:**

Resolves #606

### DIFF
--- a/src/select/index.ts
+++ b/src/select/index.ts
@@ -35,7 +35,7 @@ export interface SelectProperties extends ThemedProperties, InputProperties, Lab
 	getOptionId?(option: any, index: number): string;
 	getOptionLabel?(option: any): DNode;
 	getOptionText?(option: any): string;
-	getSelectedIndexByInput?(keyboardKey: string): number | undefined;
+	getSelectedIndexOnInput?(keyboardKey: string): number | undefined;
 	getOptionSelected?(option: any, index: number): boolean;
 	getOptionValue?(option: any, index: number): string;
 	options?: any[];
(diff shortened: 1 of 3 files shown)


src/select/index.ts
```diff
@@ -94,6 +99,30 @@ export class SelectBase<P extends SelectProperties = SelectProperties> extends T
 		return getOptionValue ? getOptionValue(option, index) === value : option === value;
 	}
 
+	private _getSelectedIndexOnInput(event: KeyboardEvent) {
+		const { key, options = [], getOptionDisabled, getOptionText } = this.properties;
+		if (event.key !== undefined && event.key.length === 1) {
+			clearInterval(this._resetInputTextTimer);
```
pottedmeat: `clearTimeout`?

src/select/index.ts
```diff
@@ -94,6 +99,30 @@ export class SelectBase<P extends SelectProperties = SelectProperties> extends T
 		return getOptionValue ? getOptionValue(option, index) === value : option === value;
 	}
 
+	private _getSelectedIndexOnInput(event: KeyboardEvent) {
+		const { key, options = [], getOptionDisabled, getOptionText } = this.properties;
```
pottedmeat: Is `key` used in this function?

src/select/index.ts
matt-gadd: good catch, updated 👍

src/select/index.ts
matt-gadd: nope, leftover 😄. removed.

PR #611, opened 2018-10-26
Range Slider

### LINKED ISSUE #554: Request for dojox.form.RangeSlider replacement (new widget or enhancement to slider)
**Enhancement**

Request for a replacement of the `dojox.form.RangeSlider` (horizontal/vertical) widget from Dojo 1.

rorticus:
**Type:** bug / feature

**Description:**

Adding range slider control.

image

Resolves #554


rorticus:
Thanks for the feedback @JamesMilnerUK ! I've resolved everything except for the comment about using `postcscs-input-style` as i feel that might be a larger discussion.

src/range-slider/styles/range-slider.m.css
```diff
@@ -0,0 +1,119 @@
...
+	color: transparent;
+	height: 100%;
+	margin: 0px;
+	overflow: visible;
+	padding: 0px;
+	width: 100%;
+}
+
+.nativeInput::-moz-range-track {
+	appearance: none;
```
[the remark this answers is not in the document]
rorticus: Interestingly, if I add this style to the rule above (with a comma), IE 11 no longer looks at this rule...

rorticus:
@smhigley , I added a README to the range-slider and updated the project README to include a link.

src/range-slider/index.ts
```diff
@@ -0,0 +1,372 @@
...
+			return (e?: Event) => {
+				callback(property, ...[isSlider1 ? e : undefined, !isSlider1 ? e : undefined]);
+			};
+		};
+
+		return {
+			...formatAriaProperties(aria),
+			'aria-invalid': invalid === true ? 'true' : null,
+			'aria-readonly': readOnly === true ? 'true' : null,
+			'aria-label': isSlider1 ? minimumLabel : maximumLabel,
```
smhigley: Sorry I didn't catch this before! This is a pretty good solution, but I think the inputs need to be labeled by the `label` property's text in some way. Since it's not simple to get the plain text from `label`, it might be better to point `aria-labelledby` from both of the `input`s to the `label`, then use `aria-describedby` for min/max text? I really wish there were an `aria-description` attribute :/

src/range-slider/index.ts
```diff
@@ -0,0 +1,372 @@
...
+			label ? w(Label, {
+				key: 'label',
+				theme,
+				disabled,
+				focused: focus.containsFocus,
+				invalid,
+				readOnly,
+				required,
+				hidden: labelHidden,
+				forId: widgetId
```
smhigley: Since this is pointing to the root `div`, and that div doesn't have any region/etc attributes to make it a landmark (which makes sense, since it isn't a landmark 😁), the `forId` isn't really doing anything. If you use `aria-labelledby` on the `input`s, you could drop this entirely

src/range-slider/index.ts
```diff
@@ -0,0 +1,372 @@
...
+
+		const size = this.meta(Dimensions).get('root');
+
+		const slider1Focus = this.meta(Focus).get('slider1');
+		const slider2Focus = this.meta(Focus).get('slider2');
+
+		const slider1 = v('input', {
+			...this._getInputProperties(true),
+			key: 'slider1',
+			value: `${minValue}`,
```
smhigley: Is there any drawback to setting `max` here and `min` on the next input to the `maxValue` and `minValue`? I think it could be useful, but up to you :)

rorticus: Yeah, so setting the max/min like that changes the position of the thumb within the slider (since the total range of the slider changes) and I couldn't get things to smoothly move when sliding the sliders around

PR #620, opened 2018-11-09
add calendar min max properties

### LINKED ISSUE #538: Specifying max/min selectable date in Calendar?
**Enhancement**

Package Version: master

Is it possible to add min/max selectable date constraints to the calendar. It would be nice that we have something like that, similar to what DOJO1 date picker calendar had.

aciccarello:
**Type:** Feature

* [ ] Unit or Functional tests are included in the PR (TBD)

**Description:** This PR adds `minDate`/`maxDate` properties to the `DatePicker` and `Calendar` widgets. It also modifies the `CalendarCell` widget to separate disabled cells from the `currentMonth` indicator.

The previous/next month buttons are disabled if the next month is outside the min/max. The radio buttons on the month/date select windows are disabled if the month is outside the min/max but the links are not removed (though they don't work). Currently there is not separate styling for disabled vs other-month cells.

calendar-min-max

Resolves #538

### DIFF
--- a/src/calendar/index.ts
+++ b/src/calendar/index.ts
@@ -295,7 +295,7 @@ export class CalendarBase<P extends CalendarProperties = CalendarProperties> ext
 		} = this._getMonthYear();
 		const {
 			onMonthChange,
-			onYearChange,
+			onYearChange
 		} = this.properties;
 
 		if (!this._monthInMin(month === 0 ? year - 1 : year, month === 0 ? 11 : month - 1)) {
--- a/src/calendar/index.ts
+++ b/src/calendar/index.ts
@@ -351,7 +351,7 @@ export class CalendarBase<P extends CalendarProperties = CalendarProperties> ext
 
 		const date = dateObj.getDate();
 		const { theme, classes } = this.properties;
-		const outOfRange = ((minDate && dateObj < minDate) || (maxDate && dateObj > maxDate));
+		const outOfRange = Boolean((minDate && dateObj < minDate) || (maxDate && dateObj > maxDate));
 		const focusable = currentMonth && date === this._focusedDay;
 
 		return w(CalendarCell, {
--- a/src/calendar/index.ts
+++ b/src/calendar/index.ts
@@ -272,14 +272,14 @@ export class CalendarBase<P extends CalendarProperties = CalendarProperties> ext
 
 	private _ensureDayIsInMinMax(newDate: Date, update: (day: number) => void) {
 		const {
-			minDate = newDate,
-			maxDate = newDate
+			minDate,
+			maxDate
 		} = this.properties;
 
-		if (newDate < minDate) {
+		if (minDate && newDate < minDate) {
 			update(minDate.getDate());
 		}
-		else if (newDate > maxDate) {
+		else if (maxDate && newDate > maxDate) {
 			update(maxDate.getDate());
 		}
 	}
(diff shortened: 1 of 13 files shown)


src/calendar/index.ts
```diff
@@ -121,11 +125,22 @@ export class CalendarBase<P extends CalendarProperties = CalendarProperties> ext
 		const {
 			month,
 			selectedDate = this._defaultDate,
-			year
+			year,
+			minDate,
+			maxDate
 		} = this.properties;
+
+		let selected = selectedDate;
```
tomdye: If I follow this right, it's going to set the `selected` date to the min/max if it's out of range rather than setting the component to invalid or not selecting anything. Is this the desired effect? Do we think this is a good user experience?

src/calendar/index.ts
```diff
@@ -158,16 +177,16 @@ export class CalendarBase<P extends CalendarProperties = CalendarProperties> ext
 		this.invalidate();
 	}
 
-	private _onDateClick(date: number, disabled: boolean) {
+	private _onDateClick(date: number, disabled: boolean, currentMonth: boolean) {
+		if (disabled) {
```
tomdye: this seem's a strange place do to this, I'd have expected the `calendarCell` to simply not call the `onClick` if it's disabled / invalid

src/calendar/index.ts
aciccarello: We actually don't call this if the date is out of range so this can be removed.

aciccarello:
I've added additional tests and fixed some implementation bugs. Still working on improving the calendar tests and functional tests.

aciccarello:
When unit testing, I realized that we allow the user to pass selected dates outside the min/max. I wasn't sure how this should be handled so there is a unit test which is failing. Two of the untested lines of code seem like tests for things that won't happen due to constraints that prevent the callbacks from being called with an invalid date. We could consider removing those checks.

aciccarello:
Fixed an accessibility issue with the focused date. If the focused date was outside the min/max, the calendar cells would not respond to keyboard navigation.

For now, the calendar does not do anything special when the selected date is outside the min/max. The calendar is fully interactive and the selection will change when the user selects the next date.

Some future improvements which I think should be done in a follow-up PR would include styling for disabled cells. Currently the disabled cells are styled the same as the other months. Additionally, a disabled, selected cell is styled the same as a selected cell.

src/calendar/index.ts
```diff
@@ -210,20 +218,58 @@ export class CalendarBase<P extends CalendarProperties = CalendarProperties> ext
...
 			year
 		} = this._getMonthYear();
 		const {
 			onMonthChange,
-			onYearChange
+			onYearChange,
+			minDate
 		} = this.properties;
 
+		if (!monthInMin(year, month - 1, minDate)) {
```
aciccarello: I don't think that it is possible to reach inside of this block (I wasn't able to figure out how to test it). We may be able to remove it.

Same with line [295](https://github.com/dojo/widgets/pull/620/files#diff-663a5c4d6e355c4bd09383f0b5119996R295).

src/calendar/index.ts
```diff
@@ -324,20 +358,24 @@ export class CalendarBase<P extends CalendarProperties = CalendarProperties> ext
 		return weeks;
 	}
 
-	protected renderDateCell(date: number, index: number, selected: boolean, currentMonth: boolean, today: boolean): DNode {
+	protected renderDateCell(dateObj: Date, index: number, selected: boolean, currentMonth: boolean, today: boolean): DNode {
```
aciccarello: If someone extends this class and overrides this method, the change from `date` to `dateObj` could be a breaking change.

tomdye: we're moving away from those kind of extensions in favour of custom render properties etc so I shouldn't worry.

tomdye: Why did this need to be changed to a dateObj btw?

src/calendar/index.ts
```diff
@@ -324,20 +343,29 @@ export class CalendarBase<P extends CalendarProperties = CalendarProperties> ext
...
+		const {
+			minDate = dateObj,
+			maxDate = dateObj
+		} = this.properties;
+
+		const date = dateObj.getDate();
 		const { theme, classes } = this.properties;
+		const outOfRange = (dateObj < minDate || dateObj > maxDate);
```
tomdye: how about conditionally running this check rather than defaulting the min / max values above?

src/calendar/index.ts
aciccarello: Okay, great. I figured this was an unlikely method to override anyway.

I switched to a date object because we need to do the min/max. We were already creating the object in the caller of this function but were were just using it for string comparison.

src/calendar/index.ts
aciccarello: Sure, I can change this. I wen't back and forth on how to write it. This was a little more legible but might be slightly less performant.

PR #631, opened 2018-11-22
Grid propagate theme down to all child widgets

agubler:
**Type:** bug

**Description:**

Pass `theme` property to all child widgets.


agubler:
Perhaps we should back port this to 4.0.0, it should enable complete customisation of the grid css... not the most ideal way as would require composing the one off theme with the application theme and tweaking the parts that need changing but until we have the `classes` API available I think it could work.

PR #642, opened 2019-01-16
Added overflow-y: auto; to slide-pane

### LINKED ISSUE #641: Content which does not fit into the slide-pane is cut out and cannot be reached.
**Enhancement**

When I am using the slide-pane widget with contents bigger than the slide-pane size I want to be able to scroll down so that I can see each parts of the slide-pane's contents.

I think https://github.com/dojo/widgets/issues/454 is similar to this one.

I also discussed certain issues on discord, where @agubler helped me to apply different styles to links when the slide-pane opens.

Package Version: 4.0.2-pre

**Code**

When you put more lorem ipsum contents into the example code of the slide-pane (See https://github.com/SimonScholz/widgets/commit/e4b185e6b1fc0e035483930d84361305a7d7f6ea) , build the examples with _grunt dev_, select the _slide-pane_ example, press _open slidepane_ and make the window smaller or use the mobile device viewer in chrome dev tools you will the  see that some parts of the contents is cut out and cannot be reached.

**Expected behavior:**

The slide-pane becomes scrollable once it has more contents which does not fit in the slide-pane any more.

**Actual behavior:**

Content which does not fit into the slide-pane is cut out and cannot be reached.

SimonScholz:
**Type:** bug / feature

* [ x ] There is a related issue
* [ x ] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)

Not sure whether Unit or Functional tests are necessary for this PR, since I just changed the CSS.

**Description:**

Resolves #641

### DIFF
--- a/src/slide-pane/styles/slide-pane.m.css
+++ b/src/slide-pane/styles/slide-pane.m.css
@@ -9,7 +9,13 @@
 .paneFixed {
 	position: fixed;
 	box-sizing: border-box;
+}
+
+.contentFixed {
 	overflow-y: auto;
+	position: fixed;
+	margin-top: 5px;
+	height: 100%;
 }
 
 .leftFixed,
--- a/src/slide-pane/styles/slide-pane.m.css
+++ b/src/slide-pane/styles/slide-pane.m.css
@@ -9,11 +9,13 @@
 .paneFixed {
 	position: fixed;
 	box-sizing: border-box;
+	display: flex;
+	flex-direction: column;
 }
 
 .contentFixed {
 	overflow-y: auto;
-	height: 100%;
+	flex-grow: 1;
 }
 
 .leftFixed,
--- a/src/slide-pane/styles/slide-pane.m.css
+++ b/src/slide-pane/styles/slide-pane.m.css
@@ -9,6 +9,7 @@
 .paneFixed {
 	position: fixed;
 	box-sizing: border-box;
+	overflow-y: auto;
 }
 
 .leftFixed,
--- a/src/slide-pane/styles/slide-pane.m.css
+++ b/src/slide-pane/styles/slide-pane.m.css
@@ -13,8 +13,6 @@
 
 .contentFixed {
 	overflow-y: auto;
-	position: fixed;
-	margin-top: 5px;
 	height: 100%;
 }
 
--- a/src/slide-pane/example/index.ts
+++ b/src/slide-pane/example/index.ts
@@ -38,6 +38,27 @@ export default class App extends WidgetBase {
 			}, [
 				`Lorem ipsum dolor sit amet, consectetur adipiscing elit.
 				Quisque id purus ipsum. Aenean ac purus purus.
+				Nam sollicitudin varius augue, sed lacinia felis tempor in.
+				Lorem ipsum dolor sit amet, consectetur adipiscing elit.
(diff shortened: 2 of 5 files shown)


src/slide-pane/styles/slide-pane.m.css
```diff
@@ -11,6 +11,13 @@
 	box-sizing: border-box;
 }
 
+.contentFixed {
+	overflow-y: auto;
+	position: fixed;
```
tomdye: I believe this should just be the `overflow-y` style, we wouldn't want to apply `position: fixed` or a set margin-top here

SimonScholz: Ok just using height 100% did the trick as well. But without it, it does not work.

SimonScholz: So I removed position:fixed; and margin-top with my lastest commit

SimonScholz:
Unfortunately this change breaks the slide pane tests, it seems to me that the harness.expect now also needs the new _fixedCss.contentFixed_, right?
Can you tell me how to properly fix this?

tomdye:
Morning @SimonScholz , I've checked out your code to take a look and I noticed that the `height: 100%` addition appears to push the content off of the bottom of the page. I believe that the correct way to deal with this is to use `display: flex;` as this will allow the content to take up the full height and adapt to any layout changes.

To implement this you'll need to change the `paneFixed` and `contentFixed` classes to:

```css
.paneFixed {
	position: fixed;
	box-sizing: border-box;
	display: flex;
	flex-direction: column;
}

.contentFixed {
	overflow-y: auto;
	flex-grow: 1;
}
```

To fix the issue you are having with your tests, you'll need to add your new `fixedCss.contentFixed` class to the `slide-pane` tests [here](https://github.com/dojo/widgets/blob/master/src/slide-pane/tests/unit/SlidePane.ts#L78) and anywhere else that you see the `css.content` class being used within that test.

Hope this unblocks you, please let me know here or on Discord if you need any more assistance. Thanks again for contributing. 👍

SimonScholz:
> Morning @SimonScholz , I've checked out your code to take a look and I noticed that the `height: 100%` addition appears to push the content off of the bottom of the page. I believe that the correct way to deal with this is to use `display: flex;` as this will allow the content to take up the full height and adapt to any layout changes.
> 
> To implement this you'll need to change the `paneFixed` and `contentFixed` classes to:
> 
> ```css
> .paneFixed {
> 	position: fixed;
> 	box-sizing: border-box;
> 	display: flex;
> 	flex-direction: column;
> }
> 
> .contentFixed {
> 	overflow-y: auto;
> 	flex-grow: 1;
> }
> ```
> 

This fixes the scroll problems. Thanks!

> To fix the issue you are having with your tests, you'll need to add your new `fixedCss.contentFixed` class to the `slide-pane` tests [here](https://github.com/dojo/widgets/blob/master/src/slide-pane/tests/unit/SlidePane.ts#L78) and anywhere else that you see the `css.content` class being used within that test.
> 

Thanks a lot for the pointer. Tests have been fixed by that.

> Hope this unblocks you, please let me know here or on Discord if you need any more assistance.
> Thanks again for contributing. +1

It's a pleasure to help and being a part of the community. Open Source rocks ;-)

tomdye:
Thanks @SimonScholz

PR #647, opened 2019-01-17
Ensure that the store id as the path to subscribe to changes

### LINKED ISSUE #639: Grid does not invalidate when using a custom storeId
**Bug**

When a store is passed to the grid, with a custom store path id the grid does not invalidate correctly as it always incorrectly listens to the default path `_grid`.

agubler:
**Type:** bug

**Description:**

Subscribe to the store using the custom `storeId` when passed.

Resolves #639


agubler:
@tomdye I think handle will aways be set because it is initialised in the constructor.

Issue #656, opened 2019-02-04
Select widget doesn't scroll to the initial selection


evra:
**Bug**

The `Select` widget doesn't scroll to the active index selection on initial render. 

Package Version: 4.0.1

- A `select` widget has a long list of items so that the underlying `Listbox` displays a vertical scroll.
- The last item in the list is marked as active selection in the `Select` properties

**Expected behavior:**

When user opens a `Scroll` list the list is automatically scrolled to the selected item.

**Actual behavior:**

The list doesn't automatically scroll.

Please, note that this happens only for the initial render. If user scrolls down manually or uses a keyboard to select a matching item then the scroll position stays at a correct position (probably using an internal state).

maier49:
This was addressed by #660

PR #660, opened 2019-02-14
Listbox and select fixes

### LINKED ISSUE #659: Native select renders inappropriate `value` attributes
**Bug** 

When passed a simple array of strings as the `options` property, the native mode of Select is rendering `value` attributes on each child `option` tag. This results in every selection being interpreted as setting the `select` to `""`. As a result the `onChange` handler is not fired and the select's value does not properly update.

Package Version: 5.0.2

**Code**
```ts
w(Select, { 
  options: ['foo', 'bar'],
  useNativeElement: true
});
```

**Expected behavior:**

The `value` attribute should be empty in this case so the text of the `option` element is used as the value.
**Actual behavior:**

A `value` attribute with no value is being added to each `option`, causing the select to have incorrectly.

maier49:
* Fix listbox and select initial value behavior

* Fix native select behavior with simple options

**Type:** bug

**Description:**
There were a couple of related issues causing the behavior described in 659 and 656. The select widget was passing `''` as the value for `option` elements when there was no value, resulting in empty `value` attributes after some related vdom fixes. It was also not properly treating the option text as the value when no value was specified. This resulted in `onChange` never being called and selections not being persisted for the native select.

Additionally, the select was not passing in the appropriate `activeIndex` to the listbox in the custom select scenario, so an initial selected value would be selected but not displayed. Fixing this revealed a related issue in listbox as it was not properly scrolling to an initially provided `activeIndex`. I'm not sure I'm using the correct approach for this last fix. The issue is that the `activeIndex` property diff for an initial value runs when the element has a height of `0`, so it does not scroll. I attempted to resolve this using an `@afterRender` decorator to check for the first render where the element has height and perform a scroll calculation at that time to handle the initial value of `activeIndex`. 

Resolves #659 #656

### DIFF
--- a/src/listbox/index.ts
+++ b/src/listbox/index.ts
@@ -174,17 +174,8 @@ export class ListboxBase<P extends ListboxProperties = ListboxProperties> extend
 		}
 	}
 
-	protected animateScroll(scrollValue: number) {
-		this.meta(ScrollMeta).scroll('root', scrollValue);
-	}
-
-	@diffProperty('activeIndex', auto)
-	protected calculateScroll(previousProperties: ListboxProperties, newProperties: ListboxProperties) {
-		this._calculateScroll(newProperties);
-	}
-
 	@afterRender()
-	protected afterRender(dNode: DNode) {
+	private afterRender(dNode: DNode) {
 		if (!this._rendered) {
 			const dimensionMeta = this.meta(Dimensions);
 			if (dimensionMeta && dimensionMeta.get('root').offset.height) {
@@ -194,6 +185,15 @@ export class ListboxBase<P extends ListboxProperties = ListboxProperties> extend
 		}
 
 		return dNode;
+	}
+
+	protected animateScroll(scrollValue: number) {
+		this.meta(ScrollMeta).scroll('root', scrollValue);
+	}
+
+	@diffProperty('activeIndex', auto)
+	protected calculateScroll(previousProperties: ListboxProperties, newProperties: ListboxProperties) {
+		this._calculateScroll(newProperties);
 	}
 
 	protected getModifierClasses() {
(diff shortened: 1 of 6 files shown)


src/listbox/index.ts
```diff
@@ -177,6 +174,28 @@ export class ListboxBase<P extends ListboxProperties = ListboxProperties> extend
...
+		this.meta(ScrollMeta).scroll('root', scrollValue);
+	}
+
+	@diffProperty('activeIndex', auto)
+	protected calculateScroll(previousProperties: ListboxProperties, newProperties: ListboxProperties) {
+		this._calculateScroll(newProperties);
+	}
+
+	@afterRender()
+	protected afterRender(dNode: DNode) {
```
tomdye: Could this be done via deferred properties instead?

maier49: I'm not sure I know what you mean by a deferred property, but it certainly sounds useful in this situation.

maier49: Addressed this by using `onAttach` instead of an `afterRender`, as this scrolling only needs to be a one time thing

src/listbox/index.ts
```diff
@@ -1,5 +1,6 @@
 import { auto, reference } from '@dojo/framework/widget-core/diff';
 import { diffProperty } from '@dojo/framework/widget-core/decorators/diffProperty';
+import { afterRender } from '@dojo/framework/widget-core/decorators/afterRender';
```
tomdye: do you still need this import?

maier49: No, good catch. I've removed that now

maier49:
@tomdye I've updated this to use the resize meta instead of `diffProperty` and `onAttach`. Seems to be working fine

Issue #616, opened 2018-11-09
Text-Input widget


tomdye:
- mdc readme https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield
- mdc docs https://material.io/develop/web/components/input-controls/text-field/
- mdc demo https://material-components.github.io/material-components-web-catalog/#/component/text-field

<img width="258" alt="screenshot 2018-11-09 13 09 46" src="https://user-images.githubusercontent.com/814453/48264255-c0d3ea00-e420-11e8-89c7-ba2cdb6ff978.png">

## Overview

This issue is concerned with implementing a material themed text input component.
Native browser validation will be added to the widget controlled by a `validate` flag. A custom validator may also be used.

## Changes to structure

- Widget will be expanded to take arguments of `leading: DNode` and  `trailing: DNode`. This will replace `Enhanced-Text-Input`.
- `label` if specified will be rendered `before` the widget and the option to render `after` will be removed. Users wishing to use a `label` after an input can do so by manually rendering a `label` widget and linking it to the input via `for` and `id`.

## New properties

- `onValue(value: string) => void`: callback upon value change, this will replace `onInput` callback.
- `onKey(key: number, preventDefault: () => void)`: callback on `keyPress`
- `leading: DNode` will render a leading section before the input
- `trailing: DNode` will render a trailing section after the input
- `validate: boolean | (value: string) => { message: string; valid: boolean }`: flag to indicate if widget should validate or callback to perform custom validation.
- `onValidate: (valid: boolean) => void`: callback upon validation to inform parent of valid state.

## Breaking changes

- will no longer offer `labelAfter` property
- will no longer call an `onChange` callback
- will no longer call `keyDown` / `keyPress`
- specific touch events have been removed
- `onBlur` / `onFocus` will no longer be passed the `value`

## Future work / properties

- `outlined: boolean` renders the text input with an outline rather than with a line beneath it (this is a material property)
- `fullWidth: boolean` renders the text input full width  and treats the label differently (this is a material property)
- `helperText: string` used to provide helper text beneath the text-input (this is a material property)

## Tasks

- [ ] Restructure and update props
- [ ] Implement theme and update dojo theme
- [ ] Add validation

tomdye:
This has been completed

PR #664, opened 2019-02-19
Add validation support for text-input

nicknisi:
**Type:** feature

**Description:**

This adds validation support as partially described in #616. The user can provide a `validate` property to the widget, which if `true`, will validate against the `type` (`"email"`, for example) and use the browser's built-in validation. 

Alternatively, `validate` can be a function that accepts the widget's current value as an argument and must return `{ valid: boolean; message: string | undefined; }`. The user can also provide an `onValidate(valid: boolean, message?: string)` which will be called to pass the validation value back up to the parent, along with any potential message.

With the current behavior, if `validate` is a function, it is the primary validation, but the other validation (email, for example) is also being checked.

**Example:**

```typescript
w(TextInput, {
	key: 't2',
	type: 'email',
	label: 'Email (required)',
	required: true,
	// validate: true,
	validate(value: string) {
		const valid = !!~value.indexOf('nicknisi');
		return {
			valid,
			message: !valid ? 'Must include nicknisi' : ''
		};
	},
	onValidate(valid: boolean, message: string | undefined) {
		const color = valid ? 'green' : 'red';
		console.log(`%c${message}`, `font-weight:bold;color:${color};`, { valid });
	},
	value: this._value2,
	onInput: (value: string) => {
		this._value2 = value;
		this.invalidate();
	}
})
```

### Breaking changes

This PR removes the `invalid` property. The way to mark an input as invalid is to use custom validation. This also slightly affects combobox and enhanced text input as their `invalid` properties will no longer be passed down to the underlying text-input widget.


src/text-input/index.ts
```diff
@@ -11,9 +11,12 @@ import { uuid } from '@dojo/framework/core/util';
 import * as css from '../theme/text-input.m.css';
 import { customElement } from '@dojo/framework/widget-core/decorators/customElement';
 import diffProperty from '@dojo/framework/widget-core/decorators/diffProperty';
+import ValidityMeta from './ValidityMeta';
 
 export type TextInputType = 'text' | 'email' | 'number' | 'password' | 'search' | 'tel' | 'url';
 
+type TextInputInternalState = any;
```
tomdye: does this need typing here like this?

src/text-input/index.ts
```diff
@@ -166,13 +202,14 @@ export class TextInputBase<P extends TextInputProperties = TextInputProperties>
...
 		} = this.properties;
+		const { valid } = this._state;
 		const focus = this.meta(Focus).get('root');
 		return [
 			css.root,
 			disabled ? css.disabled : null,
 			focus.containsFocus ? css.focused : null,
-			invalid === true ? css.invalid : null,
-			invalid === false ? css.valid : null,
+			invalid === true || valid === false ? css.invalid : null,
```
tomdye: If we're adding this we will want to remove support for passing `invalid` manually.

src/text-input/ValidityMeta.ts
```diff
@@ -0,0 +1,25 @@
...
+		}
+
+		return {
+			valid: node.validity.valid,
+			message: node.validationMessage
+		};
+	}
+}
+
+export default InputValidity;
```
tomdye: perhaps move this to `src/common` for now?

src/text-input/index.ts
```diff
@@ -27,16 +37,19 @@ export type TextInputType = 'text' | 'email' | 'number' | 'password' | 'search'
  * @property value           The current value
  */
 
-export interface TextInputProperties extends ThemedProperties, InputProperties, FocusProperties, LabeledProperties, PointerEventProperties, KeyEventProperties, InputEventProperties, CustomAriaProperties {
+export interface TextInputProperties extends ThemedProperties, Omit<InputProperties, 'invalid'>, FocusProperties, LabeledProperties, PointerEventProperties, KeyEventProperties, InputEventProperties, CustomAriaProperties {
```
tomdye: May well be better if you just inline the appropriate properties from `InputProperties` here. Probably going to phase out all these extended properties

src/text-input/index.ts
```diff
@@ -152,27 +165,54 @@ export class TextInputBase<P extends TextInputProperties = TextInputProperties>
...
+			return;
+		}
+
+		let { valid, message } = this.meta(ValidityMeta).get('input', value);
 
-	constructor() {
-		super();
-		this._uuid = uuid();
+		if (typeof validate === 'function') {
+			const { valid: customValid, message: customMessage } = validate(value);
```
tomdye: do these need typing here given that `validate` is typed?

src/text-input/index.ts
```diff
@@ -231,34 +272,60 @@ export class TextInputBase<P extends TextInputProperties = TextInputProperties>
 		});
 	}
 
+	protected renderHelperText(): DNode {
+		const { properties: { helperText }, _state: { message } } = this;
+		const text = message || helperText;
+		const invalid = !!message;
```
tomdye: why don't you use the `_state.valid` boolean here?

src/text-input/index.ts
nicknisi: Well, `validate` could be a boolean, a function, or `undefined`, so I need to type guard it and treat it and call it if it is a function.

src/text-input/ValidityMeta.ts
nicknisi: Moved!

src/text-input/index.ts
nicknisi: Done. I copied `InputProperties` into this interface and then removed the `Omit`.

src/text-input/index.ts
nicknisi: Fixed.

src/text-input/README.md
```diff
@@ -20,8 +20,10 @@ If the `label` property is not used, we recommend creating a separate `label` an
 ```typescript
 // Basic usage
 w(TextInput, {
```
tomdye: can you search for other mentions of `invalid` in this file and correct them please?

src/text-input/index.ts
```diff
@@ -231,34 +273,59 @@ export class TextInputBase<P extends TextInputProperties = TextInputProperties>
 		});
 	}
 
+	protected renderHelperText(): DNode {
+		const { properties: { helperText }, _state: { valid, message } } = this;
+		const text = message || helperText;
+
+		return text ? v('div', {
```
tomdye: should this perhaps be an `aside` and the text itself a `span` ? could be wrong, but just thinking out loud.

src/text-input/tests/unit/TextInput.ts
```diff
@@ -7,21 +7,32 @@ import { v, w } from '@dojo/framework/widget-core/d';
...
 import Label from '../../../label/index';
-import TextInput from '../../index';
+import TextInput, { TextInputProperties } from '../../index';
+import InputValidity from '../../../common/InputValidity';
 import * as css from '../../../theme/text-input.m.css';
 import { compareForId, compareId, createHarness, MockMetaMixin, noop, stubEvent } from '../../../common/tests/support/test-helpers';
 
 const harness = createHarness([ compareId, compareForId ]);
 
+const sleep = (time: number = 0) => new Promise((resolve) => setTimeout(resolve, time));
```
tomdye: I don't believe this is used anywhere?

nicknisi: Deleted.

src/text-input/index.ts
nicknisi: Either is fine with me. I was just using the same markup as in the example [here](https://material.io/develop/web/components/input-controls/text-field/).
screen shot 2019-02-25 at 1 51 46 pm

src/text-input/README.md
nicknisi: I also added a small section on validation to try and make things a bit more clear.

nicknisi:
@tomdye I made changes to how Combobox internally uses `invalid`. Instead of just passing its `invalid` value down to the underlying TextInput, it instead sets `validate` to a function that passes the inverse of `invalid` if `invalid` is a boolean. It does nothing if `invalid` is `undefined`.

src/text-input/index.ts
tomdye: fair enough then!

PR #666, opened 2019-02-22
Use resize observer in split pane

### LINKED ISSUE #665: Split Pane does not render correctly in closed tab
**Bug** 

If a split pane is rendered inside a tab pane which is initially closed, it is not laid out properly and does not render correctly until the page has been resized. A potential fix for this is to use the resize observer instead of the global resize event.

The larger, breaking, fix is to change the tab controller to render / not render tables rather than show / hide them.

Example here:
https://codesandbox.io/s/r7w65po7n

tomdye:
**Type:** bug

**Description:**

Changes global resize observer to resizeObserver. Tidies up some implementation.

- No longer defaults to min collapsewidth of 600

Resolves #665

### DIFF
--- a/src/split-pane/index.ts
+++ b/src/split-pane/index.ts
@@ -8,6 +8,7 @@ import { diffProperty } from '@dojo/framework/widget-core/decorators/diffPropert
 import * as fixedCss from './styles/split-pane.m.css';
 import * as css from '../theme/split-pane.m.css';
 import { Dimensions } from '@dojo/framework/widget-core/meta/Dimensions';
+import { Resize, ContentRect } from '@dojo/framework/widget-core/meta/Resize';
 import { GlobalEvent } from '../global-event/index';
 import { customElement } from '@dojo/framework/widget-core/decorators/customElement';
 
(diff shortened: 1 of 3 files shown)


src/split-pane/index.ts
```diff
@@ -55,6 +56,23 @@ export class SplitPaneBase<P extends SplitPaneProperties = SplitPaneProperties>
...
+	private _width = 0;
+
+	@diffProperty('collapseWidth')
+	@diffProperty('direction')
+	private collapseWidthDiff(oldProperties: SplitPaneProperties, { collapseWidth, direction, onCollapse }: SplitPaneProperties) {
+		if (collapseWidth) {
+			if (direction === Direction.row || collapseWidth < this._width && this._collapsed) {
+				this._collapsed = false;
+				this._resizeResultOverridden = true;
+			} else if (collapseWidth > this._width && !this._collapsed) {
```
sammenza: I'm new here so not sure about this but the style guide says the closing brace of `if` shouldn't be cuddled with the `else if` statement

tomdye: I think we may need to update those style guides. To be honest, most of our repo's use `prettier` now.

src/split-pane/index.ts
```diff
@@ -55,6 +56,23 @@ export class SplitPaneBase<P extends SplitPaneProperties = SplitPaneProperties>
...
+	private _resizeResultOverridden = false;
+	private _width = 0;
+
+	@diffProperty('collapseWidth')
+	@diffProperty('direction')
+	private collapseWidthDiff(oldProperties: SplitPaneProperties, { collapseWidth, direction, onCollapse }: SplitPaneProperties) {
+		if (collapseWidth) {
+			if (direction === Direction.row || collapseWidth < this._width && this._collapsed) {
+				this._collapsed = false;
+				this._resizeResultOverridden = true;
```
maier49: Should `onCollapse(false)` be getting called in this case? It looks like it would have been under the old logic.

src/split-pane/index.ts
```diff
@@ -55,6 +56,23 @@ export class SplitPaneBase<P extends SplitPaneProperties = SplitPaneProperties>
 	private _lastSize?: number;
 	private _position = 0;
 	private _collapsed = false;
+	private _resizeResultOverridden = false;
+	private _width = 0;
+
+	@diffProperty('collapseWidth')
+	@diffProperty('direction')
+	private collapseWidthDiff(oldProperties: SplitPaneProperties, { collapseWidth, direction, onCollapse }: SplitPaneProperties) {
```
maier49: It looks like you're effectively removing the default `collapseWidth` value of 600 that was set in the old `diffProperty`s. Is that a breaking change?

src/split-pane/index.ts
tomdye: good eye

src/split-pane/index.ts
tomdye: again, good eye, thanks

PR #679, opened 2019-03-04
Simplify Input Validation

nicknisi:
**Type:** feature

**Description:**

Simplify text-input validation

- remove `validate` property
- add `valid?: boolean | { valid?: boolean; message?: string; }` property
- call `onValidate` and pass message and valid from meta
- onValidate is now a callback to receive native validation information from the input. Custom validation can now be achieved simply by setting the `valid` property.
- helperText will show a message if the message value exists. Otherwise, it'll show the set helperText.


src/text-input/index.ts
```diff
@@ -274,16 +269,18 @@ export class TextInputBase<P extends TextInputProperties = TextInputProperties>
 	}
 
 	protected renderHelperText(): DNode {
-		const { properties: { helperText }, _state: { valid, message } } = this;
-		const text = message || helperText;
+		const { properties: { helperText }, validity: { valid, message } } = this;
+		const text = (valid && message) || helperText;
```
tomdye: I think this should be `!valid && message`

src/text-input/index.ts
```diff
@@ -47,10 +48,11 @@ export interface TextInputProperties extends ThemedProperties, FocusProperties,
 	helperText?: string;
 	value?: string;
 	valid?: { valid?: boolean; message?: string; } | boolean;
+	customValidator?: (value: string) => { valid?: boolean; message?: string; } | void;
 	pattern?: string | RegExp;
 	autocomplete?: boolean | string;
 	onClick?(value: string): void;
-	onValidate?: (valid: boolean, message?: string) => void;
+	onValidate?: (valid?: boolean, message?: string) => void;
```
tomdye: is it correct that this is optional? I thought that it would always be passed but it may be `undefined`, in which case perhaps it should be `onValidate?: (valid: boolean | undefined, message?: string) => void;` ?

src/text-input/index.ts
```diff
@@ -16,8 +16,9 @@ import InputValidity from '../common/InputValidity';
 export type TextInputType = 'text' | 'email' | 'number' | 'password' | 'search' | 'tel' | 'url';
 
 interface TextInputInternalState {
-	valid: boolean | undefined;
-	message: string;
+	value?: string;
```
tomdye: do you think it would be clearer if these were prefixed with `previous` ?

PR #685, opened 2019-03-11
Export classes rather than bases

### LINKED ISSUE #684: Remove `Base` exports with generics from widgets and simply export the widget
**Enhancement** 

Currently we define out widgets like the following:

`export class ComboBoxBase<P extends ComboBoxProperties = ComboBoxProperties> extends ThemedBase<P, null> {`

This allows extension of widgets, however, we would be better to allow greater customisation where required via custom render functions etc.
Widgets such as the above would now simple be defined as:

`export class ComboBox extends WidgetBase<ComboBoxProperties>`

tomdye:

**Description:**

Remove bases with generic typings from widgets.
This is part of a move towards encouraging extension via properties instead of class extension. The abstraction involved with having `protected` functions internally and ensuring we don't break those api's is hard to manage and this is not the way we wish to support people using and customising our widgets etc.

Resolves #684

### DIFF
--- a/src/split-pane/index.ts
+++ b/src/split-pane/index.ts
@@ -131,8 +131,8 @@ export class SplitPane extends ThemedMixin(WidgetBase)<SplitPaneProperties> {
 		return content ? [ content ] : [];
 	}
 
-	private _shouldCollapse(dimensions: ContentRect) {
-		const { onCollapse, direction, collapseWidth } = this.properties;
+	private _shouldCollapse(dimensions: ContentRect, collapseWidth: number) {
+		const { onCollapse, direction } = this.properties;
 
 		if (direction === Direction.row) {
 			return false;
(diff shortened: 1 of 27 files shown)


tomdye:
@pottedmeat @bitpshr @aciccarello @rorticus could I get eyes on this please??

src/split-pane/index.ts
```diff
@@ -133,8 +131,8 @@ export class SplitPaneBase<P extends SplitPaneProperties = SplitPaneProperties>
 		return content ? [ content ] : [];
 	}
 
-	private _shouldCollapse(dimensions: ContentRect) {
-		const { onCollapse, direction, collapseWidth } = this.properties;
+	private _shouldCollapse(dimensions: ContentRect, collapseWidth: number) {
+		const { onCollapse, direction } = this.properties;
```
aciccarello: Why did this get added?

tomdye: because this is one of the examples where moving the properties suddenly broke the typings. Previously collapseWidth (although optional) wasn't causing any issues here. This function is only called from one place and is only called when `collapseWidth` exists, so it seemed sensible to do it like this. Otherwise id have had to do the check in two places.

sammenza:
I'm curious why are we exporting twice? Also doesn't exporting the class as default mean you can name the import anything you want?

tomdye:
@sammenza some people like to do `import thing from 'place';`, others like to `import { thing } from 'place';` This just gives that flexibility.

sammenza:
Huh alright 👍

PR #688, opened 2019-03-14
use reference diffProperty

### LINKED ISSUE #687: TextInput: Changing `leading` / `trailing` render function has no effect
**Bug** 

**Expected behavior:**

When changing the leading / trailing render function I would expect the rendered output to change.

**Actual behavior:**

There is no change.

tomdye:
**Type:** bug

**Description:**

This PR uses the `diffProperty` `reference` function to ensure that the `TextInput` re-renders when a new `leading` / `trailing` render function is passed.

This will cause the widget to re-render if the user passes a new `leading` / `trailing` function but will have the side effect that it will also re-render whenever the parent re-renders if the user is using an arrow function or a bind within their renderers.

The most efficient way to use these functions is with named functions / widget-level functions (making use of the autobind functionality).

`w('TextInput', { leading: this.renderLeading }`
or in the case of a changing leading render:
`w('TextInput', { leading: this.toggle ? this.renderA : this.renderZ }`

Resolves #687

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@dojo/widgets",
-  "version": "6.0.0-alpha.3",
+  "version": "6.0.0-pre",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


tomdye:
@nicknisi hopefully this should be everything you've ever dreamed of.

bitpshr:
CI is complaining about a few tests, not sure if it's related. This looks good otherwise.

tomdye:
@bitpshr those tests are unrelated. not sure why they've started failing

PR #697, opened 2019-03-22
Removing custom element decorators (mostly) 

rorticus:
**Type:** feature

**Description:**

Now that custom elements are added automagically, we can get rid of most usages of the custom element decorator.

**Note: You need @dojo/framework 6.0.0-pre to build this. Due to type changes in this version, the grid fails to build. Also note that the FocusMixin declares a `focus` property, which when read by `registerCustomElement` will cause an invocation exception (because it's trying to call the native focus code).**

Relates: https://github.com/dojo/webpack-contrib/issues/43


rorticus:
Note that CI isn't going to pass on this until we release @dojo/widgets (need custom element decorator changes from https://github.com/dojo/framework/pull/243)

PR #699, opened 2019-03-23
670 - Toolbar children: set horizontal and vertical layout

### LINKED ISSUE #670: Toolbar: Should apply flex layout to menu items
**Enhancement** 

It is currently down to the widget user to implement a layout for their links when using the toolbar. This means that they must manually add classes etc to layout their links differently between the header and the slide pane.

We should change the toolbar to automatically layout the toolbar children using `flex-direction: row` for the toolbar and `flex-direction: column` for the slide pane.

msssk:
**Type:** feature

* [x] Unit tests are updated in the PR

**Description:**
Toolbar: add CSS classes and rules to lay out children using `flex-direction: row` for the toolbar and `flex-direction: column` for the slide pane.

Resolves #670

### DIFF
--- a/src/theme/toolbar.m.css
+++ b/src/theme/toolbar.m.css
@@ -3,10 +3,8 @@
 	flex-direction: row;
 }
 
-.collapsed {
-	& .actions {
-		flex-direction: column;
-	}
+.collapsed .actions {
+	flex-direction: column;
 }
 
 .menuButton { }
--- a/src/theme/toolbar.m.css
+++ b/src/theme/toolbar.m.css
@@ -1,4 +1,8 @@
-.actions { }
+.actions {
+	display: flex;
+	flex-direction: row;
+}
+
 .collapsed { }
 .menuButton { }
 .root { }
--- a/src/theme/toolbar.m.css
+++ b/src/theme/toolbar.m.css
@@ -3,12 +3,12 @@
 	flex-direction: row;
 }
 
-.slidePaneActions {
-	display: flex;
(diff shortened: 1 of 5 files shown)


src/toolbar/index.ts
```diff
@@ -81,6 +81,10 @@ export class Toolbar extends I18nMixin(ThemedMixin(WidgetBase))<ToolbarPropertie
 			classes,
 			heading
 		} = this.properties;
+		const actions = v('div', {
+			classes: this.theme(this._collapsed ? css.slidePaneActions : css.actions),
```
tomdye: I think that we can always use `css.actions` here. We can differentiate between slide pane / toolbar placement via the `collapsed` class. This makes the dom / classes more predictable for the end user.

msssk: Thanks, this is definitely a better solution! Updated

src/theme/toolbar.m.css
```diff
@@ -1,5 +1,14 @@
-.actions { }
-.collapsed { }
+.actions {
+	display: flex;
+	flex-direction: row;
+}
+
+.collapsed {
+	& .actions {
```
tomdye: sorry to be a pain, but we're avoiding the use of nested selectors via the `&` operator. Could you change this to `.collapsed .actions` please.

msssk: Changed in bf161c0

tomdye:
Thanks @msssk

PR #701, opened 2019-03-25
Add ability to size SplitPane by its trailing pane

### LINKED ISSUE #698: Extending SplitPane: size trailing pane
**Enhancement** 

Extend SplitPane so that the size can apply to the trailing pane instead of the leading one. A sample use case: your site has navigation in the trailing (right) pane and you want the leading (left) pane to fill up the rest of the screen. The current implementation always "freezes" the left pane at `size` and the right pane takes up the rest of the screen.

schontz:
**Type:** feature

**Description:**

Add the ability to have `size` apply to the second pane in a `SplitPane`. This PR does it by adding a property named `sizeAppliesTo` that can be `0` (first pane, default) or `1` (second pane). Sample:

```ts
w(SplitPane, {
	key: 'secondSizeRow',
	direction: Direction.column,
	sizeAppliesTo: 1,
	onResize: (size: number) => {
		this.setState({ secondSizeRow: size });
	},
	size: this.state.secondSizeRow
})
```

Notes:
- If `sizeAppliesTo = 1`, the leading/trailing CSS classes are swapped (i.e. the second pane is considered the leading pane). If this could cause any issues, then new CSS classes should be added instead.
- I realize the name `sizeAppliesTo` may not be the best. Open to any suggestions.

Resolves #698

### DIFF
--- a/src/split-pane/README.md
+++ b/src/split-pane/README.md
@@ -83,3 +83,21 @@ w(SplitPane, {
 	})
 ])
 ```
+
+*SplitPane with size applied to the trailing pane*
+```typescript
+import SplitPane, { Direction } from '@dojo/widgets/split-pane';
+import { w } from '@dojo/framework/widget-core/d';
+
+w(SplitPane, {
+	key: 'sizeTrailing',
+	direction: Direction.column,
+	onResize: (size: number) => {
+		this.setState({ size: size });
+	},
+	size: this.state.size
+}, [
+	v('div', ['left: main content']),
+	v('div', ['right: sidebar'])
+])
+```
--- a/src/split-pane/example/index.ts
+++ b/src/split-pane/example/index.ts
@@ -184,37 +184,55 @@ export default class App extends WidgetBase {
 					size: this.state.minSize
 				})
 			]),
-			v('h3', ['Size applies to the second pane (column)']),
+			v('h3', ['Size applies to the trailing pane (column)']),
 			v('div', {
-				id: 'example-second',
+				id: 'example-trailing',
 				styles: containerStyles
 			}, [
 				w(SplitPane, {
-					key: 'secondSize',
+					key: 'trailingSize',
 					direction: Direction.column,
-					sizeAppliesTo: 1,
+					sizeTrailing: true,
 					onResize: (size: number) => {
 						size = size < 100 ? 100 : size;
-						this.setState({ secondSize: size });
+						this.setState({ trailingSize: size });
 					},
-					size: this.state.secondSize
+					size: this.state.trailingSize
 				})
 			]),
-			v('h3', ['Size applies to the second pane (row)']),
+			v('h3', ['Size applies to the trailing pane (row)']),
 			v('div', {
-				id: 'example-secondRow',
+				id: 'example-trailingRow',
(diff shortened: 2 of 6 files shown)


tomdye:
@schontz how about passing both sizes to the `onResize` callback instead? May I ask what your use-case is here? I don't follow the need for this change.

schontz:
I am trying to create three panes that are sized like so: [leftWidth, auto, rightWidth]. The current implmentation does not support this use case when nesting, rather I end up with something like: [leftWidth, middleWidth, auto].

Passing the value to `onResize` may help, but the size is still set via `size`, so there is no good way to store that.

schontz:
I should mention that this issue becomes apparent when you resize a browser window. Besides my use case, I assume others will want the determine which pane takes up the "auto" space, e.g. documentation with TOC in the right pane. Does that make more sense?

I just realized I didn't test any of this with the collapse logic yet, so this PR still needs work.

schontz:
@tomdye What do you think about those use cases? When I get some time I can continue working on the PR.

tomdye:
@schontz I took a quick look at this too and came up with a solution using the `flex-direction` css prop. Would mean that you'd have to pass the leading / trailing in the wrong order though which im not 100% about. What do you think? https://codesandbox.io/s/4zpl0n5kzx

Regarding your solution, looks like it will do the job too, I'd rather have the defining property named something like `sizeTrailing: boolean;` and defaulted to false though. I believe it would be clearer that way.

schontz:
`flex-direction` seems like the right solution. The alternative is to specify the CSS like so:

```css
.leadingFixed,
.reverse .trailingFixed { /* leading flex */ }
.trailingFixed,
.reverse .leadingFixed { /* trailing flex */ }
```

I think `sizeTrailing` is better than mine. Your sample uses `reverse`, which isn't bad either. Once we settle on that I can update my PR.

I may need some help getting the proper tests in place for it, as I'm not familiar with the testing suite.

tomdye:
if going with a solution similar to mine, i think `reverseChildren` makes it clearer perhaps as it implies the child order and explains why theyre the wrong way, however, your approach with `sizeTrailing` makes it clear exactly whats happening.

schontz:
Updated the PR with `sizeTrailing` and the `sizeTrailing` CSS so users are not expected to swap children. Once you do that with nesting it would become a nightmare.

schontz:
Not sure why the build failed. At first glance, it does not seem related to my changes:

[Line 2488](https://travis-ci.org/dojo/widgets/builds/513171759#L2488) says:
```
× chrome 73.0.3683.86 on Linux - SlidePane - a left-aligned slidepane should close when swiping from right to left (1.327s)
AssertionError: expected -250 to be close to -320 +/- 15
```

schontz:
I added a couple unit tests. Is there anything left to finish this off?

schontz:
I attempted to resolve conflicts, but https://github.com/dojo/widgets/pull/701/conflicts doesn't seem to load.

tomdye:
@schontz I think we may be reworking or potentially removing split pane for v7 so I am closing this for the time being. I'm sorry we were unable to land it.

PR #702, opened 2019-03-30
refactor slide-pane

### LINKED ISSUE #689: Investigate why slide pane functionals keep failing
Failing build: https://travis-ci.org/dojo/widgets/builds/505771674

Screenshot 2019-03-14 14 29 27

aciccarello:
**Type:** bug

**Description:**
This refactors the slide-pane to avoid a bug that was showing up (accurately) in our functional tests. I narrowed down the issue to some extra complexity in the transform implementation. The refactor includes the following changes:

- Only calculate transform styles one way
- Perform open/close change checking in a diff property rather than in the renderer
- Use a local state to define whether it is sliding in or out (still gets cleared on next render)
- Remove onAttach hook/state that hasn't been needed since a previous change which removed dom manipulation

Resolves #689

### DIFF
--- a/src/slide-pane/index.ts
+++ b/src/slide-pane/index.ts
@@ -280,6 +280,7 @@ export class SlidePane extends I18nMixin(ThemedMixin(WidgetBase))<SlidePanePrope
 			closeText = `${messages.close} ${title}`;
 		}
 
+		// This is a side-effect of rendering, it clears the slide styles for the next render.
 		this._slide = undefined;
 
 		return v('div', {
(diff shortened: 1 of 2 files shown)


aciccarello:
Helps if you run the unit tests too...

aciccarello:
@tomdye The unit tests should pass now. One thing I realized while testing is that there is a `transitionend` callback. I wonder if that should be used to remove the slide styles rather than the next render so the render function doesn't have an odd side-effect.

tomdye:
@aciccarello I think the `this._slide = undefined` call is fine. It was previously doing a `this._slide = false` anyway. I've merged this as is as I want to unblock the rest of the PRs. If you feel it's worth changing to `transitionEnd` please raise an issue to that effect but I think it may be overkill. Thanks for doing this work 👍

PR #707, opened 2019-04-25
Add control indicator dom to radio widget

nicknisi:
**Type:** feature

Fixes: #709 

**Description:**
Add some control indicator DOM, hidden by default, to the radio widget. Visually, this adds no change as the nodes are hidden in the default theme. However, the nodes exist as hooks to allow for more powerful theming of the radio widget.


nicknisi:
Alternatively, this DOM could be be rendered only when a specific property is passed to the widget, but it seemed like maybe a not something you'd want to have to remember to add to every `Radio` in your app to make it properly themeable.

tomdye:
Can you raise an issue on @dojo/themes to address this please?

src/radio/index.ts
```diff
@@ -164,7 +164,13 @@ export class Radio extends ThemedMixin(FocusMixin(WidgetBase))<RadioProperties>
 					ontouchstart: this._onTouchStart,
 					ontouchend: this._onTouchEnd,
 					ontouchcancel: this._onTouchCancel
-				})
+				}),
+				v('div', {
+					classes: this.theme(css.controlIndicator)
```
tomdye: what do you think about using `radioBackground`, `radioOuter`, `radioInner` class names?

tomdye:
Did you raise a widget issue for this too? Please put the reasons plus links to the material spec in an issue for this (there may already be a material radio issue) and link it to this PR and then we can get it merged. Thanks

nicknisi:
Linked to issue #709. Also, the relevant theme changes are available in dojo/themes#49.

PR #708, opened 2019-04-27
Update combo a11y

smhigley:
Hey, remember me? 😉

I've been working on combobox accessibility, so I updated the combo widget with a few things that should make it play nice with more screen reader/browser pairings.

**Type:** feature

**Description:**

Improves semantics surfaced by screen readers, particularly when using form navigation shortcuts and in announcing combobox role and expanded/collapsed state when landing on the input. The changes include:

- Using `role="presentation"` on intermediary divs to establish a direct relationship between the combobox and the input
- Remove `aria-controls` as it isn't well used or supported, and the functionality it provides would not be desired in this case, as focus should remain on the input
- Make the clear/trigger buttons hidden from a screen reader, so the input is the only control surfaced when navigating by form field or tabbing. It is already easy to clear and open the combo with a keyboard, so this seems like a better user experience.
- Add aria-autocomplete. This is making an assumption, since it's set to the correct value if the behavior is to filter options in some way based on the typed text.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1071,7 +1071,6 @@
       "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
       "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
       "dev": true,
-      "optional": true,
       "requires": {
         "kind-of": "^3.0.2",
         "longest": "^1.0.1",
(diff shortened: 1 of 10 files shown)


matt-gadd:
@smhigley hey stranger 👋 , this looks good to me and tests all pass. Thanks for the contribution as always, looking forward to more a11y updates in the future too maybe?! 🎉

tomdye:
Nice one @smhigley, thanks for this contribution. We plan to shake up `select` and `combobox` for the next release to improve the api etc. I'll be sure to keep everything in line with your `a11y` changes! Hope all's well!

smhigley:
@matt-gadd @tomdye dunno why I didn't see your comments initially, but thanks! I'll do my best to keep contributing a11y learnings back :). ❤️ 🐉

PR #719, opened 2019-05-03
Add prettier and husky hook

tomdye:
**Type:** feature

**Description:**

- Add's `prettier` and husky pre commit hook.
- initial prettier run
- future widgets to be written in `tsx`

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "prepublish": "grunt peerDepInstall",
-    "prettier": "prettier --write \"**/*.{ts,tsx}\"",
+    "prettier": "prettier --write \"src/**/*.{ts,tsx}\"",
     "test": "grunt test",
     "build-custom-elements": "shx rm -rf output && dojo build widget",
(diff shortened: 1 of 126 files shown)


package.json
```diff
@@ -46,9 +52,25 @@
...
+    "typescript": "~2.6.1",
+    "husky": "2.2.0",
+    "lint-staged": "8.1.6",
+    "prettier": "1.17.0"
   },
   "dependencies": {
     "tslib": "~1.8.1"
+  },
+  "lint-staged": {
+    "*.{ts,tsx}": [
```
aciccarello: should also lint css

tomdye: would require a different linter etc though right?

tomdye:
Only the `package.json` has been manually updated.

package.json
aciccarello: I think it picks it up automatically

msssk:
Do not like prettier. It completely fails to live up to its name.

sammenza:
Late to the party but this looks great! Super useful, thanks for setting up 🎉

PR #721, opened 2019-05-03
Add Snackbar

### LINKED ISSUE #713: Snackbar
Port across snack bar widget.

nicknisi:
**Type:** feature

**Description:**

Add a Snackbar widget

Properties:

- `type?: 'success' | 'error'` - whether the Snackbar should display with its success or error state
- `open?: boolean;` - Whether the Snackbar is open
- `message: string;` - The title/text to display within the Snackbar
- `actions?: DNode | DNode[]`- Buttons to render in the actions portion of the Snackbar

Example use:

```tsx
protected render() {
  return (
    <div id="example-autoclose">
        <h3>Auto close Snackbar</h3>
        <Button
            onClick={() => {
                this.open = true;
                setTimeout(() => {
                    this.open = false;
                }, 5000);
            }}
        >
            Show Snackbar
        </Button>
        <Snackbar
            type="success"
            open={this.open}
            actions={<Button onClick={() => this._dismiss()}>Dismiss</Button>}
            message="Test Snackbar Error"
        />
    </div>
  );
}
```

**Screenshot**
image

Resolves #713

### DIFF
--- a/src/snackbar/index.tsx
+++ b/src/snackbar/index.tsx
@@ -1,52 +1,40 @@
 import { WidgetBase } from '@dojo/framework/widget-core/WidgetBase';
+import { DNode } from '@dojo/framework/widget-core/interfaces';
 import { tsx } from '@dojo/framework/widget-core/tsx';
 import I18nMixin from '@dojo/framework/widget-core/mixins/I18n';
 import Button from '../button/index';
 import * as css from '../theme/snackbar.m.css';
-import bundle from './nls/Snackbar';
+
+export type SnackbarType = 'success' | 'error';
 
 export interface SnackbarProperties {
-	success?: boolean;
 	open: boolean;
-	title: string;
-	onDismiss?: () => void;
+	message: string;
+	actions?: DNode | DNode[];
+	type?: SnackbarType;
 }
 
 export class Snackbar extends I18nMixin(WidgetBase)<SnackbarProperties> {
-	private _onDismiss() {
-		const { onDismiss } = this.properties;
-		onDismiss && onDismiss();
-	}
-
 	protected render() {
-		const { messages } = this.localizeBundle(bundle);
-		const { success, open, title } = this.properties;
+		const { type, open, message, actions = null } = this.properties;
 		const classes = [css.root];
 
 		if (open) {
 			classes.push(css.open);
 		}
 
-		if (success === true) {
-			classes.push(css.success);
-		} else if (success === false) {
-			classes.push(css.error);
+		if (type) {
+			classes.push(css[type]);
 		}
 
 		return (
 			<div key="root" classes={classes}>
 				<div key="content" classes={css.content}>
 					<div key="label" classes={css.label} role="status" aria-live="polite">
-						{title}
+						{message}
 					</div>
 					<div key="actions" classes={css.actions}>
-						<Button
-							name="snackbar-dismiss"
-							key="dismiss"
-							onClick={() => this._onDismiss()}
-						>
-							{messages.dismiss}
-						</Button>
+						{actions}
 					</div>
 				</div>
 			</div>
--- a/src/snackbar/index.tsx
+++ b/src/snackbar/index.tsx
@@ -5,12 +5,6 @@ import Button from '../button';
 import * as css from '../theme/snackbar.m.css';
 import bundle from './nls/Snackbar';
 
-export interface InfoMessage<S = string> {
-	success?: boolean;
-	title: S;
-	id?: number;
-}
-
 export interface SnackbarProperties {
 	success?: boolean;
 	open?: boolean;
--- a/src/snackbar/index.tsx
+++ b/src/snackbar/index.tsx
@@ -23,33 +23,27 @@ export interface SnackbarProperties {
 export class Snackbar extends ThemedMixin(WidgetBase)<SnackbarProperties> {
 	protected render(): DNode {
 		const { type, open, leading, stacked, message, actionsRenderer } = this.properties;
-		const classes = [css.root];
-
-		if (open) {
-			classes.push(css.open);
-		}
-
-		if (type) {
-			classes.push(css[type]);
-		}
-
-		if (leading) {
-			classes.push(css.leading);
-		}
-
-		if (stacked) {
-			classes.push(css.stacked);
-		}
 
 		return (
-			<div key="root" classes={classes}>
+			<div
+				key="root"
+				classes={[
+					css.root,
+					open ? css.open : null,
+					type ? css[type] : null,
+					leading ? css.leading : null,
+					stacked ? css.stacked : null
+				]}
+			>
 				<div key="content" classes={css.content}>
 					<div key="label" classes={css.label} role="status" aria-live="polite">
 						{message}
 					</div>
-					<div key="actions" classes={css.actions}>
-						{actionsRenderer && actionsRenderer()}
-					</div>
+					{actionsRenderer && (
+						<div key="actions" classes={css.actions}>
+							{actionsRenderer()}
+						</div>
+					)}
 				</div>
 			</div>
 		);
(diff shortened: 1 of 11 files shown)


src/snackbar/index.tsx
```diff
@@ -0,0 +1,51 @@
+import { WidgetBase } from '@dojo/framework/widget-core/WidgetBase';
+import { tsx } from '@dojo/framework/widget-core/tsx';
+import I18nMixin from '@dojo/framework/widget-core/mixins/I18n';
+import Button from '../button/index';
+import * as css from '../theme/snackbar.m.css';
+import bundle from './nls/Snackbar';
+
+export interface SnackbarProperties {
+	success?: boolean;
+	open?: boolean;
```
matt-gadd: this doesn't really make sense as an optional property does it?

src/snackbar/index.tsx
```diff
@@ -0,0 +1,51 @@
+import { WidgetBase } from '@dojo/framework/widget-core/WidgetBase';
+import { tsx } from '@dojo/framework/widget-core/tsx';
+import I18nMixin from '@dojo/framework/widget-core/mixins/I18n';
+import Button from '../button/index';
+import * as css from '../theme/snackbar.m.css';
+import bundle from './nls/Snackbar';
+
+export interface SnackbarProperties {
+	success?: boolean;
```
matt-gadd: are there any more plans for different states in the future? would this make sense to be a `type` property and string literal types rather than a boolean?

src/snackbar/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+	open?: boolean;
+	title: string;
+	onDismiss?: () => void;
```
matt-gadd: are we planning on implementing a generic action for the snackbar too rather than just on dismiss? and is it possible to hide the `dismiss` message like in the material guide too?

src/snackbar/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export class Snackbar extends I18nMixin(WidgetBase)<SnackbarProperties> {
+	private _onDismiss() {
+		const { onDismiss } = this.properties;
+		onDismiss && onDismiss();
+	}
+	protected render() {
+		const { messages } = this.localizeBundle(bundle) as { messages: { [key: string]: string } };
```
matt-gadd: how come this needed manually typing?

src/snackbar/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+		const { success, open, title } = this.properties;
+		return (
+			<div
+				key="root"
+				classes={[
+					css.root,
+					open ? css.open : null,
+					success === true ? css.success : success === false ? css.error : null
```
matt-gadd: this is a tough read 🙈

src/snackbar/index.tsx
nicknisi: Good point! I changed this to be an optional `type?: 'success' | 'error'` property, so it can be extended later.

src/snackbar/index.tsx
nicknisi: After thinking about this more and realizing that in material there can be more than one action, I replaced this with an `actions?: DNode | Dnode[]` property, allowing the user to control what actions are there, completely.

Examples:

```tsx
<Snackbar
	open={this._showSnackbar}
	message="Test Snackbar"
	actions={<Button onClick={() => this._showSnackbar = false}>Dismiss</Button>}
/>
```

and 

```tsx
<Snackbar
	open={this._showSnackbar}
	type="error"
	message="Task failed."
	actions={[
		<Button onClick={() => retry()}>Retry</Button>,
		<Button onClick={() => this._showSnackbar = false}>Dismiss</Button>
	]}
/>
```

nicknisi: @matt-gadd Thinking about this more... should it maybe be a property like `actionsRenderer?: () => DNode | DNode[]`?

src/snackbar/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		return (
+			<div key="root" classes={classes}>
+				<div key="content" classes={css.content}>
+					<div key="label" classes={css.label} role="status" aria-live="polite">
+						{message}
+					</div>
+					<div key="actions" classes={css.actions}>
+						{actions}
```
tomdye: can you make this an `actionsRenderer` function instead please? That makes it more consistent with others.

src/theme/snackbar.m.css
```diff
@@ -0,0 +1,54 @@
+@import '../common/styles/variables.css';
+
+.root {
```
tomdye: not sure if this should all be here or in the dojo theme?

nicknisi: @tomdye Good question. I added the minimum structure and positioning to this CSS, just so that it uses flex box and displays at the bottom of the window. Should this not be here?

src/snackbar/index.tsx
nicknisi: Done. I wondered about that above!

src/snackbar/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+	open: boolean;
+	message: string;
+	actionsRenderer?: () => RenderResult;
+	type?: 'success' | 'error';
+}
+
+export class Snackbar extends WidgetBase<SnackbarProperties> {
+	protected render(): DNode {
+		const { type, open, message, actionsRenderer } = this.properties;
+		const classes = [css.root];
```
KaneFreeman: Need to make Widget themable

src/snackbar/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+@theme(css)
+@customElement<SnackbarProperties>({
+	tag: 'dojo-snackbar',
+	properties: ['actionsRenderer', 'leading', 'open', 'stacked'],
+	attributes: ['message', 'type']
+})
+export class Snackbar extends ThemedMixin(WidgetBase)<SnackbarProperties> {
+	protected render(): DNode {
+		const { type, open, leading, stacked, message, actionsRenderer } = this.properties;
+		const classes = [css.root];
```
tomdye: Is there a reason you didn't use ternary for adding conditional classes to the root here like we do elsewhere.

src/snackbar/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+			classes.push(css.stacked);
+		}
+
+		return (
+			<div key="root" classes={classes}>
+				<div key="content" classes={css.content}>
+					<div key="label" classes={css.label} role="status" aria-live="polite">
+						{message}
+					</div>
+					<div key="actions" classes={css.actions}>
```
tomdye: I'm not sure the `actions` section should be rendered at all if there is no `actionsRenderer` property passed

src/snackbar/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+	tag: 'dojo-snackbar',
+	properties: ['actionsRenderer', 'leading', 'open', 'stacked'],
+	attributes: ['message', 'type']
+})
+export class Snackbar extends ThemedMixin(WidgetBase)<SnackbarProperties> {
+	protected render(): DNode {
+		const { type, open, leading, stacked, message, actionsRenderer } = this.properties;
+		const classes = [css.root];
+
+		if (open) {
```
tomdye: Please see previous comment re. ternary for these classes as we have done elsewhere.

src/snackbar/index.tsx
nicknisi: I changed this to be consistent with what we've done elsewhere. I originally did it this way because I'm not a fan of testing the placement of `null`.

```tsx
const successTemplate = template.setProperty('@root', 'classes', [
	css.root,
	css.open,
	null,
	null,
	css.stacked
]);
```

PR #723, opened 2019-05-04
[ISSUE-677] Helpertext on combobox

### LINKED ISSUE #677: Add `helperText` to appropriate input widgets
**Enhancement** 

As part of validate text-input we have added a `helperText` property. We should add this to all other appropriate input widgets.

Widgets requiring helpertext:

- [ ] timepicker
- [ ] slider
- [ ] range slider

sammenza:
**Type:** feature

**Description:** 
- Adds helpertext to combobox widget
- Adds onValidate property to the combobox widget
- Updates valid property to use most recent type interface `{valid: boolean, message: string}` instead of a boolean.

Contributes to #677

### DIFF
--- a/src/combobox/example/index.ts
+++ b/src/combobox/example/index.ts
@@ -65,7 +65,10 @@ export default class App extends WidgetBase {
 	private _value7 = '';
 	private _value8 = '';
 	private _value9 = '';
-	private _valid = true;
+	private _valid: { valid: boolean | undefined; message: string } = {
+		valid: true,
+		message: 'Please enter value of state'
+	};
 
 	onChange(value: string, key?: string) {
 		if (!key) {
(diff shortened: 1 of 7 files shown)


sammenza:
Does this functional tests? I wasn't sure

src/combobox/example/index.ts
```diff
@@ -180,14 +179,18 @@ export default class App extends WidgetBase {
...
 					},
 					getResultLabel: (result: any) => result.value,
 					onRequestResults,
 					results: this._results,
 					value: this._value9,
-					invalid: this._invalid,
+					valid: { valid: this._valid, message: 'Please enter value of state' },
+					helperText: 'helper text',
+					onValidate: (valid: boolean | undefined, message: string) => {
+						console.log('onValidate called', valid, message);
```
tomdye: this callback should be used to set `valid` back on the widget

src/combobox/index.ts
```diff
@@ -290,32 +294,54 @@ export class ComboBox extends I18nMixin(ThemedMixin(FocusMixin(WidgetBase)))<Com
...
+		const {
+			clearable
+		} = this.properties;
+
+		const {
+			valid
+		} = this.validity;
 		const focus = this.meta(Focus).get('root');
 
 		return [
```
tomdye: can you just move this into the main `render` function please?

src/combobox/example/index.ts
sammenza: Why would I want to set the returned valid value to `this._valid`? Doesn't onChange set `this._valid`?

src/combobox/index.ts
sammenza: Done

src/combobox/example/index.ts
sammenza: And if the `onValidate` function is something that will update the valid boolean based on the most up to date version of the dom shouldn't that be abstracted away so the user doesn't have to think about that?

tomdye: it's a controlled component, the onValidate is a callback to inform the parent that the valid status has changed. It's upto the parent to set `valid` back on the widget so that it can display the new status.

tomdye:
Looks good @sammenza , but I think you still need to remove the `labelAfter` property. 
It doesn't make sense / work with helperText.

sammenza:
Alright cool, `labelAfter` has been removed

PR #727, opened 2019-05-07
Added command to host widget examples locally

### LINKED ISSUE #718: Validate instructions for running widget examples
We should validate that the existing instructions for running widget examples are correct.
I believe the correct way to run these is via a http-server running from the project root. The npm module `http-server` is perfect for such a task.

sammenza:
**Type:** feature

**Description:** Widget example app had no included way to host the project locally. Instead the project asked developers to load the `_build/common/example/index.html` file in the web browser. This caused the widget examples to not load icons on firefox due to CORs issues as the request is not over http.

Resolves #718

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "url": "https://github.com/dojo/widgets.git"
   },
   "scripts": {
+    "start": "serve",
     "prepublish": "grunt peerDepInstall",
     "prettier": "prettier --write \"src/**/*.{ts,tsx}\"",
     "test": "grunt test",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -46,16 +47,17 @@
     "grunt": "^1.0.1",
     "grunt-dojo2": "~4.0.0",
     "grunt-tslint": "^5.0.1",
+    "husky": "2.2.0",
```
tomdye: did you mean to add husky etc here?

sammenza:
Good point, I'll add that that will run both watch and serve 👍

PR #730, opened 2019-05-10
Add size property to text-input

### LINKED ISSUE #729: Text Input min-width
Text-input appears to have a min-width that you can't control via css due to the browser default for `size`. This is particularly problematic with flex layouts where text-inputs currently do not go smaller than the browser default for `size`.

 We should add it as a widget property and default it to `1`. This appears to fix the issue.

stack overflow link:

https://stackoverflow.com/questions/29470676/why-doesnt-the-input-element-respect-min-width

tomdye:
**Type:** bug

**Description:**

Adds `size` property to `text-input`

Resolves #729


tomdye:
@nicknisi defaulting the `size` to `1` means that it can shrink down to the min size specified by the css.

tomdye:
Looks like this can be fixed with a combination of `appearance: 1` and `width: 100%` on the `input` instead. I believe this is already in place in the dojo theme.

PR #733, opened 2019-05-21
[WIP] Switch onChange and onInput to onValue

grayrest:
**Type:** feature

* [X] There is a related issue ( #724 )
* [X] Unit or Functional tests are included in the PR

**Description:**

This is an in-progress branch for collapsing `onInput`, `onChange`, and the custom-widget specific update events into a single `onValue` with the goal of simplifying state update code and allowing better generic handling at the form level.

Incomplete:

* I was working through the components incrementally and haven't yet changed the definitions in `interfaces.d.ts` so the components have individual `onValue` definitions that must be removed when that happens.
* Select has a more complicated interaction here than other widgets. As a result, I was hoping to keep `onChange` as an escape hatch for handling an entire option object and building `onValue` alongside it. I'm missing one of the interactions and as a result a unit test is failing.
* What's here is mostly a mechanical translation from `onChange` and friends. I think there should be some additional API changes to further the goal of generic form handling. In particular I think the first argument should *always* be the value and the second *always* the name. This means that the range slider's onValue callback should take a `[low, high]` array instead of two separate arguments and changes to the radio and checkbox versions. The last two are most problematic because they have both `checked` and `value`. We can pass `checked` as the third argument easily enough but I'm not sure if the first argument should be `undefined` if not checked or the empty string.

Resolves #???

### DIFF
--- a/src/button/index.ts
+++ b/src/button/index.ts
@@ -55,7 +55,6 @@ export interface ButtonProperties
 	attributes: ['widgetId', 'name', 'type', 'value'],
 	events: [
 		'onBlur',
-		'onChange',
 		'onClick',
 		'onFocus',
 		'onInput',
--- a/src/calendar/README.md
+++ b/src/calendar/README.md
@@ -36,7 +36,7 @@ w(Calendar, {
 	year: this.state.year,
 	onMonthChange: (month: number) => { this.setState({ 'month': month }); },
 	onYearChange: (year: number) => { this.setState({ 'year': year }); },
-	onDateSelect: (date: Date) => {
+	onValue: (date: Date) => {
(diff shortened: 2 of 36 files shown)


tomdye:
Do you think we _need_ both `value` and `checked` returned in the checkbox / radio callbacks? I don't believe this is how they'll be used within dojo applications. We should probably just return the checked boolean and implement `radioGroup` / `checkboxGroup` to handle the rest. What do you think?

grayrest:
I don't think people will need both if the `Group` versions are available so that works for me.

tomdye:
Closing in favour of https://github.com/dojo/widgets/pull/785

PR #748, opened 2019-06-10
Add script to run functionals

tomdye:
`npm run test:functional`


sammenza:
Can you update the readme as well? `grunt test:local` seems to be having issues so it might be best to instruct people to use this new script instead.

PR #751, opened 2019-06-13
Changed onclick to onmousedown on listboxoption

### LINKED ISSUE #704: Time picker drop down does not close
**Bug**

When the property `openOnFocus` is true on the timePicker widget the drop down will not close if option is selected with your cursor.

Package Version: 6.0.0

**Screenshot**

dojo-time-picker-bug

**Expected behavior:**

Drop down should close when option is selected with the cursor.

**Actual behavior:**

Drop down does not close when option is selected with the cursor.

### LINKED ISSUE #750: Issue
**Bug**
When the height property is set on the ListboxOption widget (on the `.option` class in `listbox.m.css`) this causes the onclick event to stop working in IE 11. 

Package Version: 6.0.0

**Expected behavior:**

When clicking on a select option on the custom select widget, the value of select should change.

**Actual behavior:**

When clicking on a select option on the custom select widget, the value of select does not change.

**Screenshot**

select-ie-bug

sammenza:
**Type:** Bug
This PR resolves two bugs:

- When the height property is set on the ListboxOption widget (on the .option class in listbox.m.css) this causes the onclick event to stop working in IE 11.

- In the example of the timePicker widget, the drop down will not close if option is selected with your cursor.

**Description:**

Resolves #750
Resolves #704

### DIFF
--- a/src/listbox/ListboxOption.ts
+++ b/src/listbox/ListboxOption.ts
@@ -19,7 +19,7 @@ export interface ListboxOptionProperties extends ThemedProperties {
 
 @theme(css)
 export class ListboxOption extends ThemedMixin(WidgetBase)<ListboxOptionProperties> {
-	private _onClick(event: PointerEvent) {
+	private _onClick(event: MouseEvent) {
 		event.stopPropagation();
 		const { index, key, option, onClick } = this.properties;
 		onClick && onClick(option, index, key);
--- a/src/listbox/ListboxOption.ts
+++ b/src/listbox/ListboxOption.ts
@@ -36,7 +36,7 @@ export class ListboxOption extends ThemedMixin(WidgetBase)<ListboxOptionProperti
 				classes: this.theme(css),
 				id,
 				role: 'option',
-				onpointerdown: this._onClick
+				onmousedown: this._onClick
 			},
 			[label]
 		);
--- a/src/listbox/ListboxOption.ts
+++ b/src/listbox/ListboxOption.ts
@@ -36,7 +36,7 @@ export class ListboxOption extends ThemedMixin(WidgetBase)<ListboxOptionProperti
 				classes: this.theme(css),
 				id,
 				role: 'option',
-				onclick: this._onClick
+				onmousedown: this._onClick
 			},
 			[label]
 		);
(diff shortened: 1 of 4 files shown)


src/listbox/ListboxOption.ts
```diff
@@ -14,15 +14,15 @@ export interface ListboxOptionProperties extends ThemedProperties {
 	label: DNode;
 	option: any;
 	selected?: boolean;
-	onClick?(option: any, index: number, key?: string | number): void;
+	onMouseDown?(option: any, index: number, key?: string | number): void;
```
agubler: This change is going to need to be back-ported to version 5, so changing the event from `onclick` to `onmousedown` should really just be an internal detail - We can leave all the other references to `onClick` and leave the API as `onClick`.

sammenza: Cool got it, I'll just include this one line change then

src/listbox/ListboxOption.ts
```diff
@@ -14,15 +14,15 @@ export interface ListboxOptionProperties extends ThemedProperties {
...
 
 @theme(css)
 export class ListboxOption extends ThemedMixin(WidgetBase)<ListboxOptionProperties> {
-	private _onClick(event: MouseEvent) {
+	private _onMouseDown(event: MouseEvent) {
 		event.stopPropagation();
-		const { index, key, option, onClick } = this.properties;
-		onClick && onClick(option, index, key);
+		const { index, key, option, onMouseDown } = this.properties;
+		onMouseDown && onMouseDown(option, index, key);
```
tomdye: could this be pointer down?

sammenza: Let me see if that works

sammenza: `onpointerdown` works fine, code is updated 👍

src/time-picker/tests/functional/TimePicker.ts
```diff
@@ -158,12 +159,6 @@ registerSuite('TimePicker', {
 					'The input value should not contain the disabled value.'
 				);
 			})
-			.end()
```
sammenza: The failing test was taken out as it does not test for the right thing. When a disabled menu item is clicked the correct behavior is that the menu should close.

PR #756, opened 2019-06-24
Updating old ARIA describedBy examples

rorticus:
**Type:** bug

**Description:**

Updating old examples that show setting the ARIA `describedBy` with just a `describedBy` property, which is gone now. The new way is to set the `aria` property like `aria: { describedBy: 'instructions' }`.

Resolves #775


rorticus:
Should be! I searched all the `.md` files for `describedBy` and checked each hit.

PR #759, opened 2019-07-09
Migrates to using `@dojo/cli-build-widget` and `@dojo/scripts`

agubler:
**Type:** feature

**Description:**

Migrates the project to use the Dojo cli command for building both the widget library and custom elements.

The release and testing is provided by `@dojo/scripts`


package.json
```diff
@@ -45,14 +55,15 @@
...
     "execa": "^0.9.0",
-    "grunt": "^1.0.1",
-    "grunt-dojo2": "~4.0.0",
-    "grunt-tslint": "^5.0.1",
+    "glob": "7.1.4",
     "husky": "2.2.0",
     "jsdom": "^9.5.0",
     "lint-staged": "8.1.6",
-    "load-grunt-tasks": "^3.5.2",
+    "npm-run-all": "^4.1.5",
```
mwistrand: Is there any reason this shouldn't be pinned?

agubler: missed it! thanks!

intern.json
```diff
@@ -10,16 +10,16 @@
...
+		"./dist/dev/src/common/tests/functional/all.js"
 	],
 
 	"browser": {
 		"loader": {
-			"script": "./node_modules/grunt-dojo2/lib/intern/internLoader.js",
+			"script": "./node_modules/@dojo/scripts/intern/loader.js",
 			"options": {
 				"packages": [
 					{ "name": "src", "location": "_build" }
```
rorticus: Should this `_build` be `dist/dev/src` as well?

src/grid/index.ts
```diff
@@ -8,6 +8,7 @@ import { reference } from '@dojo/framework/core/diff';
 import { Store } from '@dojo/framework/stores/Store';
 import Dimensions from '@dojo/framework/core/meta/Dimensions';
 import Resize from '@dojo/framework/core/meta/Resize';
+import './utils';
```
rorticus: Is this imported for side-effects, or was this not intentional?

intern.json
agubler: probably... I don't know what this does tho lol

src/grid/index.ts
agubler: yup this needs to be imported for the module to be included in the library built.

intern.json
mwistrand: Just `dist/dev`, although I don't see any place we're actually using this `browser` config.

PR #765, opened 2019-07-19
Fixed problematic functional tests

sammenza:
**Type:** bug

**Description:**
Functional tests have been erroring out and causing disruption to development

### DIFF
--- a/src/tab-controller/tests/functional/TabController.ts
+++ b/src/tab-controller/tests/functional/TabController.ts
@@ -5,11 +5,8 @@ import { Remote } from 'intern/lib/executors/Node';
 import { services } from '@theintern/a11y';
 import * as css from '../../../theme/tab-controller.m.css';
 import { uuid } from '@dojo/framework/core/util';
-import pollUntilTruthy from '@theintern/leadfoot/helpers/pollUntilTruthy';
 
 const axe = services.axe;
-const CLICK_WAIT_TIMEOUT = 1000;
-const POLL_INTERVAL = 20;
 
 function getPage(remote: Remote) {
 	return remote
@@ -42,22 +39,7 @@ registerSuite('TabController', {
 			})
 			.click()
 			.end()
-			.then(
-				pollUntilTruthy(
-					function(rootClass, activeTabButtonClass) {
-						const activeTab = document.querySelector(
-							`.${rootClass} .${activeTabButtonClass}`
-						);
-						return document.activeElement === activeTab;
-					},
-					[css.root, css.activeTabButton],
-					CLICK_WAIT_TIMEOUT,
-					POLL_INTERVAL
-				)
-			)
-			.then((isEqual) => {
-				assert.isTrue(isEqual);
-			})
+			.findByCssSelector(`.${css.root} .${css.activeTabButton}`)
 			.end()
 			.end();
 	},
--- a/src/tab-controller/tests/functional/TabController.ts
+++ b/src/tab-controller/tests/functional/TabController.ts
@@ -7,6 +7,7 @@ import * as css from '../../../theme/tab-controller.m.css';
 import { uuid } from '@dojo/framework/core/util';
 
 const axe = services.axe;
+const DELAY = 1000;
 
 function getPage(remote: Remote) {
 	return remote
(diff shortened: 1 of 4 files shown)


src/tab-controller/tests/functional/TabController.ts
```diff
@@ -7,6 +7,7 @@ import * as css from '../../../theme/tab-controller.m.css';
 import { uuid } from '@dojo/framework/core/util';
 
 const axe = services.axe;
+const DELAY = 1000;
```
sammenza: It anyone knows of an async way of waiting for `click` and `type` to finish before continuing please let me know. I would rather not use the synchronous `sleep` method.

src/tab-controller/tests/functional/TabController.ts
```diff
@@ -27,9 +28,9 @@ registerSuite('TabController', {
 	},
 	'tabs should be changable'() {
 		return getPage(this.remote)
-			.findByCssSelector(`.${css.root}`)
-			.findByCssSelector(`.${css.tabButton}:last-child`)
-			.getProperty('className')
+			.find('css selector', `.${css.root}`)
```
msssk: What's the reasoning behind switching to `find`?

src/tab-controller/tests/functional/TabController.ts
msssk: The general idea is to identify something that changes as a result of `click` or `type` and then use [`find`](https://theintern.io/docs.html#Leadfoot/2/api/Session/find), [`executeAsync`](https://theintern.io/docs.html#Leadfoot/2/api/Session/executeasync), or [`pollUntil`](https://theintern.io/docs.html#Leadfoot/2/api/helpers%2FpollUntil) to poll for the condition.

I have some work in [my branch](https://github.com/dojo/widgets/compare/master...msssk:test-async-improvements), but the 'Validated TextInput should update style based on validity' test is still failing. It does successfully demonstrate using `pollUntilTruthy` which replaces a fixed 250ms sleep with a poll that will finish as soon as the DOM updates (or potentially wait as long as 5s for a severely resource-constrained test environment to respond to the action).

sammenza: Hmm ok I see what your doing here, let me check it out for my fixes here

src/tab-controller/tests/functional/TabController.ts
sammenza: I'm replacing `find` with `findByCssSelector` as you said before that the `find` function is better as it will keep polling until the it reaches timeout.

src/text-input/tests/functional/TextInput.ts
```diff
@@ -120,22 +123,36 @@ registerSuite('TextInput', {
...
 			.end()
-			.sleep(10)
-			.getProperty('className')
-			.then((className: string) => {
-				assert.notInclude(className, css.valid);
-				assert.include(className, css.invalid);
+			.then(
+				pollUntilTruthy(
+					function(validClass) {
+						return !!document.querySelector(`#example-validated .${validClass}`);
```
msssk: Is the goal here to confirm that a previously valid input is no longer valid?

src/text-input/tests/functional/TextInput.ts
```diff
@@ -120,22 +123,36 @@ registerSuite('TextInput', {
...
+					},
+					[css.valid],
+					CLICK_WAIT_TIMEOUT,
+					POLL_INTERVAL
+				)
+			)
+			.then((isEqual) => {
```
msssk: Looking at line 129 above, if you had to name that function, what would be a good name? I think `isEqual` on this line could use a better name, perhaps something like `is<Line29FunctionName>`

src/text-input/tests/functional/TextInput.ts
sammenza: Yup that's right, the goal here is to determine when you type "foo" does the widget add a `valid` class, and when you add invalid text "foofoobar" does the widget add the invalid class.

src/text-input/tests/functional/TextInput.ts
sammenza: Maybe `doesValidClassExist`?

src/text-area/tests/functional/Textarea.ts
```diff
@@ -146,15 +147,19 @@ registerSuite('Textarea', {
 				.end()
 				.end()
 				// focus another input
-				.findByCssSelector(`#example-t1 .${css.root} .${css.input}`)
+				.findByCssSelector(`#example-t3 .${css.root} .${css.input}`)
 				.click()
+				.type(space)
```
msssk: I think it's fine to just use `keys.SPACE` directly here

PR #767, opened 2019-07-29
Basic validation for select widget

KaneFreeman:
**Type:** feature

**Description:**

Add basic validation to the select widget, along with an `onValidate` event to match `TextInput`.

**Use case**: checking the validity of a whole form. Currently validation logic for select needs to be written each time the widget is used (or a wrapper widget needs to be written).


src/select/index.ts
```diff
@@ -149,6 +157,40 @@ export class Select extends ThemedMixin(FocusMixin(WidgetBase))<SelectProperties
...
+		const {
+			_state: state,
+			properties: { onValidate, value, required }
+		} = this;
+
+		if (!onValidate) {
+			return;
+		}
+
+		if (
```
tomdye: if this widget has no value, but it is required then should it not report as `invalid` once it's been interacted with?

KaneFreeman: This has been addressed. I have also matched the logic to be closer to TextInput and TextArea.

agubler:
@tomdye is this outdated now that we are working on the replacement for `Select`?

tomdye:
I will shortly be re-writing Select to use the new menu / popup functionality so will bake this into the new code rather than landing this PR. Thanks

PR #768, opened 2019-07-29
Add option to hide helper text

KaneFreeman:
**Type:** bug / feature

**Description:**

Add `helperTextHidden` option (similar to `labelHidden`) that prevents the rendering of the `HelperText` when `true`.

**Use Case**: When application styling does not allow for the extra space needed to display the helper text. Currently the only option to hide it is to override it with some hacky css.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ yarn.lock
 deploy_key
 deploy_key.pub
 package.json.bk
+/.vscode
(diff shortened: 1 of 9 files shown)


tomdye:
If the sizing doesn't allow for showing the helper text, then could you simply pass `helperText: undefined` instead. It seems counter-intuitive to me that you'd pass both a string for the helperText and a boolean to then hide it. The only reason we do this for `label` is for accessibility reasons. I don't believe that there are any such concerns for this with helpertext as it's not a recognised / established accessibility tag. 

I don't think this is a change we should be making.

PR #769, opened 2019-07-29
Add basic textarea validation and pass valid to helper text

KaneFreeman:
**Type:** feature

**Description:**

Add basic validation to the textarea widget, along with an `onValidate` event to match `TextInput`.

**Use case**: checking the validity of a whole form. Currently validation logic for textarea needs to be written each time the widget is used (or a wrapper widget needs to be written).

### DIFF
--- a/src/text-area/example/index.ts
+++ b/src/text-area/example/index.ts
@@ -5,7 +5,11 @@ import Textarea from '../../text-area/index';
 export default class App extends WidgetBase {
 	private _value1: string | undefined;
 	private _value2: string | undefined;
-	private _invalid: boolean | undefined;
+	private _valid1: undefined | boolean | { valid?: boolean; message?: string };
+	private _customValid: undefined | boolean | { valid?: boolean; message?: string };
+	private _customValue: string | undefined;
+
+	// private _valid1: undefined | { valid: boolean, message?:string };
 
 	render() {
 		return v('div', [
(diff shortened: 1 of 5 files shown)


src/text-area/index.ts
```diff
@@ -47,6 +53,7 @@ export interface TextareaProperties
 	placeholder?: string;
 	value?: string;
 	onClick?(value: string): void;
+	onValidate?: (valid: boolean | undefined) => void;
```
tomdye: does this not need `valid?: { valid?: boolean; message?: string } | boolean;` plus logic for displaying it?

KaneFreeman: Yup, you are right. Will add that.

tomdye:
@KaneFreeman i've picked this up to make the required changes, will push to your branch shortly

src/text-area/example/index.ts
```diff
@@ -5,7 +5,11 @@ import Textarea from '../../text-area/index';
 export default class App extends WidgetBase {
 	private _value1: string | undefined;
 	private _value2: string | undefined;
-	private _invalid: boolean | undefined;
+	private _valid1: undefined | boolean | { valid?: boolean; message?: string };
+	private _customValid: undefined | boolean | { valid?: boolean; message?: string };
+	private _customValue: string | undefined;
+
+	// private _valid1: undefined | { valid: boolean, message?:string };
```
ericos: Remove comment.

PR #770, opened 2019-07-29
Mixin/merge passed in classes instead of overwriting

### LINKED ISSUE #762: Classes property doesn't work for Raised/Outline buttons
**Bug** 

The raised/outline buttons override the default button styles using a `classes` property, but also mix their properties into the button. If `classes` is passed to the raised/outline button, it'll override this property and lose the raised/outline styling. We could deeply mixin a passed classes property to avoid this scenario.

Package Version: 6.0.0-pre

KaneFreeman:
**Type:** bug

**Description:**

Merge in passed in classes property with classes defined in `OutlinedButton` and `RaisedButton` instead of overwriting.

Resolves #762


KaneFreeman:
> Is this something that could be a framework enhancement rather than a widget specific implementation?

I believe that would be a call for @agubler and @matt-gadd, but I would assume this is a common enough issue to maybe warrant the utility being in framework.

tomdye:
I don't believe that we need this anymore now that we have moved over to the theme.compose pattern

PR #777, opened 2019-08-21
Removes sleep from functional tests

### LINKED ISSUE #749: Switch test from using `sleep` to `find`
**Enhancement** 
Currently we use `sleep` in our functional tests for widgets. However `sleep` makes the runner wait a fixed time and if the element isn't there in time the test fails. This creates brittle tests and is also bad for pipeline run time.

A more efficient way would be to use leadfoot's `find` or `executeAsync` which will keep polling for the client side data until the the element is found or it reaches the timeout set by `setExecuteAsyncTimeout`.

samenz:
**Type:** bug / feature

**Description:**

Resolves #749 
Removes `sleep` from tests and moves to using `find` instead.


tomdye:
This is no longer required after: https://github.com/dojo/widgets/pull/784

Issue #780, opened 2019-09-13
Widgets 7 Plan


tomdye:
## Widgets 7 Plan

Epic for all the planned/desired enhancements to be released alongside Dojo v7.

**Note: Issue creation in progress across these**

### High Level Goals
- Widgets to be built and tested using the Dojo CLI exclusively
- Complete the Material theme
- Align and improve upon API of existing widgets
- Expansion of Widget library
- Create high-level layout / app widgets

### Use of CLI

Widgets currently use a fragmented dev toolchain. We should convert to using the dojo cli tooling consistently and ensure it has capability to build both `@dojo/widgets` and our user's third party component libraries.

- [x] [Use the dojo cli to test widgets](https://github.com/dojo/widgets/issues/787)
- [x] [Use the dojo cli to build and run widget examples](https://github.com/dojo/widgets/issues/747)

### Theme

The goal here is to create a material theme for our entire widget library. The structure for this was started in widgets 6 but not completed. The epic below links to the themes repo.

- [ ] [Material Theme](https://github.com/dojo/themes/issues/64)
- [x] [Upgrade build-theme to webpack 4](https://github.com/dojo/cli-build-theme/issues/20)

### Example / Test changes

Our existing functional tests are known to be flakey and to absorb large quantities of developer time both writing and fixing them. The framework itself is well tested across a range of browsers so there is little need to do this here also. We should ensure that adequate manual testing / usable examples are created when widgets are worked on.

- [x] [Drop existing functional tests](https://github.com/dojo/widgets/issues/781)
- [x] [Use codesandbox example approach to demonstrate widget functionality](https://github.com/dojo/widgets/issues/782)

### Existing widget improvements

This will be partly hangovers that did not get fully completed for dojo-6.0. This will make our existing widgets more compatible with the base / reusability direction we intend to take for dojo-7.0.

- [x] [Align `onValue` / `value` / `onKey` apis](https://github.com/dojo/widgets/issues/724)
- [x] [Align mouse / pointer events](https://github.com/dojo/widgets/issues/783)
- [x] [Change `invalid` -> `valid` properties](https://github.com/dojo/widgets/issues/786)
- [x] [Add `helperText` where appropriate](https://github.com/dojo/widgets/issues/677)
- [x] Select / Combobox property alignment / menu usage
- [ ] [Dialog improvements](https://github.com/dojo/widgets/issues/982)
- [ ] [~custom validator helpers~](https://github.com/dojo/widgets/issues/788)

[...]

PR #784, opened 2019-09-17
Drop functional tests from widgets

### LINKED ISSUE #781: Drop functional tests for widgets
For dojo-7.0 I propose that we remove and cease development on functional tests for the widget library. This is for several reasons:

- The dojo framework is extensively tested across multiple browsers already
- The functional tests also form the examples and aren't always the best example of widget use
- Our functional tests are flakey and often fail for no detectable reason
- Our time would be better spent writing / improving our widgets and examples

tomdye:
As part of Widgets for dojo-7.0 we are dropping functional tests.
Please see the draft widgets 7.0 plan issue [here](https://github.com/dojo/widgets/issues/780) for more information.

Resolves #781


intern.json
```diff
@@ -13,10 +13,6 @@
 		"./dist/dev/src/common/tests/unit/all.js"
 	],
```
tomdye: I believe more of this file can be removed but I am unsure which parts.

samenz: I'm pretty sure everything with browserstack, saucelabs, selenium are not going to be needed for unit tests. Maybe try taking out most of the configuration here and see if the unit tests still run.

samenz:
Looks good to me, I'm on board with removing these. The only thing that could be improved is trimming up the intern.json file but it seems your on that.

Issue #783, opened 2019-09-13
Align and cleanup widget mouse / pointer events


tomdye:
Currently most of our widgets have almost every available dom event hooked up to a callback of the same name. This is overkill and tightly couples our widgets to the underlying dom elements. Furthermore, we use both mouse, touch and pointer events when we could simply use pointer events for consistency.

**proposed new callbacks;**
- `onOver(): void` This will be triggered when a pointer moves over the widget
- `onOut(): void` This will be triggered when a pointer moves out of the widget.

**Changed callbacks**

- `onFocus` / `onBlur` / `onClick` to only be added where appropriate and will no longer receive a value or a key.

**Removed callbacks**
- `onInput` / `onChange` Will no longer be used.
- `onTouch*` / `onMouse*` Will no longer be used.

tomdye:
closing as this is complete

PR #802, opened 2019-09-25
Add onOver and onOut callbacks and friends

tomdye:
- Add `onOver` and `onOut` pointer event callbacks to form widgets.
- Add `onKeyDown` and `onKeyUp` events to text-input/area
- Add `onDown` and `onUp` events to button

Resoves: #783


src/button/index.ts
```diff
@@ -19,7 +18,11 @@ export type ButtonType = 'submit' | 'reset' | 'button' | 'menu';
  * @property name           The button's name attribute
  * @property onBlur
```
JamesMilnerUK: Do these need comments or are we scrapping those?

src/checkbox/index.ts
```diff
@@ -18,6 +18,8 @@ import * as css from '../theme/checkbox.m.css';
  * @property mode           The type of user interface to show for this Checkbox
  * @property offLabel       Label to show in the "off" positin of a toggle
  * @property onLabel        Label to show in the "on" positin of a toggle
+ * @property onOut
```
JamesMilnerUK: Same as above about JSDoc comments

PR #805, opened 2019-10-05
Chip widget

### LINKED ISSUE #799: Chip widget
The chip widget should resemble the one in the [material suite](https://material.io/components/chips/#). It has a variety of applications in other widgets such as multi select, search filters etc.

It should accept an icon and a label as well as a close action.

We may want to consider a toggle chip for making selections in future as well. This should probably be a separate widget using the same CSS.

maier49:
**Type:** feature

Adds a chip widget that supports:
* A required `label` properyt
* An optional `icon` that can be an `IconType` for a dojo `Icon` widget, a VNode or a WNode
* An optional `onClose` property that will be called when the close icon (by default the dojo `closeIcon` type `Icon` widget) is clicked
* An optional `closeIcon` for customizing display of the close icon with the same type as the `icon` property
* An optional `onClick` callback for the chip
* A disabled property for use with clickable chips
 
Resolves #799


tomdye:
@maier49 I think that this first `chip` needs to form the `base` chip that other pre-defined chips can be created from, ie. `toggleChip`, `closableChip` etc...
With that in mind I think it's important that we enable both an `iconRenderer` and `closeRenderer` callback which can be better utilised than an `icon` property. This would allow greater flexibility for future use.
We likely want a `checked` property as well which gets passed onto the `iconRenderer` so that we can use this in future to colour / distinguish both the icon and the chips background when it has been selected as a `toggleChip`.

Couple of images from the material spec which I've found useful:

Screenshot 2019-10-15 17 22 31
Screenshot 2019-10-15 17 22 59

maier49:
@tomdye I've updated the PR per your feedback and to follow the new structure. It now has a `iconRenderer` and `closeRenderer` property, and the former gets passed the new `checked` property as well.

src/chip/README.md
```diff
@@ -0,0 +1,28 @@
...
+- Displays a label, optional icon, and an optional close icon
+
+## Example Usage
+
+```typescript jsx
+// Basic chip
+<Chip label="Displayed value" iconRenderer={() => <Icon type="plusIcon"/>}/>
+```
+
+## Theming
```
tomdye: We no longer need to put theming information in here as it's picked up by the example app

src/chip/README.md
```diff
@@ -0,0 +1,28 @@
+# @dojo/widgets/chip widget
+
+Dojo's `Chip` widget allows content and particularly user inputs to be displayed more clearly and enables interaction such as removing entered values or clicking them to trigger some action
+
+
+## Features
+
+- Displays a label, optional icon, and an optional close icon
+
+## Example Usage
```
tomdye: This can just go in the basic example now, no need to put it in the readme

src/chip/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+		label,
+		closeRenderer,
+		onClose,
+		onClick,
+		disabled,
+		checked
+	} = properties();
+	const themedCss = theme.classes(css);
+
+	const createKeydown = (callback?: () => void) => (event: KeyboardEvent) => {
```
tomdye: Does this need to be extracted out to here?

src/chip/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+	return (
+		<div
+			key="root"
+			classes={[
+				themedCss.root,
+				disabled && themedCss.disabled,
+				clickable && themedCss.clickable
+			]}
+			role={clickable ? 'button' : undefined}
+			onclick={clickable ? onClick : undefined}
```
tomdye: Can you create an interim function for this as we do elsewhere. We don't want to expose the underlying dom event

src/chip/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+					tabIndex={0}
+					role="button"
+					onclick={(event) => {
+						event.stopPropagation();
+						onClose();
+					}}
+					onkeydown={(event) => {
+						if (event.which === Keys.Enter || event.which === Keys.Space) {
+							event.stopPropagation();
+							onCloseKeydown(event);
```
tomdye: I'm not sure that we really need to expose the `key down` property for this widget. Also, would you not consider `Esc` to be a valid `close` key?

src/examples/src/config.ts
```diff
@@ -209,6 +217,52 @@ export const config: Config = {
...
+				title: 'Icon',
+				module: IconChip,
+				filename: 'Icon'
+			},
+			{
+				title: 'Clickable',
+				module: ClickableChip,
+				filename: 'Clickable'
+			},
+			{
```
tomdye: Do you think that togglechip / closeablechip etc should be showcased here or split into their own opinionated variant?

src/chip/index.tsx
maier49: If you tab over the close "button" and hit enter you'd expect it to close right? As for escape, I guess I'd never think to focus a button and click escape to close it but if that's expected behavior I can certainly add it. Should I handle that in the close button handler specifically or would handling escape anywhere on the chip component be more appropriate?

src/examples/src/config.ts
maier49: Toggling seems different enough to merit its own variant, mainly because it might want to add classes depending on the toggle state. I can't think of a good reason to creating a separate widget just for the `onClose` property.

src/chip/index.tsx
tomdye: I agree that key press should close it, just saying that I don't think we need to expose an `onKey` property as parent widgets etc won't need it.

src/chip/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+import Icon from '../icon/index';
+import { Keys } from '../common/util';
+
+export interface ChipProperties {
+	/** Renders an icon, provided with the value of the checked property */
+	iconRenderer?(checked?: boolean): RenderResult;
+	/** The label to be displayed in the widget */
+	label: string;
+	/** Renders a close icon, ignored if `onClose` is not provided */
+	closeRenderer?(): RenderResult;
```
tomdye: Doesn't this need to be passed the `onClose` callback?

src/chip/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+					}}
+					onkeydown={(event) => {
+						if (event.which === Keys.Enter || event.which === Keys.Space) {
+							event.stopPropagation();
+							event.preventDefault();
+							onClose();
+						}
+					}}
+				>
+					{closeRenderer ? closeRenderer() : <Icon type="closeIcon" />}
```
tomdye: I'd imagine this needs to have an `onRequestClose` or similar callback attached to this? We may want to differentiate between clicks on the whole chip vs. the close icon?

maier49: We are differentiating; the wrapping span around this icon calls onClose when it's clicked. That being said if you'd prefer I can handle the click here for the default case only and otherwise pass the callback to the renderer.

src/chip/index.tsx
maier49: Just to clarify, we're not exposing any type of `onKey` property. The `onCloseKeydown` _was_ only internal and has since been removed.  It's just delegating to `onClick` when for Enter and Space keys.

src/examples/src/widgets/chip/Basic.tsx
```diff
@@ -0,0 +1,14 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Chip from '@dojo/widgets/chip';
+
+const factory = create();
+
+const App = factory(function() {
+	return (
+		<virtual>
```
tomdye: Do you need the virtual tag here?

src/examples/src/widgets/chip/Clickable.tsx
```diff
@@ -0,0 +1,22 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Chip from '@dojo/widgets/chip';
+import { icache } from '@dojo/framework/core/middleware/icache';
+
+const factory = create({ icache });
+
+const App = factory(function({ middleware: { icache } }) {
```
tomdye:
```suggestion
const App = factory(function Clickable({ middleware: { icache } }) {
```

src/examples/src/widgets/chip/Basic.tsx
```diff
@@ -0,0 +1,14 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Chip from '@dojo/widgets/chip';
+
+const factory = create();
+
+const App = factory(function() {
```
tomdye:
```suggestion
const App = factory(function Basic() {
```

src/examples/src/widgets/chip/ClickableClosable.tsx
```diff
@@ -0,0 +1,32 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Chip from '@dojo/widgets/chip';
+import { icache } from '@dojo/framework/core/middleware/icache';
+
+const factory = create({ icache });
+
+const App = factory(function({ middleware: { icache } }) {
```
tomdye:
```suggestion
const App = factory(function ClickableClosable({ middleware: { icache } }) {
```

src/examples/src/widgets/chip/Closable.tsx
```diff
@@ -0,0 +1,25 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Chip from '@dojo/widgets/chip';
+import { icache } from '@dojo/framework/core/middleware/icache';
+
+const factory = create({ icache });
+
+const App = factory(function({ middleware: { icache } }) {
```
tomdye:
```suggestion
const App = factory(function Closeable({ middleware: { icache } }) {
```

src/examples/src/widgets/chip/ClosableRenderer.tsx
```diff
@@ -0,0 +1,27 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Chip from '@dojo/widgets/chip';
+import Icon from '@dojo/widgets/icon';
+import { icache } from '@dojo/framework/core/middleware/icache';
+
+const factory = create({ icache });
+
+const App = factory(function({ middleware: { icache } }) {
```
tomdye:
```suggestion
const App = factory(function CloseableRenderer({ middleware: { icache } }) {
```

src/examples/src/widgets/chip/Disabled.tsx
```diff
@@ -0,0 +1,20 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Chip from '@dojo/widgets/chip';
+
+const factory = create();
+
+const App = factory(function() {
```
tomdye:
```suggestion
const App = factory(function Disabled() {
```

src/examples/src/widgets/chip/Disabled.tsx
```diff
@@ -0,0 +1,20 @@
...
+	return (
+		<virtual>
```
tomdye: I don't think you need the virtual here

src/examples/src/widgets/chip/Icon.tsx
```diff
@@ -0,0 +1,15 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Chip from '@dojo/widgets/chip';
+import Icon from '@dojo/widgets/icon';
+
+const factory = create();
+
+const App = factory(function() {
+	return (
+		<virtual>
```
tomdye: dont need virtual

src/examples/src/widgets/chip/Icon.tsx
```diff
@@ -0,0 +1,15 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Chip from '@dojo/widgets/chip';
+import Icon from '@dojo/widgets/icon';
+
+const factory = create();
+
+const App = factory(function() {
```
tomdye:
```suggestion
const App = factory(function Icon() {
```

src/examples/src/widgets/chip/Basic.tsx
maier49: Nope

PR #807, opened 2019-10-10
Form

### LINKED ISSUE #800: Form
We should create a Form pattern that can be used to easily set and get values from forms.
It should handle things like dirty state and validation. 
This could be implemented as a widget or as a middleware that can be used in applications.

- The Form should use the `name` / `setValue` / `getValue` properties of the form fields created within it.
- The Form should handle the `dirty` state of it's fields
- The form should handle field validation
- The form should have it's own `onValue` callback which reports back the form value.

KaneFreeman:
**Type:** feature

**Description:**

Form middleware and wrapping widget. Following a similar pattern to #803 for the widget.

POC: https://codesandbox.io/s/github/KaneFreeman/form-middleware-poc/tree/master/

Resolves #800

### DIFF
--- a/src/examples/src/widgets/form/Basic.tsx
+++ b/src/examples/src/widgets/form/Basic.tsx
@@ -1,0 +1,173 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import { createICacheMiddleware } from '@dojo/framework/core/middleware/icache';
+
+import Button from '@dojo/widgets/button';
+import TextInput from '@dojo/widgets/text-input';
+import Form from '@dojo/widgets/form';
+import { FormMiddleware } from '@dojo/widgets/form/middleware';
+
+const icache = createICacheMiddleware<{
+	basic: Partial<Fields>;
+	basicOnValue: Partial<Fields>;
+	dirty: boolean;
+}>();
+
+const factory = create({ icache });
+
+interface Fields {
+	firstName: string;
+	middleName?: string;
+	lastName: string;
+	email?: string;
+}
+
+const App = factory(function({ middleware: { icache } }) {
+	const results = icache.get('basic');
+	const onValueResults = icache.get('basicOnValue');
+
+	return (
+		<virtual>
+			<Form
+				initialValue={{
+					firstName: 'Billy'
+				}}
+				onSubmit={(values) => icache.set('basic', values)}
+				onValue={(values) =>
+					icache.set('basicOnValue', { ...icache.get('basicOnValue'), ...values })
+				}
+			>
+				{({ value, valid, disabled, field, reset }: FormMiddleware<Fields>) => {
+					const firstName = field('firstName', true);
+					const middleName = field('middleName');
+					const lastName = field('lastName', true);
+					const email = field('email');
+
+					return [
+						<TextInput
+							key="firstName"
+							label="First Name"
+							placeholder="Enter first name (must be Billy)"
+							pattern="Billy"
+							required={true}
+							value={firstName.value()}
+							valid={firstName.valid()}
+							onValue={firstName.value}
+							onValidate={firstName.valid}
+							disabled={firstName.disabled()}
+						/>,
+						<TextInput
+							key="middleName"
+							label="Middle Name"
+							placeholder="Enter a middle name"
+							required={middleName.required()}
+							value={middleName.value()}
+							valid={middleName.valid()}
+							onValue={middleName.value}
+							onValidate={middleName.valid}
+							maxLength={5}
+							disabled={middleName.disabled()}
+						/>,
+						<TextInput
+							key="lastName"
+							label="Last Name"
+							placeholder="Enter a last name"
+							required={true}
+							value={lastName.value()}
+							valid={lastName.valid()}
+							onValue={lastName.value}
+							onValidate={lastName.valid}
+							minLength={2}
+							disabled={lastName.disabled()}
+						/>,
+						<TextInput
+							key="email"
+							label="Email"
+							placeholder="Enter an email address"
+							required={false}
+							value={email.value()}
+							valid={email.valid()}
+							onValue={email.value}
+							onValidate={email.valid}
+							type="email"
+							pattern="^[^\s@]+@[^\s@]+\.[^\s@]{2,}$"
+							disabled={email.disabled()}
+						/>,
+						<Button
+							key="fill"
+							type="button"
+							disabled={disabled()}
+							onClick={() => {
+								value({
+									firstName: 'Billy',
+									middleName: '',
+									lastName: 'Bob'
+								});
+							}}
+						>
+							Fill
+						</Button>,
+						<Button
+							key="requireMiddleName"
+							type="button"
+							disabled={disabled()}
+							onClick={() => middleName.required(!middleName.required())}
+						>
+							{`Make middle name ${middleName.required() ? 'optional' : 'required'}`}
+						</Button>,
+						<Button
+							key="reset"
+							type="button"
+							disabled={disabled()}
+							onClick={() => reset()}
+						>
+							Reset
+						</Button>,
+						<Button
+							key="disableForm"
+							type="button"
+							onClick={() => disabled(!disabled())}
+						>
+							{`${disabled() ? 'Enable' : 'Disable'} Form`}
+						</Button>,
+						<Button
+							key="disableEmail"
+							type="button"
+							disabled={disabled()}
+							onClick={() => email.disabled(!email.disabled())}
+						>
+							{`${email.disabled() ? 'Enable' : 'Disable'} Email`}
+						</Button>,
+						<Button key="submit" type="submit" disabled={!valid() || disabled()}>
+							Submit
+						</Button>
+					];
+				}}
+			</Form>
+			{}
+			{onValueResults && (
+				<div key="onValueResults">
+					<h2>onValue Results</h2>
+					<ul>
+						<li>First Name: {onValueResults.firstName}</li>
+						<li>Middle Name: {onValueResults.middleName}</li>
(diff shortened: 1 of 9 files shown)


samenz:
Looks good @KaneFreeman ! 👍 Could you write a quick form example for the example page as well?

src/form/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+interface FormProperties {
+	onSubmit(values: any): void;
+	renderer(options: {
+		values: <S extends Record<string, any> = any>(
+			values?: Partial<S>
+		) => {
+			set<K extends keyof S>(key: K, value: S[K]): S[K];
+			get<K extends keyof S>(key: K): S[K] | undefined;
+		};
+		required: (key: string, required?: boolean) => boolean;
```
samenz: Want to make `required`, `valid`, `disabled` optional params?

src/form/middleware.ts
```diff
@@ -0,0 +1,155 @@
+import { create } from '@dojo/framework/core/vdom';
```
samenz: Looking at [Tom's PR](#803) it seems he tested the middleware as well. Might be something to consider here

KaneFreeman: Yeah I can do that. It will be almost an exact copy of the widget version (which does fully cover the middleware) but probably a good idea regardless.

tomdye:
Thanks for your efforts on this so far @KaneFreeman. 
How about adding an `onValue` callback that gets called each time a value is updated and a `initialValue` property that sets up the form with initial values. (Both of these can be found in the checkbox-group widget)

dojo-site:
* dojo.widgets: https://dojowidgets-j6a94v0m5.now.sh

KaneFreeman:
@tomdye This is ready for review again. I have updated the `renderer` to be a child (same as you are making the menu widget) and added the `initialValue` and `onValue` properties. Tests and example have been updated appropriately.

dojo-site:
* dojo.widgets: https://dojowidgets-1cfqspdg4.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-ht6qnk7yj.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-9dh2gzrlx.now.sh

src/common/tests/unit/all.ts
```diff
@@ -0,0 +1,43 @@
+import './util';
```
tomdye: this file has been removed elsewhere

src/examples/src/widgets/form/Basic.tsx
```diff
@@ -0,0 +1,165 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import { icache } from '@dojo/framework/core/middleware/icache';
+
+import Button from '@dojo/widgets/button';
+import TextInput from '@dojo/widgets/text-input';
+import Form from '@dojo/widgets/form';
+
+const factory = create({ icache });
```
tomdye: how about typing this via `createICacheMiddleware<interface>()`?

src/examples/src/widgets/form/Basic.tsx
```diff
@@ -0,0 +1,165 @@
...
+	const results = icache.get<Partial<Fields>>('basic');
+	const onValueResults = icache.get<Partial<Fields>>('basicOnValue');
+
+	return (
+		<virtual>
+			<Form
+				initialValue={{
+					firstName: 'Billy'
+				}}
+				onSubmit={(values) => icache.set('basic', values)}
```
tomdye: do you think this needs both an `onSubmit` and an `onValue` ?

src/examples/src/widgets/form/Basic.tsx
```diff
@@ -0,0 +1,165 @@
...
+					required('lastName', true);
+
+					return [
+						<TextInput
+							key="firstName"
+							label="First Name"
+							placeholder="Enter first name (must be Billy)"
+							pattern="Billy"
+							required={true}
+							value={get('firstName')}
```
tomdye: could this work as `field('firstName').value`, `field('firstName').required(true)` etc perhaps ? then this would line up to checkbox group etc better and make the number of functions passed into the renderer more succinct.

src/form/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+import createFormMiddleware, { Validity } from './middleware';
+import * as css from '../theme/form.m.css';
+
+const form = createFormMiddleware();
+
+interface FormProperties {
+	/** The initial form value */
+	initialValue?: Record<string, any>;
+	/** Callback for when the form is submitted */
+	onSubmit(values: Record<string, any>): void;
```
tomdye: as per the example comment, do you think we need both `onSubmit` and `onValue` ?

src/examples/src/widgets/form/Basic.tsx
KaneFreeman: I would imagine we need some kind of onSubmit call back that can be hooked up to the form. Passing the values as part of that just seems natural.

dojo-site:
* dojo.widgets: https://dojowidgets-ltw0fqdz5.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-7ztlb5u29.now.sh

src/form/middleware.ts
```diff
@@ -0,0 +1,195 @@
...
+						...requiredValues,
+						[name]: required
+					});
+				}
+				return {
+					value: (newValue?: any): any => {
+						const values = icache.getOrSet('values', {}) as S;
+						if (newValue !== undefined && values[name] !== newValue) {
+							icache.set('values', { ...values, [name]: newValue });
+							const callbacks = cache.get<Callbacks<S>>('callbacks');
```
tomdye: why is this using both cache and icache?

src/form/middleware.ts
```diff
@@ -0,0 +1,195 @@
...
+				icache.set('values', {});
+				let valid: Valid = {};
+				const values = icache.getOrSet('valid', {});
+				Object.keys(values).map((key) => {
+					valid[key] = { valid: undefined, message: '' };
+				});
+				icache.set('valid', valid);
+				icache.set('required', {});
+			},
+			field<K extends keyof S>(name: K, required = false): Field<S, K> {
```
tomdye: can name / `k` be changed to extend or use typeof string here to avoid the use of `as string` elsewhere?

src/form/middleware.ts
KaneFreeman: was aiming to avoid another render when the callback gets registered. I can switch it back though if desired.

dojo-site:
* dojo.widgets: https://dojowidgets-blvwnu87l.now.sh

tsconfig.json
```diff
@@ -34,7 +34,8 @@
 		"types": [ "intern" ],
 		"paths": {
 			"@dojo/widgets/*": ["src/*"]
-		}
+		},
+		"keyofStringsOnly": true
```
matt-gadd: will this not fail downstream? i'm not sure if we can do this

dojo-site:
* Docs Deployment: https://dojowidgets-az0al8w6s.now.sh
* Tests Deployment: https://dojowidgets-c98wlj6at.now.sh

src/examples/src/config.ts
```diff
@@ -25,6 +25,7 @@ import Username from './widgets/constrained-input/Username';
 import BasicDialog from './widgets/dialog/Basic';
 import BasicEmailInput from './widgets/email-input/Basic';
 import BasicGrid from './widgets/grid/Basic';
+import BasicForm from './widgets/form/Basic';
```
rorticus: I believe these imports should be sorted alphabetically now (as per @tomdye 's request)

src/form/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { RenderResult } from '@dojo/framework/core/interfaces';
+
+import createFormMiddleware, { FormMiddleware } from './middleware';
+import * as css from '../theme/form.m.css';
+
+const form = createFormMiddleware();
+
+interface FormProperties {
+	/** The initial form value */
+	initialValue?: Record<string, any>;
```
rorticus: I wonder if this should be `type FormValue = Record<string, any>` so it's repeated in here a bunch of times. Not sure if we'll ever need to change the type, but if so it might make lives easier!

src/form/middleware.ts
```diff
@@ -0,0 +1,192 @@
...
+}
+
+export const createFormMiddleware = <S extends Record<string, any> = any>() => {
+	const icache = createICacheMiddleware<State<S>>();
+	const factory = create({ icache, cache });
+
+	const formMiddleware = factory(function Form({
+		middleware: { icache, cache }
+	}): FormMiddleware<S> {
+		return {
```
rorticus: Can we include a dirty state? It would be nice to use this to tell if the current values don't match the initial values. This would be used for "you have unsaved changes" confirmations and such.

src/examples/src/config.ts
tomdye: yes please, helps massively with merge conflict avoidance.

src/form/middleware.ts
tomdye: perhaps this is something you could raise an issue for and we could do as a follow up?

KaneFreeman: Created https://github.com/dojo/widgets/issues/954

dojo-site:
* Docs Deployment: https://dojowidgets-5c6zir661.now.sh
* Tests Deployment: https://dojowidgets-lff5ly6e3.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-5c6zir661.now.sh
* Tests Deployment: https://dojowidgets-lff5ly6e3.now.sh

Issue #740, opened 2019-05-30
Min, max and step missing from text-input


tomdye:
** Enhancement** 

We should add min, max, and step properties to the text-input widget. These will be used when type is set to number.

PR #736, opened 2019-05-23
Add Mix, Max, Step properties to Text-Input

thomasxb:
**Type:** enhancement

Resolves: 

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code matches the [style guide](https://github.com/dojo/framework/blob/master/STYLE.md)
* [ ] Unit or Functional tests are included in the PR

**Description:**

Resolves #740

PR #839, opened 2019-10-16
Add min / max / step to text-input

tomdye:
Replaces: https://github.com/dojo/widgets/pull/736
Resolves: #740


tomdye:
@thomasxb sorry this has taken so long. You should see this surfacing as a specific `NumberInput` in the near future.

PR #840, opened 2019-10-16
Email input

### LINKED ISSUE #789: Email input widget
The email input widget will essentially be a base `text-input` with opinionated default properties (such as `type: email`) and a subtly different styling. ie. an `@`/ `mail` icon symbol in the leading section of the input.

The email input should not expose properties such as `min` and `max` and others that do not apply to an email input. Native validation should work for email addresses.

This will expand our widget suite and allow a quick and consistently validated email input for use in dojo applications.

samenz:
**Type:** Feature

**Description:** Adds simple email input widget. The email input removes the following text-input properties from it's interface:
- customValidator
- max
- min
- step
- onKeyDown
- onKeyUp
- pattern
- type

**Code Description:**
- Add email input widget
- Add email input widget to examples

Resolves #789


src/email-input/index.ts
```diff
@@ -0,0 +1,125 @@
...
+			onBlur,
+			onFocus,
+			onClick,
+			onOver,
+			onOut,
+			valid,
+			controls
+		} = this.properties;
+
+		return v(
```
tomdye: Do you need the wrapping `div` ?

src/email-input/index.ts
```diff
@@ -0,0 +1,125 @@
...
+			'div',
+			{
+				key: 'root',
+				role: 'presentation'
+			},
+			[
+				w(TextInput, {
```
tomdye: How about doing `<TextInput {...this.properties} type="email" />`

tomdye: Do you think you could change this to `tsx` ?

src/email-input/index.ts
samenz: Nope, it looked like a pattern that was done in text-input, I can take it out

src/email-input/index.ts
samenz: Yea I can change this to `tsx`, are we trying to write all new widgets in `tsx` cause I'll note that for the next one.

samenz: Yea I can change this to tsx, are we trying to write all new widgets in tsx cause I'll note that for the next one.

PR #842, opened 2019-10-17
Revert "Email input"

tomdye:
Reverts dojo/widgets#840


tomdye:
Reverting this as I made a mistake in merging it prematurely.

PR #852, opened 2019-10-21
Fix Time-Picker example.

tomdye:
This PR removes the `setLocaleData` that `Time-Picker` was previously using and replaces it with the appropriate `dojorc` config for cldr/locale.

It also fixes a bug where locale date etc was being initialised outside of the functional widget scope and thus was causing issues.
For example, in the 12-hour time format example, `getDateFormatter` was being called outside of the functional widget and thus would be ran when the widget was imported rather than when it was actually used / rendered. This meant that globalise was being used before it had been configured and thus threw the following error:

Screenshot 2019-10-21 11 01 40


tomdye:
@maier49 FYI ^^

PR #865, opened 2019-11-05
framework #569: Focus meta invalidation change

tomdye:
**Type:** bug

**Description:**

Creates local focus meta copy that only invalidates the calling widget when the node it is interested in gains / loses focus.

Resolves https://github.com/dojo/framework/issues/569


src/meta/Focus.ts
```diff
@@ -0,0 +1,79 @@
...
+					!!(
+						node.contains(global.document.activeElement) ||
+						(this._activeElement && node.contains(this._activeElement))
+					)
+			)
+		) {
+			this.invalidate();
+		}
+
+		this._activeElement = global.document.activeElement;
```
tomdye: this is the only change

PR #866, opened 2019-11-05
Rename numberInput config entry

tomdye:
**Type:** bug 

Previous naming of `NumberInput` config entry in the example app broke our doc generation.
The standard is to use hyphenated naming as per the folder name / css file name.


tomdye:
@vansimke FYI

PR #885, opened 2019-11-14
Menu

### LINKED ISSUE #801: Menu widget
We should create a generic Menu widget that can be consistently used as the drop down / context menu for the rest of our widget suite.
Currently our widget suite uses a common `ListBox` widget but it is not particularly flexible or easily reusable as a standard menu.

The widget should be similar to the [menu in the material design suite](https://material.io/components/menus/).

The menu contents should be populated by `MenuItem` widgets (which we should also create).

### Menu

- Menu should support `open` booean
- Menu should support `onClose` callback

### MenuItem

- MenuItem should support `onClick` callback
- MenuItem should support `leading` callback to specify leading dom section
- MenuItem should support `trailing` callback to specify trailing dom section
- MenuItem should support children for nested menus

### Example

```ts
<Menu open={false}>
   <MenuItem onClick={doSomething}>Save</MenuItem>
   <MenuItem onClick={doAnotherThing}>Edit</MenuItem>
</Menu>
```

tomdye:
**Type:** feature

**Description:**

Adds a Menu widget and associated MenuItem widget. This will replace existing `Listbox` component.
Can be used as a standalone menu or as part of a select / typeahead / context menu etc.
Popup / Select / Typeahead to follow.

Resolves #801


dojo-site:
* dojo.widgets: https://dojowidgets-qpw2ic6oa.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-f8gqhxl71.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-8s0o5ykxq.now.sh

src/menu/Menu.tsx
```diff
@@ -0,0 +1,274 @@
...
+	/** Optional proprty to set the activeIndex when it is being controlled externally */
+	activeIndex?: number;
+	/** Determines if the widget can be focused or not. If the active index is controlled from elsewhere you may wish to stop the menu being focused and receiving keyboard events */
+	focusable?: boolean;
+	/** Callback called when menu root is focused */
+	onFocus?(): void;
+	/** Callback called when menu root is blurred */
+	onBlur?(): void;
+	/** Property to determine how many items to render. Not passing a number will render all results */
+	numberInView?: number;
```
bryanforbes: What do you think about `itemsInView` instead of `numberInView`?

src/menu/Menu.tsx
```diff
@@ -0,0 +1,274 @@
...
+	function getComputedIndexFromInput(key: string) {
+		const existingTimer = cache.get<NodeJS.Timer>('resetInputTextTimer');
+		let inputText = cache.get<string>('inputText') || '';
+		existingTimer && clearTimeout(existingTimer);
+
+		cache.set(
+			'resetInputTextTimer',
+			setTimeout(() => {
+				cache.set('inputText', '');
+			}, 800)
```
bryanforbes: Why 800? Should this be configurable?

src/menu/Menu.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+		inputText += `${key}`;
+		cache.set('inputText', inputText);
+
+		return findIndex(options, ({ disabled, value, label }, i) => {
+			if (disabled) {
+				return false;
+			}
+
+			if ((label || value).toLowerCase().indexOf(inputText.toLowerCase()) === 0) {
```
bryanforbes: Should the user be able to pass in their own comparison function to do case-sensitive matching?

src/examples/src/widgets/menu/Controlled.tsx
```diff
@@ -0,0 +1,43 @@
...
+			<Menu
+				focusable={false}
+				numberInView={6}
+				options={states}
+				onActiveIndexChange={(index: number) => {
+					icache.set('activeIndex', index);
+				}}
+				activeIndex={activeIndex}
+				onValue={(value) => {
+					console.log(`selected: ${value}`);
```
rorticus: Would this value be better being displayed, rather than logged?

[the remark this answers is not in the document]
tomdye: can do

src/examples/src/widgets/menu/ItemRenderer.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const factory = create();
+
+export default factory(function ItemRenderer() {
+	return (
+		<Menu
+			options={states}
+			initialValue={'cat'}
+			onValue={(value) => {
+				console.log(`selected: ${value}`);
```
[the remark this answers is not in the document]
tomdye: can do

src/menu/Menu.tsx
```diff
@@ -0,0 +1,274 @@
...
+	const div = global.document.createElement('div');
+	div.style.position = 'absolute';
+	global.document.body.appendChild(div);
+	r.mount({ domNode: div, sync: true });
+	const dimensions = div.getBoundingClientRect();
+	global.document.body.removeChild(div);
+	return dimensions.height;
+};
+
+const menuFactory = create({
```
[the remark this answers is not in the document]
tomdye: I believe, at the time I was exporting two widgets from this file, hence the need for different names, can easily be renamed to `factory`

src/menu/Menu.tsx
tomdye: fine by me

src/menu/Menu.tsx
tomdye: I took this delay from the existing select that does the same, I don't think it needs to be configurable

src/menu/Menu.tsx
tomdye: again, this is a relatively opinionated widget that deals with text labels / values. this is simply allowing the user to type `n` when the menu is focused to select `nevada` (in the states example)

dojo-site:
* dojo.widgets: https://dojowidgets-gja8kbrn0.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-ei8kiesu7.now.sh

src/examples/src/config.ts
```diff
@@ -333,6 +336,18 @@ export const config: Config = {
...
 	},
+	menu: {
+		overview: {
+			example: {
+				module: BasicMenu,
+				filename: 'Basic'
+			}
+		},
+		examples: [
+			{ title: 'ControlledMenu', module: ControlledMenu, filename: 'ControlledMenu' },
```
agubler: Is the filename `ControlledMenu`?

src/examples/src/widgets/menu/Controlled.tsx
```diff
@@ -0,0 +1,44 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Menu from '@dojo/widgets/menu';
+import icache from '@dojo/framework/core/middleware/icache';
+import states from './states';
+
+const factory = create({ icache });
+
+export default factory(function Controlled({ middleware: { icache } }) {
+	const activeIndex = icache.getOrSet<number>('activeIndex', 0);
```
agubler: Do you need to pass the generic? Does it not infer from `0`?

src/menu/index.tsx
```diff
@@ -0,0 +1,276 @@
...
+			icache.set('activeIndex', index);
+		}
+	}
+
+	function _setValue(value: string) {
+		icache.set('value', value);
+		onValue(value);
+	}
+
+	function _onKeyDown(event: KeyboardEvent) {
```
agubler: Do we need to have an underscore prefix? It was originally to indicate private functions on a class, but this is not a class widget.

src/menu/index.tsx
```diff
@@ -0,0 +1,276 @@
...
+							_setValue(value);
+						}}
+						active={active}
+						theme={themeProp}
+						onRequestActive={() => {
+							if (focus.isFocused('root') || !focusable) {
+								_setActiveIndex(index);
+							}
+						}}
+						onActive={(dimensions: DimensionResults) => {
```
agubler: You shouldn't need to type the argument as it should be inferred

src/menu/index.tsx
```diff
@@ -0,0 +1,276 @@
...
+	icache: createICacheMiddleware<MenuICache>(),
+	focus,
+	dimensions,
+	cache,
+	theme
+}).properties<MenuProperties>();
+
+export const Menu = factory(function({
+	properties,
+	middleware: { icache, cache, focus, dimensions, theme }
```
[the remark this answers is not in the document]
tomdye: because not all of the cache uses require an invalidation, was that wrong?

dojo-site:
* dojo.widgets: https://dojowidgets-cbyfhf9vh.now.sh

src/examples/src/config.ts
tomdye: no, fixed

src/menu/MenuItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+	if (icache.get('active') !== active) {
+		icache.set('active', active);
+		active && onActive(dimensions.get('root'));
+	}
+
+	return (
+		<div
+			key="root"
+			onpointermove={throttle(() => {
```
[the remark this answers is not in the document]
tomdye: stops the `onRequestActive` call from being made thousands of times.

dojo-site:
* dojo.widgets: https://dojowidgets-f421k6su6.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-4r2uy1fl1.now.sh

src/menu/MenuItem.tsx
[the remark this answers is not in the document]
agubler: Did it cause a problem when it was?

src/menu/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+			onSelect: () => {},
+			active: false,
+			onRequestActive: () => {},
+			onActive: () => {},
+			scrollIntoView: false
+		};
+
+		const itemHeight = icache.getOrSet(
+			'itemHeight',
+			offscreenHeight(
```
agubler: I do think these things are clearer when they are not inlined.

tomdye: i thought i already did that... looking

dojo-site:
* dojo.widgets: https://dojowidgets-4z2slx4u4.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-30yz11o1g.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-30yz11o1g.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-pbaj9kwzg.now.sh

src/menu/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+			key="root"
+			classes={classes.menu}
+			tabIndex={focusable ? 0 : -1}
+			onkeydown={onKeyDown}
+			focus={() => shouldFocus}
+			onfocus={onFocus}
+			onblur={onBlur}
+			styles={rootStyles}
+			role="menu"
+			aria-hidden="true"
```
mwistrand: Why is `aria-hidden` always `"true"`?

src/menu/MenuItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+				classes.item,
+				selected && classes.itemSelected,
+				active && classes.itemActive,
+				disabled && classes.itemDisabled
+			]}
+			onpointerdown={() => {
+				!disabled && onSelect();
+			}}
+			scrollIntoView={scrollIntoView}
+			role="menuitem"
```
mwistrand: `role="menuitem"` is correct for action or context menus, but if this is intended for use in other contexts like selects or typeaheads, then `"menuitem"` would no longer be correct. In other words, `Menu` and `MenuItem` are generic enough that no `role` or `aria-` property will be correct in every use case, so it would be better to require each that implementation provides its own accessibility properties.

src/menu/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+		<div
+			key="root"
+			classes={classes.menu}
+			tabIndex={focusable ? 0 : -1}
+			onkeydown={onKeyDown}
+			focus={() => shouldFocus}
+			onfocus={onFocus}
+			onblur={onBlur}
+			styles={rootStyles}
+			role="menu"
```
mwistrand: Same comment as earlier: `role="menu"` is not correct for every use case for which we intend this widget, and it would be best to allow each implementation to specify its own accessibility-related properties.

matt-gadd: can we think of any other scenario's other than `menu` and `listbox` that the menu would ever be reasonably used for? I think we should really try and avoid basically the aria html forwarding bags we've done elsewhere if this is designed to be easily used by an end user.

src/menu/index.tsx
tomdye: I was copying the aria usage of the material menu

src/menu/index.tsx
tomdye: agreed, menu / listbox are the only cases i can think of

dojo-site:
* dojo.widgets: https://dojowidgets-mgcrw54zc.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-4ovhgtfl1.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-3rcws6o97.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-d9hbaw1g0.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-n41wy4vwx.now.sh

src/menu/index.tsx
```diff
@@ -0,0 +1,314 @@
...
+			classes={classes.menu}
+			tabIndex={focusable ? 0 : -1}
+			onkeydown={onKeyDown}
+			focus={() => shouldFocus}
+			onfocus={onFocus}
+			onblur={onBlur}
+			styles={rootStyles}
+			role={listBox ? 'listbox' : 'menu'}
+			aria-orientation="vertical"
+			aria-activedescendant={`${idBase}-item-${activeIndex}`}
```
mwistrand: This should not be set if there if `activeIndex == null`. For example, on the deployed site for this PR, the example menu's `aria-activedescendant` is always `{id}-undefined` since the active index is never set.

tomdye: fixing that now

dojo-site:
* dojo.widgets: https://dojowidgets-jxrz4zxzk.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-f7ydtqqwl.now.sh

PR #906, opened 2019-11-19
Add theme prefix and fix classes clash

### LINKED ISSUE #904: Theme composition as way of theming child widgets
The current theme composition works only for widget variants. With a small tweak to add a `prefix` we would be able to use this to theme child widgets via their parents.

---

Given a parent widget with a child widget and a `theme.compose` `prefix` of `child`.

We should be able to compose together themes such that css rules for `childRoot` passed to `parent` can be applied to the `child` as `root`.
The `child` should only receive css rules from `parent` if those classes are actually themed. The vanilla implementation of `parent` theme should not be passed down to `child`.

**Rules:**

1) If no theme is passed to parent, child should use it's base theme.
2) If a theme is passed to child, it should be used.
3) On a class-by-class basis, if a theme is passed to parent intended for child, it should override the child theme.
4) If the parent specifies classes for child in it's base css, but no theme is passed, child shoudl use it's own base css.
5) The base css of parent should not specify any rules for child, these classes are themeing extensions only.

This pattern could be very useful when theming things like a menu with menu items or an accordion with title panes. Previously, both of these would have required the developer to know which widgets both of these were using internally. This new approach increases class name discoverability and means the user no longer needs to know about the internal workings of each widget.

### LINKED ISSUE #905: Theme composition breaks classes
# bug

Currently, using `theme.compose` does not take into consideration that there may be multiple classes returned from `theme.classes` due to use of `composes` or `classes`. This means that any combined use of these will result in a widget being incorrectly themed.

tomdye:
**Type:** bug

**Description:**

Adds `prefix` and fixes issue related to `theme.compose` when used with `classes`.

Resolves #904 
Resolves #905

### DIFF
--- a/src/middleware/tests/theme.spec.tsx
+++ b/src/middleware/tests/theme.spec.tsx
@@ -200,9 +200,6 @@ registerSuite('theme-middleware', {
 			const baseCss = { ' _key': 'base', root: 'base-root', input: 'base-input' };
 			const variantCss = { ' _key': 'variant', prefixRoot: 'variant-root' };
 			const theme = {
-				base: {
-					root: 'base-theme-root'
-				},
 				variant: {
 					prefixInput: 'variant-theme-input'
 				}
@@ -218,7 +215,7 @@ registerSuite('theme-middleware', {
 
 			h.expect(
 				baseTemplate.setChildren('@root', () => [
-					JSON.stringify({ root: 'variant-theme-root', input: 'base-input' })
+					JSON.stringify({ root: 'base-root', input: 'variant-theme-input' })
 				])
 			);
 		}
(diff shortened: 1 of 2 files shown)


dojo-site:
* dojo.widgets: https://dojowidgets-1ujfu6qsb.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-837oh3n5c.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-8qkwoi5ne.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-ckndwuvne.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-qdczj95t1.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-670rym5ay.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-1o8du7v8u.now.sh

dojo-site:
* dojo.widgets: https://dojowidgets-521340kun.now.sh

PR #914, opened 2019-11-22
Removing i18n from constrained input tests

rorticus:
There is an issue running the constrained input tests on the site because of some inconsistencies with i18n. I'm removing i18n from the equation by always mocking the validation middleware since its not really important for these tests.


dojo-site:
* Docs Deployment: https://dojowidgets-6r3xhl4zx.now.sh
* Tests Deployment: https://dojowidgets-2z47jp3oy.now.sh

Issue #847, opened 2019-10-18
Theming for widget variants


tomdye:
We need to develop a way to theme widget variants.

For example, it should be possible for a user to target the `raised-button` or the `email-input` directly and theme is differently to the standard `button` / `input`.

### Current code (wrong)

Currently we are going some way to achieving this using `classes` along with `this.theme`, as seen [here](https://github.com/dojo/widgets/blob/master/src/raised-button/index.tsx#L15) in the `raised-button` render function.

The problem with this approach is that `classes` is an additive function whereas `theme` is designed to replace the existing styling. This means that the raised button, when themed, will receive the theme styles for both `button` and `raised-button`.

eg. given the following code and the current approach, the raised button will have a green background and italic purple text. This is not correct as the `raisedButtonCss` root class should override the `buttonCss` root class.

```css
/* buttonCss */
.root {
	color: red;
	background: green;
	padding: 20px;
	font-size: 20px;
}

/* raisedButtonCss */
.root {
	color: purple;
	font-style: italic;
}
```

```ts
export default factory(function Basic() {
	const theme = {
		'@dojo/widgets/button': buttonCss,
		'@dojo/widgets/raised-button': raisedButtonCss
	};

	return (
		<virtual>
			<Button disabled theme={theme}>Normal Buton</Button>
			<RaisedButton disabled theme={theme}>Raised Button</RaisedButton>
		</virtual>
	);
});
```

In order to achieve the correct result, where we only get the styles from the `raisedButtonCss` and see purple italics without the green background we need to do something cleverer with `this.theme` and the object that gets passed to the `button`.

### Class based solution (better)

To fix this issue correctly and allow widget variants (ie. `raised-button`) to be fully themeable we need to compose together the themed classes for both `button` and `raised-button` and then pass this object as `theme` to the underlying widget.

``` ts
@theme(css)
@theme(buttonCss)
export class RaisedButton extends ThemedMixin(WidgetBase)<RaisedButtonProperties> {
	protected render(): DNode {
		const classes = Object.keys(css).reduce((existing, key) => {
			(existing as any)[key] = this.theme((css as any)[key]);
			return existing;
		}, {});

		const buttonClasses = Object.keys(buttonCss).reduce((existing, key) => {
			(existing as any)[key] = this.theme((css as any)[key]);
			return existing;
		}, {});

		const combined = { ...buttonClasses, ...classes };

		return (
			<Button
				{...this.properties}
				theme={{
					'@dojo/widgets/button': combined
[...]

PR #922, opened 2019-11-26
Theme compose reloaded

### LINKED ISSUE #916: Amend theme composition to allow variant classes to be used from base class
**Enhancement**

In it's current form, the theme composition code only uses the variant classes if they have been themed. This means that styles cannot currently be overridden using the base theme.
This is problematic for widgets such as outlined button which needs to appear visually different to a standard button even in it's un-themed view.

tomdye:
**Type:** bug / feature

**Description:**

Amends the `theme.compose` code to use the base-variant css if the base component is not themed.
The effect of this is that the base-theme for the variant will be used so that widgets such as `outline-button` can have an appropriate base theme.

**impacts of this change:**

- all widgets that use `theme.compose` will have to manually use `composes` within their base-theme css on classes they wish to be themeable but to use the base-component's css.

ie. 

```css
/* outline-button.m.css */
.root {
   composes: root from './button.m.css';
   border: 1px solid black;
}
```

- In the above, outline-button case, if a dojo / material etc theme is created and applied via a theme injector for `button`, but not `outline-button`, the outline-button will lose it's base-css and use the button's theme.

Resolves #916

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@dojo/widgets",
-  "version": "6.0.1-pre",
+  "version": "7.0.0-pre",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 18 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-3qn37u5fl.now.sh
* Tests Deployment: https://dojowidgets-g49rcu5xt.now.sh

aciccarello:
Great, I had noticed that you could compose the root from the base component.

dojo-site:
* Docs Deployment: https://dojowidgets-3qn37u5fl.now.sh
* Tests Deployment: https://dojowidgets-fztogmyn3.now.sh

PR #940, opened 2019-12-02
Fix the theme middleware to deal with the compose scenarios

agubler:
**Type:** bug

The following has been addressed in the PR:

* [ ] There is a related issue
* [x] All code matches the [style guide](https://github.com/dojo/framework/blob/master/STYLE.md)
* [x] Unit tests are included in the PR
* [ ] Any widget variant uses `theme.compose` like [this](https://github.com/dojo/widgets/issues/847)
* [ ] WidgetProperties are exported

**Description:**

Update `theme` middleware to support default variant themes without requiring the use of `composes` for the default base theme in the variant css.

1) The combined classes from the variant css and base css are available for theming.

```css
// base widget default theme
.root {

}

// variant widget default theme
.extra {

}
```

This will enable both the `root` and `extra` classes to be themed by targeting the variants key, for example:

```tsx
// theme
{
    '@dojo/widgets/variant': {
        root: 'variant_theme_root',
        extra: 'variant_theme_extra'
    }
}
```

2) Falls back to the base theme when not specified in the variants theme

```tsx
// theme
{
    '@dojo/widgets/base': {
        root: 'base_theme_root'
    },
    '@dojo/widgets/variant': {
        extra: 'variant_theme_extra'
    }
}
```

3) Add additional classes specified in the `classes` for both the variant and base css

```tsx
// theme
{
    '@dojo/widgets/base': {
        root: 'base_theme_root'
    },
    '@dojo/widgets/variant': {
        extra: 'variant_theme_extra'
    }
}

// classes
{
    '@dojo/widgets/base': {
        root: ['base_classes_root']
    },
    '@dojo/widgets/variant': {
        root: ['variant_classes_root'],
        extra: ['variant_classes_extra']
    }
}
```

This will result in a theme objected constructed that includes the extra root class from `@dojo/widgets/variant` and the extra root class for `@dojo/widgets/base` will be applied in the base widget

4) Supports a known prefix for classes that need to passed as part of the base widgets theme

```tsx
    '@dojo/widgets/base': {
        root: ['base_classes_root']
    },
    '@dojo/widgets/variant': {
        baseRoot: ['variant_prefix_theme_root']
    }
```

When passing `base` as the `prefix` will use the `variant_prefix_theme_root` class as the base widget's `root` class.

When a prefix is specified, only prefixed classes will be composed for the child's theme. It will not pass the class for an `active` class, even if the `active` class exists in the child's css.

[...]

PR #943, opened 2019-12-03
Context menu

### LINKED ISSUE #938: Context menu
We should create a context menu using the existing `Menu` and the `body` tag.
This widget would show it's menu on `right-click` within a given area.
An example of such a context menu can be found [here](https://material-ui.com/components/menus/#context-menu).

Notes:
- The menu widget does not support menu nesting so this would not be offered at this stage
- The menu widget does not currently support `dividers`, an issue is raised for this [here](https://github.com/dojo/widgets/issues/939)

maier49:
**Type:** feature

**Description:**

Resolves #938


dojo-site:
* Docs Deployment: https://dojowidgets-f6rya87dr.now.sh
* Tests Deployment: https://dojowidgets-ankjcdmjh.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-md7ampjjc.now.sh
* Tests Deployment: https://dojowidgets-a66d6js13.now.sh

src/context-menu/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { createICacheMiddleware } from '@dojo/framework/core/middleware/icache';
+import theme from '../middleware/theme';
+import Menu, { MenuOption } from '../menu/index';
+import * as menuCss from '../theme/default/menu.m.css';
+import * as css from '../theme/default/context-menu.m.css';
+import * as fixedCss from './context-menu.m.css';
+
+export interface ContextMenuProperties {
+	options: MenuOption[];
+	onValue(value: string): void;
```
tomdye: As this is a menu with selection, do you think `onSelect` might be better?

src/context-menu/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+	menuWidth?: number;
```
tomdye: Do we really need `menuWidth` here? The `popup` widget uses `dimensions` to figure out the size of the menu where appropriate.

dojo-site:
* Docs Deployment: https://dojowidgets-md7ampjjc.now.sh
* Tests Deployment: https://dojowidgets-a66d6js13.now.sh

src/context-menu/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+				<body>
+					<div key="underlay" classes={fixedCss.underlay} onclick={onClose} />
+					<div
+						key="wrapper"
+						classes={[fixedCss.root, classes.wrapper]}
+						styles={wrapperStyles}
+					>
+						<Menu
+							key="menu"
+							theme={{
```
tomdye: bear in mind this will need to change after https://github.com/dojo/widgets/pull/940

src/examples/src/widgets/context-menu/Basic.tsx
```diff
@@ -0,0 +1,42 @@
...
+							`${text.slice(0, selection.anchorOffset)}${text.slice(
+								selection.focusOffset
+							)}`
+						);
+					}
+				}}
+			>
+				{text}
+			</ContextMenu>
+			<div>Printed Text: {printedText}</div>
```
tomdye: what impact does this usage have on positioning / layout of the chid this is passed?

src/context-menu/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+		wrapperStyles.width = `${menuWidth}px`;
+	}
+
+	return (
+		<virtual>
+			<div
+				key="content"
+				oncontextmenu={(event: MouseEvent) => {
+					event.preventDefault();
+					icache.set('x', event.clientX - 2);
```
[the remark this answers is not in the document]
maier49: I'll update those to be constants, it's just an offset to make it look like the menu's coming off your cursor the same way a real context menu would, though obviously this wouldn't take into account a different sized cursor.

src/context-menu/index.tsx
maier49: Sure, I'll update that.

src/context-menu/index.tsx
maier49: I'm not sure using `dimensions` would make sense for a context menu. In the example the context menu can be applied to any region of text. In general you could be applying the context menu to any kind of content, and unlike the popup which sizes to the specific content it's attached to I would think that the context menu's width should be consistent regardless of what component is clicked on?

src/context-menu/index.tsx
maier49: 👍 I believe I've updated it accordingly

src/examples/src/widgets/context-menu/Basic.tsx
maier49: It wraps it in a single `div` with no additional styles, but I added a `contentWrapper` class to the default theme so that you could style the wrapper however you'd like.

dojo-site:
* Docs Deployment: https://dojowidgets-gpwzuy4vb.now.sh
* Tests Deployment: https://dojowidgets-ak6lcf5ck.now.sh

tomdye:
Looks like focus / mouse hover is still not working on the example

maier49:
@tomdye 
I think I should have all feedback addressed now:
* Added check for whether the menu will fit under the cursor or needs to be placed above it (and initial transparent render until this is determined)
* Using the focus meta to focus the menu when its opened
* Added a border around the menu in the default theme following the example of the select
* Renamed `menuWrapper` class to `wrapper` to not conflict with the prefixed menu classes
* Updated positioning so it works if you're scrolled 🤦‍♂

dojo-site:
* Docs Deployment: https://dojowidgets-7cmibeviw.now.sh
* Tests Deployment: https://dojowidgets-6qblvyjdn.now.sh

tomdye:
Thanks for continuing to work on this @maier49. It seems now that there is a lot of overlap between `Popup` and `ConextMenu` in terms of positioning and space calculation. Do you think it might be better than the further abstract `Popup` into `ContextPopup` / `TriggerPopup` which use a common `Popup` to do those calcs under the hood? That would remove the need for this code duplication and give us ways to popup _anything_ on demand via right click or via a trigger?

dojo-site:
* Docs Deployment: https://dojowidgets-btmjjm7uo.now.sh
* Tests Deployment: https://dojowidgets-cvalttv43.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-bj2ryebu3.now.sh
* Tests Deployment: https://dojowidgets-iqfd04aot.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-mq7zqzhb7.now.sh
* Tests Deployment: https://dojowidgets-ls4ommrkx.now.sh

src/context-menu/index.tsx
```diff
@@ -0,0 +1,45 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import theme from '../middleware/theme';
+import Menu, { MenuOption } from '../menu/index';
+import * as menuCss from '../theme/default/menu.m.css';
+import * as css from '../theme/default/context-menu.m.css';
+import ContextPopup from '../context-popup';
+
+export interface ContextMenuProperties {
+	options: MenuOption[];
```
tomdye: can we get comments on these please?

src/context-menu/context-menu.m.css
```diff
@@ -0,0 +1,13 @@
+.root {
```
tomdye: comments please

src/context-menu/context-menu.m.css
```diff
@@ -0,0 +1,13 @@
...
+	position: absolute;
+	z-index: 11;
+}
+
+.underlay {
```
tomdye: is this still needed? I'd have imagined that the popup deals with the underlay?

src/context-menu/context-menu.m.css
```diff
@@ -0,0 +1,13 @@
+.root {
+	position: absolute;
+	z-index: 11;
```
tomdye: we need to make these z-indexes higher really, i believe we have some guidance on this somewhere

tomdye:
@maier49 btw i updated this PR with latest widgets to benefit from no body scroll when popup is open

src/context-menu/context-menu.m.css
tomdye: https://github.com/dojo/widgets#z-index-layering

src/context-menu/context-menu.m.css
maier49: I've removed these CSS files, as you mentioned in another comment these styles are handled by the popup now and I just failed to remove the unused file

src/context-menu/context-menu.m.css
maier49: Removed

src/context-menu/context-menu.m.css
maier49: I've removed these CSS files, as you mentioned in another comment these styles are handled by the popup now and I just failed to remove the unused file

dojo-site:
* Docs Deployment: https://dojowidgets-jqlpubhk8.now.sh
* Tests Deployment: https://dojowidgets-lemr5azf0.now.sh

PR #945, opened 2019-12-03
ignore time when checking min and max dates

### LINKED ISSUE #857: Calendar widget minDate / maxDate interpreted in a wrong way
The calendar widget compares the min/max date with </> instead of <=/=> so it isn't possible to select the current day if it's set as min date.

devpaul:
**Type:** bug

**Description:**

Updated the Calendar to only take the year, month, and day portions of a date when determining if a day is outside of the min/max dates. Previously it was possible to set a min / max date with a time that would place a day outside of a given range; e.g. if `{ minDate: new Date() }` was used the current day would be outside of the minDate range.

Resolves #857

### DIFF
--- a/src/calendar/date-utils.tsx
+++ b/src/calendar/date-utils.tsx
@@ -13,3 +13,11 @@ export function monthInMax(year: number, month: number, maxDate?: Date) {
 	}
 	return true;
 }
+
+function stripTime(date: Date) {
+	return new Date(date.getFullYear(), date.getMonth(), date.getDate());
+}
+
+export function isOutOfDateRange(dateObj: Date, min?: Date, max?: Date) {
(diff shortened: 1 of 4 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-3bbxq7g9a.now.sh
* Tests Deployment: https://dojowidgets-5mnndrx8a.now.sh

src/calendar/date-utils.tsx
```diff
@@ -13,3 +13,11 @@ export function monthInMax(year: number, month: number, maxDate?: Date) {
 	}
 	return true;
 }
+
+function stripTime(date: Date) {
```
tomdye: do these need to be in a date-utils file when they're only used in a single location?

PR #946, opened 2019-12-04
Added `yearBasis` to DatePicker

### LINKED ISSUE #858: Configurable BASE_YEAR in calendar/date-picker
**Enhancement**

Package Version: master

In DatePicker.ts, there's BASE_YEAR hardcoded to 2000. This makes the DatePicker always start at 2000, even if next year is 2020 (the last year by default in that list). It would be nice so we can configure that, so we don't go that back in the past.

Additionally, there's a year range in DatePicker.ts, but never used in Calendar, so that's also not a solution.

devpaul:
**Type:** feature

**Description:**

Adds a `yearBasis` property to `Calendar` and `DatePicker` to allow better positioning of the year in the picker.

Resolves #858


dojo-site:
* Docs Deployment: https://dojowidgets-764l08qw7.now.sh
* Tests Deployment: https://dojowidgets-kwfuca0ct.now.sh

src/calendar/DatePicker.tsx
```diff
@@ -58,10 +59,11 @@ export interface DatePickerProperties extends ThemedProperties {
 	onRequestYearChange?(year: number): void;
 	renderMonthLabel?(month: number, year: number): string;
 	year: number;
+	yearBasis?: number;
```
tomdye: How about naming this `initialYear`?

src/calendar/DatePicker.tsx
```diff
@@ -105,10 +107,10 @@ export class DatePicker extends ThemedMixin(WidgetBase)<DatePickerProperties> {
 	}
 
 	private _getYearRange() {
-		const { year, yearRange = 20 } = this.properties;
-		const offset = ((year - BASE_YEAR) % yearRange) - yearRange * this._yearPage;
+		const { year, yearRange = 20, yearBasis = DEFAULT_YEAR_BASIS } = this.properties;
```
tomdye: How about defaulting it to the current year?

src/calendar/index.tsx
```diff
@@ -55,6 +55,8 @@ export interface CalendarProperties extends ThemedProperties, I18nProperties {
 	weekdayNames?: { short: string; long: string }[];
 	/** Set the currently displayed year */
 	year?: number;
+	/** The starting year used in numbering for the year popup (defaults to 2000) */
+	yearBasis?: number;
```
tomdye: as above, perhaps `initialYear`, defaulted to the current year

src/calendar/DatePicker.tsx
devpaul: Defaulting the `yearBasis` (or `initialYear`) to the current year would mean that the first year show on a page started with a year +/- the `yearRange`. I think that works well if the date is set to the current year, otherwise the page numbering would seem weird.

i.e. For this year the pages would be:

..., [1999 ... 2018 ], [2019 ... 2038], [2039 ... 2048], ...

image

The `yearBasis` is treated like an offset in starting the year numbering. I would prefer that it was a multiple of 10. I think most people are used to seeing numbers line up in that way.

src/calendar/DatePicker.tsx
devpaul: The `yearBasis` is used as an offset to inform the year drop-down how to number the pages. Maybe `yearPageOffset` or `yearNumberingOffset` or `initialYearOffset` and set the default value to 0?

src/calendar/DatePicker.tsx
tomdye: perhaps needs to be the nearest 10 within 10 years of the current year selected then? that would keep the selected value towards the middle. Also, do we really want to expose this to the user via a property? I'd imagine that this would be better kept internal?

devpaul:
Closing this and deciding to not implement. Exposing too many properties to customize `DatePicker` and `Calendar` will take away from the simplicity of the widget. The Calendar currently behaves like other implementations (react-calendar and angular material). We decided to not change the page numbering based on the year selected as that would create inconsistent UX between uses and calendars.

Hope this helps anyone who might need this functionality. Recommendation for developers that need advanced customization is to create their own Calendar widget based on Dojo's. It's better to break the dependency on Dojo and create a custom solution when your needs are more complex.

PR #990, opened 2019-12-30
Material Theme for Text Input and Helper Text

### LINKED ISSUE #953: Material theme for text input
material Text input details can be found here: https://material.io/develop/web/components/input-controls/text-field/.
We need to consider reuse for textarea when doing this work.
Our existing dom structure should support the general styles along with leading / trailing sections. We may be unable to support the floating labels and thus may need to style them with appropriate material typography - this has been covered in the label issue.

### LINKED ISSUE #955: Material theme for helper text
Helpertext is used in many different widgets, we may need to adjust it's individual usage. the material spec for this is here: https://material.io/develop/web/components/input-controls/text-field/helper-text/.

This has already been done elsewhere.

nicknisi:
**Type:** feature

**Description:**

This PR adds a Material theme for text input by composing the styles from @material/text-field. In order to properly do this, the structure of the text input had to be modified slightly to put the `HelperText` widget outside of the component's root in order for the styles to work for both the Dojo and Material themes.

**Note:** Because the internal structure of the text input changed, I also changed this component over to use TSX.

Resolves #953
Resolves #955


dojo-site:
* Docs Deployment: https://dojowidgets-b07ssq8pj.now.sh
* Tests Deployment: https://dojowidgets-hiw3b5twd.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-grlw7ycew.now.sh
* Tests Deployment: https://dojowidgets-jt6aypwdf.now.sh

src/text-input/index.tsx
```diff
@@ -252,91 +235,110 @@ export class TextInput extends ThemedMixin(FocusMixin(WidgetBase))<TextInputProp
...
+						<Label
+							theme={theme}
+							disabled={disabled}
+							valid={valid}
+							focused={focus.containsFocus}
+							readOnly={readOnly}
+							required={required}
+							hidden={labelHidden}
+							forId={widgetId}
+							classes={
```
tomdye: this should really be done via `theme` and `theme.compose` with a prefix of `label`.

tomdye:
Could this have been achieved by moving the label within the wrapper rather than adding an extra wrapper? Alternatively could this be done via padding , translate x/y on the label to avoid changing structure?

nicknisi:
Unfortunately, I don't think so. Material needs a wrapper that includes the label and text input, but not the helper text. iirc, the label is then absolutely positioned within that wrapper.

I tried a couple of different things to get it to work but unfortunately I don't think it's possible to both use the material css and not change the structure here.

dojo-site:
* Docs Deployment: https://dojowidgets-m8fc4r6av.now.sh
* Tests Deployment: https://dojowidgets-lw7euzad0.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-m8fc4r6av.now.sh
* Tests Deployment: https://dojowidgets-lw7euzad0.now.sh

PR #1024, opened 2020-01-13
Add body scroll middleware and use on popup

### LINKED ISSUE #1023: Popups and Dialogs should disable body scroll
## Enhancement** 

Currently our popup and dialog use an underlay to prevent interaction with the underlying webpage but this does not prevent body scrolling via the mouse-wheel, scroll bar. This is particularly evident when placing popups such as a `Select` within an absolute or fixed position parent. 
When using a native select style component, the body scrollbar is automatically disabled, this is the behaviour we wish to emulate.

### Planned approach

The simplest / cleanest way to implement this is by setting `overflow: hidden` on the `body` tag. Thus preventing the scroll bar. When the popup is hidden, the scrolling will be re-enabled.

### Drawbacks

The known drawback of this approach is that it does not work for ios safari. However, this is a small subset of our users and more importantly; we would advise that mobile applications use native implementations of selects / date pickers etc as they are optimised for mobile already and do not have this issue.

tomdye:
**Type:** feature

**Description:**

PR to disable body scroll via a middleware when using the popup.
Example of usage can be seen in the select demo.

Resolves #1023

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,29 +5,30 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
-      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
+      "version": "7.8.0",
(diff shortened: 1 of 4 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-4cueet46u.now.sh
* Tests Deployment: https://dojowidgets-6yoftl7a5.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-8fyu9v970.now.sh
* Tests Deployment: https://dojowidgets-k9m600978.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-80dc7l3fv.now.sh
* Tests Deployment: https://dojowidgets-4dwoelx2w.now.sh

PR #1029, opened 2020-01-14
make onSubmit optional for the form

### LINKED ISSUE #1021: Form should not require `onSubmit` property
## Bug 

Currently the `Form` widget requires an `onSubmit` property to be passed. I do not believe this is correct as there may be occasions when we simply wish to pass the values back to a parent widget and not perform a submit. 

This could easily be the case in a wizard / step by step flow with multiple forms gathering answers. In this case, `onSubmit` is not appropriate.

We should therefore make this an optional property.

tomdye:
**Type:** enhancement

**Description:**

Makes `onSubmit` optional for the form

Resolves #1021

### DIFF
--- a/src/form/index.tsx
+++ b/src/form/index.tsx
@@ -12,7 +12,7 @@ interface FormProperties {
 	/** The initial form value */
 	initialValue?: FormValue;
 	/** Callback for when the form is submitted with valid values */
-	onSubmit(values: FormValue): void;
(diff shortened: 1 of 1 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-c3kgyckgq.now.sh
* Tests Deployment: https://dojowidgets-g7w2c3h62.now.sh

PR #1033, opened 2020-01-16
Login form

### LINKED ISSUE #1012: Login Form widget
# Login Widget

The login widget should be a basic login form with username / password inputs and submit / cancel buttons.
This should use form / password input etc and should prove out the implementation of these widgets.

The login / submit functionality should be able to use a browser post as well as an XHR request to submit details.

Validation of username / password being entered and login button being enabled should be handled by the login form widget.

tomdye:
**Type:** bug / feature

**Description:**

Login form widget using form and password-input widgets

Resolves #1012


dojo-site:
* Docs Deployment: https://dojowidgets-ptz0lm8qv.now.sh
* Tests Deployment: https://dojowidgets-q9yfijbsp.now.sh

src/examples/src/widgets/login-form/ActionForm.tsx
```diff
@@ -0,0 +1,12 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import LoginForm from '@dojo/widgets/login-form';
+
+const factory = create();
+
+export default factory(function ActionForm() {
+	return (
+		<virtual>
```
agubler: Why is this in a `virtual`?

src/login-form/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+		const { action, initialValue, method } = props;
+		formProps = {
+			action,
+			method,
+			initialValue
+		};
+	}
+
+	const [
+		{ forgotPassword, register } = { forgotPassword: undefined, register: undefined }
```
agubler: Do you have to specify `undefined` here?

src/login-form/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+	] = children();
+
+	const { messages } = i18n.localize(bundle);
```
agubler: How does a user provide additional nls? perhaps an example would be good...

src/login-form/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+const factory = create({ theme, i18n })
+	.properties<LoginFormProperties>()
+	.children<LoginFormChildren>();
+
+const LoginForm = factory(function LoginForm({
+	properties,
+	children,
+	middleware: { theme, i18n }
+}) {
+	const classes = theme.classes(css);
```
agubler: Can we call this `themedCss` or something... I believe that we've used the same in other widgets

src/login-form/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+	let formProps: FormProperties = {};
+	const props = properties();
+
+	if (isSubmitForm(props)) {
+		const { onSubmit, initialValue } = props;
+		formProps = {
+			onSubmit,
+			initialValue
+		};
+	} else if (isActionForm(props)) {
```
agubler: Presume this can be an `else` without need to use `isActionForm`?

dojo-site:
* Docs Deployment: https://dojowidgets-2535i9tl5.now.sh
* Tests Deployment: https://dojowidgets-fq2xqzsoo.now.sh

src/login-form/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+									name="username"
+									required={username.required()}
+									value={username.value()}
+									valid={username.valid()}
+									onValue={username.value}
+									onValidate={username.valid}
+								/>
+							</div>
+							<div classes={classes.field}>
+								<PasswordInput
```
agubler: Do you think the password requirements should be customisable for the login widget?

agubler: As well as this, do you think that we need to expose a way to specify requirements for the username?

src/examples/src/widgets/login-form/ActionForm.tsx
tomdye: not needed anymore

src/login-form/index.tsx
tomdye: the children are optional so it could be `[ undefined ]` rather than `[{ forgotPassword, register }]`, TS complained until i added the undefined

src/login-form/index.tsx
tomdye: do we have examples of this elsewhere?

src/login-form/index.tsx
tomdye: sure

src/login-form/index.tsx
tomdye: sure

src/login-form/index.tsx
tomdye: i do not believe so. this is for entering a username / password not for creating one. If / when we create a `register` form we would certainly want to allow rules to be passed for these. I've never seen a login for before that gave you hints as to the password requirements

dojo-site:
* Docs Deployment: https://dojowidgets-lfszzszsh.now.sh
* Tests Deployment: https://dojowidgets-bxteileh2.now.sh

PR #1037, opened 2020-01-17
Fixing material dialog animation

rorticus:
**Type:** bug / feature

**Description:**

There was a problem with the dialog exit animation using the material theme. The material theme had a `display: none` on the root of the dialog, so as soon as the dialog would close, the parent would be hidden, and thus the exit animation could never start, and therefore finish. This would leave the vdom in a weird state in that it was keeping the old nodes around waiting for the animation to finish.

### DIFF
--- a/src/theme/material/dialog.m.css
+++ b/src/theme/material/dialog.m.css
@@ -2,11 +2,11 @@
 
 /* The root class for the Dialog */
 .root {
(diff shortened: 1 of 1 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-4dni66hln.now.sh
* Tests Deployment: https://dojowidgets-aaacf5tlm.now.sh

PR #1092, opened 2020-02-07
Stop body scrolling whilst dialog open

### LINKED ISSUE #669: Stop the background scrolling while the dialog is opened
**Bug** 

I can't stop the background scrolling if  I'm using the dojo Dialog. I tried to use event.stopPropagation(), but the propagation didn't stopped. I found a workaround, but unfortunatly it's not the best solotuin because I'm modifying the global css. Is there any better solution to stop the background scrolling? 

Package Version: "@dojo/widgets": "5.0.0"

**Code**
Is there any way to stop the propagation without modifying the global css? (document.body.classList)
https://codesandbox.io/s/l4nmr06y0m

**Expected behavior:**
I shouldn't be able to scroll in the background (without modifying the global css)

**Actual behavior:**

I'm able to scroll in the background while the dialog is opened

tomdye:
**Type:** bug / feature

**Description:**

Uses the body scroll middleware with dialog

Resolves #669

### DIFF
--- a/src/dialog/index.tsx
+++ b/src/dialog/index.tsx
@@ -6,6 +6,7 @@ import { create, tsx } from '@dojo/framework/core/vdom';
 import { formatAriaProperties, Keys } from '../common/util';
 import Icon from '../icon';
 import theme, { ThemeProperties } from '../middleware/theme';
+import bodyScroll from '../middleware/bodyScroll';
 import * as css from '../theme/default/dialog.m.css';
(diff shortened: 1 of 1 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-lzqnjrfe8.now.sh
* Tests Deployment: https://dojowidgets-ebmd92fj4.now.sh

PR #1099, opened 2020-02-17
Prefer base properties over pick/exclude

tomdye:
This PR changes existing use of Pick / Exclude to use composition from base properties instead.


dojo-site:
* Docs Deployment: https://dojowidgets-whzjnq7s0.now.sh
* Tests Deployment: https://dojowidgets-gazpp4996.now.sh

PR #1103, opened 2020-02-17
Set page size on preFetch so it's ready for initial filter

### LINKED ISSUE #1083: First page of paginated grid does not load
## Bug

The first page of the paginated grid does not appear to load. 
There are no messages in the console, this will need investigating.
Can be seen here: https://widgets.dojo.io/#widget/grid/example/paginated

maier49:
**Type:** bug

On the first render the `onValue` of the filter textbox is triggered with `""`. This triggered the filter process before the fetch process resolved so the `pageSize` had not been set. Moving the setting of `pageSize` to the `preFetcher` makes it available for the first filter command. 

**Description:**

Resolves #1083

### DIFF
--- a/src/grid/processes.ts
+++ b/src/grid/processes.ts
@@ -25,12 +25,13 @@ const pageChangeCommand = commandFactory<PageChangeCommandPayload>(
 );
 
 const preFetcherCommand = commandFactory<FetcherCommandPayload>(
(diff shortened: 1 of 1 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-ofmtel0k1.now.sh
* Tests Deployment: https://dojowidgets-bbje7eabb.now.sh

PR #1106, opened 2020-02-17
Add name property to Form

### LINKED ISSUE #1044: Form needs `name` property
**Enhancement**

In order to provide meaningful value when performing action submissions and to assist with returning sensible values when using nested forms, we should add a `name` property to the `Form`.

maier49:
**Type:** feature

**Description:**
Adds a `name` property to the form.

Resolves #1044

### DIFF
--- a/src/examples/src/widgets/form/Basic.tsx
+++ b/src/examples/src/widgets/form/Basic.tsx
@@ -23,7 +23,10 @@ const App = factory(function({ middleware: { icache } }) {
 
 	return (
 		<virtual>
-			<Form onValue={(values) => icache.set('basic', { ...icache.get('basic'), ...values })}>
+			<Form
+				name="basicForm"
+				onValue={(values) => icache.set('basic', { ...icache.get('basic'), ...values })}
+			>
 				{({ field }: FormMiddleware<Fields>) => {
 					const firstName = field('firstName');
 					const middleName = field('middleName');
--- a/src/form/index.tsx
+++ b/src/form/index.tsx
@@ -15,6 +15,8 @@ interface BaseFormProperties extends ThemeProperties {
 	initialValue?: FormValue;
 	/** Callback called when a form value changes */
 	onValue?(values: FormValue): void;
+	/** The name property of the form */
+	name?: string;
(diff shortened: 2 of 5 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-rjtbi969n.now.sh
* Tests Deployment: https://dojowidgets-ngg6eytub.now.sh

src/form/index.tsx
```diff
@@ -69,6 +71,13 @@ export default factory(function Form({
 		classes: themedCss.root
 	};
 
+	if (props.name) {
```
tomdye: i think this can just be used as-is on the form render, if it's undefined it'll be ignored by the renderer

KaneFreeman: You are correct, just tested this out. This if check unnecessary.

src/examples/src/widgets/form/Basic.tsx
```diff
@@ -23,7 +23,10 @@ const App = factory(function({ middleware: { icache } }) {
 
 	return (
 		<virtual>
-			<Form onValue={(values) => icache.set('basic', { ...icache.get('basic'), ...values })}>
+			<Form
+				name="basicForm"
```
KaneFreeman: Since this is mean to help with action forms, and more specifically nested forms, maybe we should either:

1. Add `name` to an action form example.
2. Create a new example for nested forms with the `name` property.

Or maybe both.

tomdye: i think both :)

maier49: I added a name to a new action form example. I didn't add a nested form example since it's [not supported by the spec](https://html.spec.whatwg.org/multipage/forms.html#the-form-element)

src/form/index.tsx
maier49: 👍

dojo-site:
* Docs Deployment: https://dojowidgets-otukgjahc.now.sh
* Tests Deployment: https://dojowidgets-23oagh04l.now.sh

PR #1105, opened 2020-02-17
Card Improvements

nicknisi:
**Type:** feature

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code matches the [style guide](https://github.com/dojo/framework/blob/master/STYLE.md)
* [x] Unit tests are included in the PR
* [ ] For new widgets, an entry has been added to the `.dojorc`
* [ ] Any widget variant uses `theme.compose` like [this](https://github.com/dojo/widgets/issues/847)
* [x] WidgetProperties are exported

**Description:**

This PR fixes the card implementation, simplifying its use by taking advantage of named child properties. 

This is part of the work for improving the card implementation in #978.

**Properties**

- `onAction?: () => void` - an optional action handler for the card content. This adds the primary action styles to the card when provided and is called back when the card's content is clicked on
- `mediaSrc?: string` - An optional path to an image that when provided will populate the media section within the card's main content area
- `mediaTitle?: string` - An optional string property that sets the media's `title` property and will be displayed when the media is hovered over
- `square?: boolean` - When provided, the square class will be applied, rendering the media with a 1:1 ratio
- `title?: string` - When provided adds a default and consistently-styled title section to the card's content
- `subtitle?: string` - When provided adds a default and consistently-styled subtitle section to the card's content

**Named Children**
All children are optional and have the type `() => RenderResult`.

- `header` - The header section goes above the main content section
- `content` - The main content section. If `mediaSrc`, `title`, or `subtitle` are provided, they will appear above the contents of this section
- `actionButtons` - The action buttons to render on the left side of the card
- `actionIcons` - The action icons to render on the right side of the card

**Example**
```tsx
<Card
	onAction={() => {}}
	mediaSrc={mediaSrc}
	title="Hello, World"
	subtitle="A pretty picture"
>
	{{
		header: () => <div>Header Content</div>,
		content: () => <p>Travel the world today.</p>,
		actionButtons: () => (
			<virtual>
				<Button>Read</Button>
				<Button>Bookmark</Button>
			</virtual>
		),
		actionIcons: () => (
			<virtual>
				<Icon type="secureIcon" />
				<Icon type="downIcon" />
				<Icon type="upIcon" />
			</virtual>
		)
	}}
</Card>
```

**Media Card**
image

**Card with actions**
image

**Card with header, media, content, and actions**
image

PR #1104, opened 2020-02-17
Add Avatar widget

nicknisi:
**Type:** feature

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code matches the [style guide](https://github.com/dojo/framework/blob/master/STYLE.md)
* [x] Unit tests are included in the PR
* [x] For new widgets, an entry has been added to the `.dojorc`
* [x] Any widget variant uses `theme.compose` like [this](https://github.com/dojo/widgets/issues/847)
* [x] WidgetProperties are exported

**Description:**

This is part of ongoing work to improve the card implementation at #978.

The Avatar has the following properties:

- `size?: 'small' | 'medium' | 'large'` - The size of the avatar (defaults to 'medium')
- `variant?: 'square' | 'circle' | 'rounded' - The style of the avatar (defaults to 'circle')
- `secondary?: boolean` - Whether to render with secondary styles (as determined by the theme)
- `src?: string` - the path to an image file to use within the Avatar. If this is set the children are ignored
- `alt?: string` The alt text to add to the image

**Code example**
```tsx
<Avatar src="foo.png"/> // Avatar with image
<Avatar>D</Avatar> // Avatar with character text
<Avatar variant="square">D</Avatar> // square Avatar
<Avatar secondary>D</Avatar> // Avatar with secondary styling
<Avatar><Icon type="secureIcon"/></Avatar> // Avatar with Icon
```

**Avatar with image**
image

**Avatar with text**
image

**Avatar with icons**
image

**Avatar sizes**
image

PR #1107, opened 2020-02-17
Add Header Card widget

### LINKED ISSUE #978: Improve Card implementation
## Enhancement 

Current card implementation requires the user to use the card css classes from the default theme to apply styling / positioning to a card's child nodes. 
This is not an ideal solution as it does not match patterns used elsewhere in the dojo widgets nor does it allow theming to work appropriately.

We should come up with a better way to do this.
Possible solutions are as follows:

### render props / child render props

This would provide the users with props they can pass to render their content in the correct position.
ie.
```tsx
<Card primaryContent={renderPrimary} secondaryContent={renderSecondary} />
```
The down side of this is that is gives the user no control over where they place each of these render results. 

### Child widgets

This would provide the users with an assortment of smaller widgets they can use to render styled content. ie.
```tsx
<Card>
   <CardPrimary>primary stuff</CardPrimary>
   <CardButtons>{...buttons}</CardButtons>
</Card>
```
This would give users greater flexibility over where they place their children but removes the ability for our card widget to be opinionated about its children's layout / positioning which may be important for some layouts.

### Card derivatives with render / child render props

This, I feel, gives us the best all-around flexibility and control as we can specify different card types, ie. `MediaCard` / `PrimaryCard` / `SubtitleCard` and with such cards, provide the users with a specific, typed childen render object which they can use.
This maintains the ability to be opinionated about layout and removes the need to use theme classes from card. It also encourages users to create their own card templates etc which they would use within their own apps.

```tsx
<MediaCard imgSrc="imgSrc">
{{ 
   title: () => "I am the title",
   buttons: () => <RaisedButton>hello</RaisedButton>
 }}
</MediaCard>
```

```tsx
<SubtitleCard>
{{
   title: () => "card title",
   subtitle: () => "card sub title"
}}}
</SubtitleCard>
```

nicknisi:
**Type:** feature

**Note:** This PR relies on #1105 and #1104 being merged. Please review those first.

**Description:**

This adds a new card, a header card, that wraps the card implementation and takes over rendering of the card's `header` named child to provide a consistently-styled header for the card. The header-card takes all the same properties and children (minus `header` named child). It takes over rendering the `title` and `subtitle` if provided and will render them in the header (and prevent them from being passed to the underlying card). It also adds an `avatar` property which will render an avatar in the header next to the title/subtitle.

Resolves #978 

**Example**
```tsx
<HeaderCard
	avatar={() => <Avatar src={avatar} />}
	title="Hello, World"
	subtitle="Lorem ipsum"
	mediaSrc={mediaSrc}
>
	{{
		content: () => <p>Lorem ipsum</p>,
		actionButtons: () => <Button>Action</Button>,
		actionIcons: () => (
			<virtual>
				<Icon type="upIcon" />
				<Icon type="downIcon" />
			</virtual>
		)
	}}
</HeaderCard>
```
image

### DIFF
--- a/src/header-card/index.tsx
+++ b/src/header-card/index.tsx
@@ -1,0 +1,48 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import theme from '../middleware/theme';
+import Card, { CardProperties, CardChildren } from '../card';
+import { RenderResult } from '@dojo/framework/core/interfaces';
+import * as css from '../theme/default/header-card.m.css';
+
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
+export interface HeaderCardProperties extends CardProperties {
+	title: string;
+	avatar?: () => RenderResult;
+}
+
+export interface HeaderCardChildren extends Omit<CardChildren, 'header'> {}
+
+const factory = create({ theme })
+	.properties<HeaderCardProperties>()
+	.children<HeaderCardChildren>();
+
+export const HeaderCard = factory(function HeaderCard({
+	middleware: { theme },
+	properties,
+	children
+}) {
+	const themeCss = theme.classes(css);
+	const { title, subtitle, avatar, ...cardProps } = properties();
+	const cardChildren = children()[0];
+	return (
+		<Card {...cardProps}>
+			{{
+				header: () => (
+					<virtual>
+						<div classes={themeCss.header}>
+							{avatar && <div classes={themeCss.avatar}>{avatar()}</div>}
+							<div classes={themeCss.headerContent}>
+								{<h2 classes={themeCss.title}>{title}</h2>}
+								{subtitle && <h3 classes={themeCss.subtitle}>{subtitle}</h3>}
+							</div>
+						</div>
+					</virtual>
+				),
+				...cardChildren
+			}}
+		</Card>
+	);
+});
+
+export default HeaderCard;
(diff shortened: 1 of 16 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-cvwrgwsaw.now.sh
* Tests Deployment: https://dojowidgets-qt63tjh27.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-n58k0wgdu.now.sh
* Tests Deployment: https://dojowidgets-1p5cmyzwm.now.sh

src/header-card/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+	.children<HeaderCardChildren | undefined>();
+
+export const HeaderCard = factory(function HeaderCard({
+	middleware: { theme },
+	properties,
+	children
+}) {
+	const themeCss = theme.classes(css);
+	const { title, subtitle, avatar, ...cardProps } = properties();
+	const cardChildren = children()[0];
```
agubler: Could this be `const [ cardChildren ] = children();`

tomdye:
```suggestion
	const [ cardChildren ] = children();
```

dojo-site:
* Docs Deployment: https://dojowidgets-qef3ylg2b.now.sh
* Tests Deployment: https://dojowidgets-1epjlfke5.now.sh

PR #1155, opened 2020-03-18
#1224 Breadcrumb widget

### LINKED ISSUE #1124: Breadcrumb Widget
Please put design / thoughts on this ticket.
We should create a breadcrumb widget drawing inspiration from:

- https://material-ui.com/components/breadcrumbs/
- https://ant.design/components/breadcrumb/

We should keep this as simple as possible and not get tied up on routing etc.

mwistrand:
**Type:** feature

**Description:**

Resolves #1124 

Breadcrumb widget with examples: Basic, Custom Separator, and Custom Renderer.


dojo-site:
* Docs Deployment: https://dojowidgets-gfcx6h504.now.sh
* Tests Deployment: https://dojowidgets-28gfqssgf.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-6z1bkzt2j.now.sh
* Tests Deployment: https://dojowidgets-i8v8sy53v.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-hj91d6ja8.now.sh
* Tests Deployment: https://dojowidgets-1an9m0jr9.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-mm633ddam.now.sh
* Tests Deployment: https://dojowidgets-qoas67pn3.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-cta0dzi6e.now.sh
* Tests Deployment: https://dojowidgets-nqy7gpy6d.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-l3sxb7kxi.now.sh
* Tests Deployment: https://dojowidgets-icowvkj3e.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-ldaawl85u.now.sh
* Tests Deployment: https://dojowidgets-b2ccualml.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-1ypeeby5b.now.sh
* Tests Deployment: https://dojowidgets-bxv4b6txg.now.sh

src/examples/src/widgets/breadcrumb/Basic.tsx
```diff
@@ -0,0 +1,10 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Breadcrumb from '@dojo/widgets/breadcrumb';
+
+const factory = create();
+
+const App = factory(function() {
+	return <Breadcrumb current="step" item={{ label: 'Step 1' }} />;
```
tomdye: can the item properties be changed to `BreadcrumbProperties` ?
ie. `<Breadcrumb current='step' label='step1' />` etc?
When rendering them you can spread the item over the widget. ie. `<Breadcrumb current='step' {...item} />`

src/theme/material/breadcrumb-group.m.css
```diff
@@ -0,0 +1,18 @@
+@import './variables.css';
+
+.root {
```
tomdye: can you apply the mdc font / sizes / colours here please

src/breadcrumb-group/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+	middleware: { theme }
+}) {
+	const { items, label } = properties();
+	const themeCss = theme.classes(css);
+
+	const defaultRenderer: BreadcrumbGroupChildren = (items: BreadcrumbItem[]) => {
+		const lastIndex = items.length - 1;
+
+		return (
+			<ol classes={[fixedCss.listFixed, themeCss.list]}>
```
tomdye: i think this would work a lot better if the `ol` was rendered outside of the renderer and the outer nodes of the `Breadcrumb` widgets were `li`. Then users would always benefit from the ordered list semantics and the custom renderers etc would look a lot cleaner / simpler.

mwistrand: Ok. I hadn't done it that way because it requires the user to know the internals of `<BreadcrumbGroup` and `<Breadcrumb>` to ensure they use them correctly to end up with semantically-correct markup, but I can add a note to the README to that effect.

mwistrand: The reason I did not do it that way is because it requires the user to know the internals of both widgets in order to end up with semantically-correct markup, and it effectively makes `<Breadcrumb>` useless since it would only be used by `<BreadcrumbGroup>`. If a user needs a custom renderer, then chances are they want control over what is rendered inside the `<li>` and would therefore need to roll their own breadcrumb item instead of using `<Breadcrumb>`.

src/breadcrumb/index.tsx
```diff
@@ -0,0 +1,41 @@
+import theme from '@dojo/framework/core/middleware/theme';
+import { create, tsx } from '@dojo/framework/core/vdom';
+
+import * as css from '../theme/default/breadcrumb.m.css';
+
+export interface BreadcrumbItem {
+	[key: string]: any;
```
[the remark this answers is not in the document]
mwistrand: This was intended for custom renderers. If additional options were needed for each item to render a custom breadcrumb, then users could still use the same `items` property on the containing `<BreadcrumbGroup>`.

src/breadcrumb-group/index.tsx
tomdye: I get what you're saying @mwistrand , but firstly, I don't think `Breadcrumb` will be used outside of this widget and secondly, without moving the `ol` / `li` inside their widgets we won't be able to handle adding any of the fixed css styling or aria tags for them which is an important part of the widget in my opinion. We can provide them with a `BreadCrumbSeperator` widget too if that helps which simply applies `aria` and returns an `li` with the appropriate contents. This would give us the most flexible, user friendly approach.

PR #1156, opened 2020-03-18
Pagination widget

### LINKED ISSUE #1123: Pagination widget
TBD

jellison:
**Type:** feature

**Description:**

Pagination widget

image

Resolves #1123


src/pagination/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+interface PaginationCache {
+	currentPage: number;
+}
+
+const icache = createICacheMiddleware<PaginationCache>();
+const factory = create({ theme, icache, i18n }).properties<PaginationProperties>();
+
+const PAGE_LINKS = 3;
```
tomdye: Do you think this should be customisable? 3 seems a little small, especially if the widget is given a lot more space. Could you use `dimensions` / `resize` middleware perhaps to react to the space available and render what will fit?

src/pagination/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export default factory(function Pagination({ middleware: { theme, icache, i18n }, properties }) {
+	const { initialPage, onPageChange, total } = properties();
+	const classes = theme.classes(css);
+	const currentPage = icache.getOrSet('currentPage', initialPage || 1);
```
tomdye: I believe you'll need to use a `diffProperty` on the `initialPage` value as we do with `initialValue` elsewhere to ensure that it _can_ still be controlled via this property if required.

src/pagination/index.tsx
jellison: Everything is possible. ;)

jellison: Maybe up to some maximum?

jellison: Moved off to issue #1239

jellison:
Issue exists under certain controlled scenarios.

src/pagination/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+					)}
+				{pageSizes && pageSizes.length > 0 && (
+					<div classes={classes.selectWrapper}>
+						<Select
+							key="page-size-select"
+							transform={defaultTransform}
+							initialValue={pageSize.toString()}
+							resource={{
+								resource: () => createResource(memoryTemplate),
+								data: pageSizes.map((ps) => ({ value: ps.toString() }))
```
tomdye, line 161: this data array should be calculated only when `pageSizes` changes and stored in an `icache` otherwise this will be a new array ever render and cause the resource to be re-populated every time.

src/pagination/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+export interface PaginationProperties {
+	/** The initial page number */
+	initialPage?: number;
+
+	/** The initial page size */
+	initialPageSize?: number;
+
+	/** Callback fired when the page number is changed */
+	onPageChange(page: number): void;
```
tomdye, line 34: How about renaming these callbacks to `onPage` and `onPageSize` as we've done out best to drop `onChange` etc from elsewhere.

src/pagination/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+	/** Callback fired when the page size is changed */
+	onPageSizeChange?(pageSize: number): void;
+
+	/** Page size options to display in the page size selector */
+	pageSizes?: number[];
+
+	/** Custom renderer for page size selector item */
+	pageSizeRenderer?(pageSize: number): RenderResult;
```
tomdye, line 43: I'm not sure how valid this is as a use case. I think that 99.9% of the time we'd just want to see the page numbers in the drop down, not even `30 / page`, just `10, 20, 30` etc...

src/pagination/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+	const page = icache.getOrSet('currentPage', 1);
+	const pageSize = icache.getOrSet('pageSize', 10);
+
+	const pageLink = function(
+		page: number,
+		content?: WNode | WNode[],
+		extraClasses: string = classes.numberedLink,
+		key?: string
+	): WNode {
```
tomdye, line 94: You don't really need to type this return value as it is inferred.

src/pagination/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+		return (
+			<button
```
tomdye, line 96: How about moving this inline rather than having it in it's own function. I think that would make it clearer.

jellison:
Render issue has been resolved by dojo/framework#712

PR #1159, opened 2020-03-18
Rate widget

### LINKED ISSUE #1127: Star rating widget
Please put design / thoughts on this ticket.
We should create a star rating widget drawing inspiration from:

- https://ant.design/components/rate/
- https://material-ui.com/components/rating/

We should keep the design as simple as possible. The widget should be uncontrolled.

pottedmeat:
**Type:** feature

**Description:**

Resolves #1127


dojo-site:
* Docs Deployment: https://dojowidgets-fo04ykiim.now.sh
* Tests Deployment: https://dojowidgets-3fdd216gl.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-dgg03bjxz.now.sh
* Tests Deployment: https://dojowidgets-8a20h63rx.now.sh

src/radio/index.tsx
```diff
@@ -15,7 +16,7 @@ export interface RadioProperties extends ThemeProperties, FocusProperties {
 	/** Set the disabled property of the control */
 	disabled?: boolean;
 	/** Adds a <label> element with the supplied text */
-	label?: string;
+	label?: DNode;
```
tomdye: If anything, this should be `RenderResult`, we need to re-visit what we're going to do for these kind of labels

src/rate/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+	name: string;
+	/** The label to be displayed in the legend */
+	label?: string;
+	/** The id used for the radio input elements */
+	widgetId?: string;
+	/** The initial rating value */
+	initialValue?: number;
+	/** Callback fired when the rating value changes */
+	onValue?: (value?: number, mixedValue?: MixedNumber) => void;
+	steps?: number;
```
tomdye: can you add comments here please

src/rate/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+	/** The name attribute for this rating group */
+	name: string;
+	/** The label to be displayed in the legend */
+	label?: string;
+	/** The id used for the radio input elements */
+	widgetId?: string;
+	/** The initial rating value */
+	initialValue?: number;
+	/** Callback fired when the rating value changes */
+	onValue?: (value?: number, mixedValue?: MixedNumber) => void;
```
tomdye: is this `mixedValue` really required? It does not line up with other `onValue` functions that pass only a single value.

src/rate/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+	.properties<RateProperties>()
+	.children<RateChildren | undefined>();
+
+export const Rate = factory(function Rate({
+	children,
+	properties,
+	middleware: { focus, i18n, icache, theme }
+}) {
+	const { format } = i18n.localize(bundle);
+	const {
```
tomdye: could you just use `const themedCss = theme.classes(css)` as this is what we do elsewhere and keeps things clean and consistent. It's hard to reason with the source of the classes when they have been deconstructed like this.

src/rate/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+		disabled: disabledClasses,
+		readOnly: readOnlyClasses,
+		over: overClasses,
+		labelRoot,
+		characterWrapper,
+		filled,
+		partialCharacter,
+		partialRadio
+	} = theme.classes(css);
+	const _uuid = icache.getOrSet('uuid', uuid());
```
tomdye: if this needs an `id` at all, please can you use the `id` from the factory. it's just an integer so you'll need to do something like `const idBase = 'rate-${id}'` as we have elsewhere.

src/rate/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+		integers.push(integer);
+		for (const step of steps) {
+			options.push({
+				value: `${integer}-${step}`,
+				label: format('starLabels', mixedNumber(integer, step, stepsLength))
+			});
+		}
+	}
+
+	const onFocus = () => icache.set('focused', true);
```
tomdye: I think that these may be clearer inline

src/rate/index.tsx
pottedmeat: Thanks, I had trouble finding this.

src/rate/index.tsx
pottedmeat: It's not strictly required, but it makes turning decimal values into fractions a little difficult without it.

pottedmeat: value is now returned within the object

src/rate/index.tsx
```diff
@@ -0,0 +1,335 @@
...
+		if (!value) {
+			icache.set('selectedInteger', undefined);
+			icache.set('selectedStep', undefined);
+			onValue && onValue();
+		} else {
+			const [integer, step] = value.split('-').map((num) => +num);
+			const mixed = mixedNumber(integer, step, stepsLength);
+			icache.set('selectedInteger', integer);
+			icache.set('selectedStep', step);
+			onValue && onValue(mixed);
```
tomdye: can this not just return `2` / `2.5` etc?

src/rate/index.tsx
```diff
@@ -0,0 +1,335 @@
...
+			<div classes={[themedCss.groupWrapper, groupIsActive ? themedCss.active : null]}>
+				{radioIntegers}
+			</div>
+		);
+	};
+
+	return (
+		<div
+			key="radioGroupWrapper"
+			classes={[
```
tomdye: can you add a `theme.variant()` class to the root please.

src/theme/material/rate.m.css
```diff
@@ -0,0 +1,18 @@
+@import './variables.css';
```
tomdye: can you remove this variables import please, theyre loaded via variants now

src/rate/index.tsx
pottedmeat: It 100% can (as of now the property `value` is a decimal) and I originally had just the value with the detail in a second object. But I was hoping to make this easily used with formatters or to display fractions.

tomdye: in terms of the on-value callback, i think the form / parent will just want to know the decimal value.

pottedmeat: Yes, I imagine the decimal value will be frequently used.

src/rate/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+	children,
+	properties,
+	middleware: { focus, i18n, icache, theme }
+}) {
+	const { format } = i18n.localize(bundle);
+	const themedCss = theme.classes(css);
+	const {
+		name,
+		label,
+		initialValue: initialNumber = undefined,
```
tomdye, line 91: is there a reason this is renamed to `initialNumber`? I don't think there's any need to default this to undefined.

src/rate/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+		onValue,
+		max = 5,
+		steps: stepsLength = 1,
+		allowClear = false,
```
tomdye, line 95: these don't really need to be defaulted to `false`

src/rate/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+		disabled = false,
+		readOnly = false
+	} = properties();
+	const interaction = !disabled && !readOnly;
+	const initialStep = initialNumber ? Math.round((initialNumber % 1) * stepsLength) : undefined;
+	const initialInteger = initialNumber ? Math.ceil(initialNumber) : undefined;
+	const initialValue = initialNumber
+		? `${initialInteger}-${initialStep || stepsLength}`
+		: undefined;
+	const character = children()[0];
```
tomdye, line 105:
```suggestion
	const [ character ] = children();
```

src/rate/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+	const hovering = icache.getOrSet('hovering', false);
```
tomdye, line 106: Do you really need to know if the outer node is being hovered? This seems an odd thing to track.

src/rate/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+	const selectedInteger = icache.getOrSet('selectedInteger', initialInteger);
+	const selectedStep = icache.getOrSet('selectedStep', initialStep);
+	const shouldFocus = hovering ? focus.shouldFocus() : icache.getOrSet('focused', false);
```
tomdye, line 109: I think that should focus shouldn't be concerned with the hover status nor should it be set in the icache, just passed directly to the radio group.

src/rate/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+		}
+	}
+
+	const _onValue = (value?: string) => {
+		if (!value) {
+			icache.set('selectedInteger', undefined);
+			icache.set('selectedStep', undefined);
+			onValue && onValue();
+		} else {
+			const [integer, step] = value.split('-').map((num) => +num);
```
tomdye, line 140: I think that this would be a lot more straight forward if decimal values were simply passed to the radio group as options as you could just parse the radio group's `onValue` callback into an int and return it.

src/rate/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+			const mixed = mixedNumber(integer, step, stepsLength);
+			icache.set('selectedInteger', integer);
+			icache.set('selectedStep', step);
+			onValue && onValue(mixed.value);
+		}
+		focus.focus();
```
tomdye, line 146: Why is this focusing here?

src/rate/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+	/** Prevent interaction */
+	readOnly?: boolean;
+}
+
+export interface RateChildren {
+	(fill: boolean, integer: number, selected?: number, over?: number): RenderResult;
+}
+
+interface RateState {
+	uuid: string;
```
tomdye, line 50: Can you check which of these are actually used and rename to `RateICache` please.

src/rate/nls/Rate.ts
```diff
@@ -0,0 +1,21 @@
+const messages = {
+	starLabels: `{denominator, plural, offset:0
```
tomdye, line 2: I don't follow why this widget needs `nls` at all. We should just be passing the decimal values to the radio group and returning them in the `onValue`.

src/rate/index.tsx
pottedmeat: If I didn't do that, the only way of figuring out whether the user is hovering would be the state on the individual stars—any gap between stars would lose hover state, then hover state would resume again in the next star, etc.

src/rate/index.tsx
pottedmeat: I treat it internally as the radio group's initial value.

src/rate/index.tsx
pottedmeat: If the user is currently hovering, I use the focus middleware to temporarily show which star is active. On the next render, assuming they're still hovering, this is gone. The icache is used to notify me as focus changes between the radios (e.g. with keyboard navigation).

src/rate/index.tsx
pottedmeat: It tells the middleware that the next render should display what was focused.

src/rate/nls/Rate.ts
pottedmeat: This is the recommendation by WAI and the way it's implemented in Material.

pottedmeat: This is the recommendation by WAI and the way it's implemented in Material.

src/rate/index.tsx
pottedmeat: This is also related to a large problem I have with focus—if one of the radios is currently focused, I want to show the active indicator but hide it as the user hovers. If the user then moves their mouse out of the widget, I want to restore the active indicator until the user moves focus away.

src/rate/index.tsx
pottedmeat: Meaning the middleware doesn't actually perform a focus call as far as I can tell, just manages state.

src/rate/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+import RadioGroup, { RadioGroupChildren } from '../radio-group';
+import theme from '../middleware/theme';
+import * as baseCss from '../common/styles/base.m.css';
+import * as iconCss from '../theme/default/icon.m.css';
+import * as radioCss from '../theme/default/radio.m.css';
+import * as radioGroupCss from '../theme/default/radio-group.m.css';
+import * as css from '../theme/default/rate.m.css';
+import bundle from './nls/Rate';
+
+export interface MixedNumber {
```
tomdye, line 17: this isn't being used anymore

src/rate/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+import * as radioCss from '../theme/default/radio.m.css';
+import * as radioGroupCss from '../theme/default/radio-group.m.css';
+import * as css from '../theme/default/rate.m.css';
+import bundle from './nls/Rate';
+
+export interface RateProperties {
+	/** The name attribute for this rating group */
+	name: string;
+	/** The label to be displayed in the legend */
+	label?: string;
```
mwistrand, line 21: As @tomdye already pointed out, making `label` a child renderer would be more consistent with what we've been doing for the other widgets:

```
<Rate>
  {{
    label: 'Rating',
    character: (filled) => (
      <span>{filled ? <Icon type="upIcon" /> : <Icon type="downIcon" />}</span>
    )
  }}
</Rate>
```

src/rate/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+			(radioIntegers[integer] = radioIntegers[integer] || []).push(step);
+		}
+
+		return radioIntegers.reduce((nodes, radios, integer) => {
+			nodes.push(renderInteger(integer, radios));
+			return nodes;
+		}, []);
+	};
+
+	console.log('Rate.render');
```
mwistrand, line 201: Remove?

src/rate/index.tsx
pottedmeat: Yep. As soon as `Radio` is updated.

pottedmeat:
Any chance of a review?

tomdye:
Morning Neil, I think i’ve got a slightly simpler approach for this with near feature parity raised as a work in progress PR [here](https://github.com/dojo/widgets/pull/1460). Do you fancy collaborating with me on that to bring the last few features over?
I found that using a standard `input` rather than `@dojo/widgets/Radio` made styling the icons etc much simpler and I simplified the children API to make it more compatible when used as a custom element. Believe it's just missing keyboard navigation and disabled / readonly modes now.

pottedmeat:
Hey, not a chance. My dream of ever submitting a PR not rewritten by a Brit has come to an end, it seems. Following the WAI spec—using Ant and Material for reference—was a mistake.

matt-gadd:
@pottedmeat no problem, thanks for the contribution!

PR #1176, opened 2020-03-19
fix: convert RangeSlider to functional widget

### LINKED ISSUE #1028: Align existing input widgets with the `initialValue` pattern
## Enhancement

For all new input / form widgets we have implemented a pattern whereby they no longer take a controlled `value` property and instead take `initialValue`. From then on, the input will maintain and display it's own value, passing back any changes via `onValue`.

We should ensure that all existing widgets use this pattern.

### LINKED ISSUE #1147: Convert Range slider to function based and accept initialValue
Range slider should accept an `initialValue` and use the `icache` to manage it's state as per other form widgets. This will require converting it to being function based.

bitpshr:
**Type:** feature

**Description:**

This pull request updates `RangeSlider` to be a functional widget, updates its tests to use assertion templates, and updates its internal logic to use the new `initialValue` pattern.

Resolves #1028
Resolves #1147

**Preview:**

preview

### DIFF
--- a/src/range-slider/index.tsx
+++ b/src/range-slider/index.tsx
@@ -1,19 +1,18 @@
-import { uuid } from '@dojo/framework/core/util';
-import { v, w } from '@dojo/framework/core/vdom';
-import { DNode } from '@dojo/framework/core/interfaces';
-import Dimensions from '@dojo/framework/core/meta/Dimensions';
-import Focus from '../meta/Focus';
-import { theme, ThemedMixin, ThemedProperties } from '@dojo/framework/core/mixins/Themed';
-import { WidgetBase } from '@dojo/framework/core/WidgetBase';
-import { formatAriaProperties } from '../common/util';
-import Label from '../label/index';
-import * as fixedCss from './styles/range-slider.m.css';
-import * as css from '../theme/default/range-slider.m.css';
 import * as baseCss from '../common/styles/base.m.css';
+import * as css from '../theme/default/range-slider.m.css';
+import * as fixedCss from './styles/range-slider.m.css';
+import Label from '../label/index';
+import dimensions from '@dojo/framework/core/middleware/dimensions';
+import focus from '@dojo/framework/core/middleware/focus';
+import theme from '@dojo/framework/core/middleware/theme';
+import { DNode } from '@dojo/framework/core/interfaces';
+import { create, tsx } from '@dojo/framework/core/vdom';
+import { createICacheMiddleware } from '@dojo/framework/core/middleware/icache';
+import { formatAriaProperties } from '../common/util';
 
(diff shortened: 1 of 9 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-61d08p89p.now.sh
* Tests Deployment: https://dojowidgets-8awmir2wa.now.sh

src/range-slider/index.tsx
```diff
@@ -63,281 +62,253 @@ export interface RangeSliderProperties extends ThemedProperties {
...
-		);
+	if (
+		!existingInitialValue ||
+		initialValue.min !== existingInitialValue.min ||
+		initialValue.max !== existingInitialValue.max
+	) {
+		icache.set('value', initialValue);
+		icache.set('initialValue', initialValue);
+		value = initialValue;
+		onValue && onValue(initialValue);
```
tomdye: why are you calling `onValue` here? I've noticed we're doing this elsewhere too but i'm not sure why.

src/range-slider/index.tsx
```diff
@@ -63,281 +62,253 @@ export interface RangeSliderProperties extends ThemedProperties {
...
+				themeCss.root,
+				disabled ? themeCss.disabled : null,
+				focus.isFocused('root') ? themeCss.focused : null,
+				valid === false ? themeCss.invalid : null,
+				valid === true ? themeCss.valid : null,
+				readOnly ? themeCss.readonly : null,
+				showOutput ? themeCss.hasOutput : null
+			]}
+		>
+			{labelAfter ? children.reverse() : children}
```
tomdye: we've dropped `labelAfter`, I don't think we want it here anymore either, label should go above.

src/range-slider/index.tsx
bitpshr: If a consumer sets a new `initialValue` on this widget, this condition will be triggered. I think `onValue` is called here since the new `initialValue` will in turn result in a new internal `value` being set. Even though it's not technically needed, it would feel odd if this didn't trigger `onValue`, but all other internal value changes did.

dojo-site:
* Docs Deployment: https://dojowidgets-jefgpya64.now.sh
* Tests Deployment: https://dojowidgets-3ozgpe320.now.sh

src/range-slider/index.tsx
tomdye: i'm not sure we need to call `onValue` as the value was just set, could easily cause a loop

bitpshr: Updated.

dojo-site:
* Docs Deployment: https://dojowidgets-nq4b159qy.now.sh
* Tests Deployment: https://dojowidgets-6zu2d72oz.now.sh

PR #1196, opened 2020-03-25
Use Resources in Menu and rename Menu -> List

### LINKED ISSUE #1162: Convert Menu to use resources and rename to List
Now that resources / middleware have been released in framework we can convert the `Menu` widget to use these. 
Furthermore, we should rename `Menu` to `List` as it is more fitting.
This issue follows on from discussion here: https://github.com/dojo/widgets/issues/1010

tomdye:
**Type:** bug / feature

**Description:**

- Renames `menu` to `list`
- uses resources within list
- updates `select`, `context-menu` and `timepicker`
- updates examples

Resolves #1162 

Requires https://github.com/dojo/framework/pull/701 and https://github.com/dojo/framework/pull/703 and framework release.


src/examples/src/widgets/menu/FetchedResource.tsx
```diff
@@ -0,0 +1,61 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import List from '@dojo/widgets/list';
+import icache from '@dojo/framework/core/middleware/icache';
+import { createResource, createTransformer, DataTemplate } from '@dojo/widgets/common/resource';
+
+const fetcher = async (options: any) => {
```
agubler: Might as well make this non remote for the examples (as my server is slow and if it breaks the examples break)

src/examples/src/widgets/menu/MemoryResource.tsx
```diff
@@ -0,0 +1,44 @@
...
+const animals = [
+	{ name: 'cat', label: 'cat: feline' },
+	{ name: 'dog', label: 'dog: kanine', disabled: true },
+	{ name: 'mouse', label: 'mouse: rodent' },
+	{ name: 'rat', label: 'rat: rodent' }
+];
+
+export function createMemoryTemplate<S = void>(): DataTemplate<S> {
+	return {
+		read: ({ query }, put, get) => {
```
agubler: Is this the part you want to provide a helper for?

src/theme/dojo/index.ts
```diff
@@ -75,10 +75,10 @@ export default {
 	'@dojo/widgets/icon': icon,
 	'@dojo/widgets/label': label,
 	'@dojo/widgets/listbox': listbox,
-	'@dojo/widgets/list-box-item': listboxItem,
+	'@dojo/widgets/list-box-item': listItem,
 	'@dojo/widgets/loading-indicator': loadingIndicator,
 	'@dojo/widgets/menu-item': menuItem,
-	'@dojo/widgets/menu': menu,
+	'@dojo/widgets/menu': list,
```
agubler: this should be `widgets/list`

src/theme/material/index.ts
```diff
@@ -74,9 +74,9 @@ export default {
 	'@dojo/widgets/icon': icon,
 	'@dojo/widgets/label': label,
 	'@dojo/widgets/listbox': listbox,
-	'@dojo/widgets/list-box-item': listBoxItem,
+	'@dojo/widgets/list-box-item': listItem,
 	'@dojo/widgets/loading-indicator': loadingIndicator,
-	'@dojo/widgets/menu': menu,
+	'@dojo/widgets/menu': list,
```
agubler: same comment as above

src/examples/src/widgets/menu/MemoryResource.tsx
tomdye: yes, will need to support filtering too

src/theme/dojo/index.ts
tomdye: thanks

src/theme/material/index.ts
tomdye: thanks

src/examples/src/widgets/menu/FetchedResource.tsx
tomdye: no worries, can do

src/examples/src/widgets/list/FetchedResource.tsx
```diff
@@ -0,0 +1,61 @@
...
+		<virtual>
+			<List
+				resource={resource}
+				transform={transformer}
+				onValue={(value: string) => {
+					icache.set('value', value);
+				}}
+				itemsInView={10}
+			/>
+			<p>{`Clicked on: ${icache.getOrSet('value', '')}`}</p>{' '}
```
rorticus:
```suggestion
			<p>{`Clicked on: ${icache.getOrSet('value', '')}`}</p>
```

src/time-picker/index.tsx
```diff
@@ -12,6 +12,19 @@ import { createICacheMiddleware } from '@dojo/framework/core/middleware/icache';
 import { Keys } from '../common/util';
 import bundle from './nls/TimePicker';
 import i18n from '@dojo/framework/core/middleware/i18n';
+import { createResource, DataTemplate } from '@dojo/framework/core/resource';
+
+function createMemoryTemplate<S = void>(): DataTemplate<S> {
+	return {
+		read: ({ query }, put, get) => {
+			let data: any[] = get();
+			put(0, data);
```
maier49: Why do we need to `get` and then immediately `put` the same data?

src/examples/src/widgets/context-menu/Basic.tsx
```diff
@@ -11,11 +13,16 @@ export default factory(function Basic({ middleware: { icache } }) {
 		'This is some text that has a context menu. Right click to view the context menu and take an action on selected text'
 	);
 	const printedText = icache.getOrSet('printedText', '');
+	const memoryTemplate = createMemoryTemplate();
 
 	return (
 		<virtual>
 			<ContextMenu
-				options={options}
+				resource={{
```
maier49: I'm a bit out of loop on the resource stuff, but it feels like a bit complex to have to create a template to create a resource just to pass an array of data.

src/list/README.md
```diff
@@ -1,18 +1,18 @@
-# @dojo/widgets/menu
+# @dojo/widgets/list
```
rorticus: Should this talk about, or link to, resources?

src/time-picker/index.tsx
```diff
@@ -12,6 +12,19 @@ import { createICacheMiddleware } from '@dojo/framework/core/middleware/icache';
 import { Keys } from '../common/util';
 import bundle from './nls/TimePicker';
 import i18n from '@dojo/framework/core/middleware/i18n';
+import { createResource, DataTemplate } from '@dojo/framework/core/resource';
+
+function createMemoryTemplate<S = void>(): DataTemplate<S> {
```
rorticus: Can this not be imported like the other createMemoryTemplate?

src/examples/src/widgets/context-menu/Basic.tsx
tomdye: yes, agreed. We will be working on providing suitable helpers for these to make it a lot easier and simpler.

src/time-picker/index.tsx
tomdye: The `read` function gets passed  both a getter and a setter function. In the case of a memory template, the first request may only call `read` for the first few results, but we wish to get all of the data and set it back again against the query that was passed. This simple template is not the best example of this functionality really but it _does_ begin to make more sense when filters etc are concerned. Basically, for a select etc, it ensure that all the items are available and have been sideloaded into the widget meaning items can be navigated to via `a-z` press etc.

src/time-picker/index.tsx
tomdye: this is an actual widget, the helper is in the examples dir until we create a better one in framework

src/list/README.md
tomdye: i can certainly raise a follow up to discuss etc

src/examples/src/widgets/context-menu/Basic.tsx
tomdye: have made these simpler

PR #1205, opened 2020-03-25
Global event function based

### LINKED ISSUE #1183: Convert global-event to function based widget
Although global-event does not have any visible dom it is used to create global event listeners.
We should convert it to be function based and modernise where appropriate.

I do not believe there are currently any example for this widget either, these should be created - https://github.com/dojo/widgets/issues/822

ericos:
**Type:** feature

**Description:**
Updates `GlobalEvent` widget from class to function based.

Resolves #1183

### DIFF
--- a/src/global-event/index.tsx
+++ b/src/global-event/index.tsx
@@ -1,8 +1,5 @@
 import global from '@dojo/framework/shim/global';
-import WidgetBase from '@dojo/framework/core/WidgetBase';
-import { DNode } from '@dojo/framework/core/interfaces';
-import { diffProperty } from '@dojo/framework/core/decorators/diffProperty';
-import { shallow } from '@dojo/framework/core/diff';
+import { create, diffProperty, tsx } from '@dojo/framework/core/vdom';
 
 export interface ListenerObject {
 	[index: string]: (event?: any) => void;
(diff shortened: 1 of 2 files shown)


dojo-site:
* Docs Deployment: https://dojowidgets-ro6t6ptzz.now.sh
* Tests Deployment: https://dojowidgets-jxkh43hg1.now.sh

src/global-event/index.tsx
```diff
@@ -20,25 +18,30 @@ export interface RegisteredListeners {
...
 
-export class GlobalEvent extends WidgetBase<GlobalEventProperties> {
-	private _listeners: RegisteredListeners = {
+const factory = create({
+	destroy,
+	diffProperty
+}).properties<GlobalEventProperties>();
+
+export const GlobalEvent = factory(function({ children, middleware: { destroy, diffProperty } }) {
+	const listeners: RegisteredListeners = {
```
tomdye: this needs to be put into an `icache` @ericos  as they won't be saved across renders.

ericos: Good idea. Updated.

src/global-event/index.tsx
```diff
@@ -20,25 +19,37 @@ export interface RegisteredListeners {
...
+	children,
+	middleware: { destroy, diffProperty, icache }
+}) {
+	const registerListeners = (
 		type: 'window' | 'document',
 		previousListeners: RegisteredListeners,
 		newListeners: RegisteredListeners
-	) {
+	) => {
+		const currentListeners = icache.get('listeners') || {
```
tomdye: can you change this to `getOrSet` please?

src/global-event/index.tsx
```diff
@@ -56,42 +67,39 @@ export class GlobalEvent extends WidgetBase<GlobalEventProperties> {
...
 
-	@diffProperty('document', shallow)
-	protected onDocumentListenersChange(
-		previousListeners: RegisteredListeners,
-		newListeners: RegisteredListeners
-	) {
-		this._registerListeners('document', previousListeners, newListeners);
-	}
+	const removeAllRegisteredListeners = (type: 'window' | 'document') => {
+		const currentListeners = icache.get('listeners') || {
```
tomdye: getOrSet

PR #1208, opened 2020-03-26
#1171 Convert TabController to uncontrolled widget with renderer

### LINKED ISSUE #1171: Convert TabController to function based
convert to function based and use TSX

- [x] convert hyperscript to tsx (#1188)
- [x] convert `TabController` and `TabButton` to functional widgets (#1188)
- [x] inline child widgets into the main `TabController` file (#1208)
- [x] remove controlled aspect of tab changing / closing (#1208)
- [x] use child renderer for children injecting the required properties to show / hide elements (#1208)
- [x] use theme.compose for child tabs
- [x] remove `index` param from tab button callbacks (#1208)

mwistrand:
**Type:** feature

**Description:**

Resolves #1171

Remove the `activeIndex`, `onRequestChangeTab`, and `onRequestClosedTab` properties from `<TabController>` in favor of managing the active and closed tabs internally. Also, add an `initialIndex` properties so that the initial active tab index may be set.

Update `<TabController>` to accept a required `tabs` array config property, and replace its children with a renderer function that can return an array of `<TabContent>` children. Also, remove `<TabButton>` and `<Tab>` as the former has been inlined into `<TabController>` and the latter has been replaced with the simpler `<TabContent>`.


src/examples/src/widgets/tab-controller/Closeable.tsx
```diff
@@ -1,39 +1,33 @@
...
-			<Tab key="tab-three" label="Tab Three">
-				Hello Tab Three
-			</Tab>
-			<Tab key="tab-four" label="Tab Four">
-				Hello Tab Four
-			</Tab>
+				</TabContent>,
+				<TabContent key="tab1" active={isActive(1)} closed={isClosed(1)}>
+					Hello Tab Two
+				</TabContent>,
```
mwistrand: Note that `active()` and `closed()` take numerical indices instead of string keys, since otherwise `<TabController>` would need to be able to map each index to a key in order to change the active index/closed indices. If need be, we could change this by requiring each `TabItem` config to have a string `key` that would be used within the child renderer.

dojo-site:
* Docs Deployment: https://dojowidgets-dk244rhf2.now.sh
* Tests Deployment: https://dojowidgets-94t8vndo9.now.sh

dojo-site:
* Docs Deployment: https://dojowidgets-dk244rhf2.now.sh
* Tests Deployment: https://dojowidgets-94t8vndo9.now.sh

src/tab-controller/index.tsx
```diff
@@ -132,59 +153,124 @@ export const TabController = factory(function TabController({
...
+					{label}
+					{closeable ? (
+						<button
+							tabIndex={active ? 0 : -1}
+							classes={themeCss.close}
+							key={`${index}-tabbutton-close`}
+							type="button"
+							onclick={(event) => {
+								event.stopPropagation();
+								if (!disabled) {
```
maier49: If we're not going to close when disabled should we disable the button as well?

src/examples/src/widgets/tab-controller/Closeable.tsx
tomdye: if the config took anything, it should be `id` rather than `key` as key is such an overloaded term.

src/tab-controller/index.tsx
```diff
@@ -132,59 +153,124 @@ export const TabController = factory(function TabController({
...
 
-	const tabs = tabChildren.map((tab, i) => {
-		assign(tab.properties, {
-			widgetId: `${id}-tab-${i}`,
-			labelledBy: `${id}-tabbutton-${i}`,
-			show: i === activeIndex
-		});
-		return tab;
-	});
+		// Accessibility
```
tomdye: Could you remove these comments please.

src/examples/src/widgets/tab-controller/Closeable.tsx
mwistrand: I updated `TabItem` with a required `id: string` property, and updated the methods passed to the child renderer.

src/tab-controller/TabContent.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const TabContent = factory(function TabContent({
+	children,
+	middleware: { theme },
+	properties
+}) {
+	const { active, closed } = properties();
+	const themeCss = theme.classes(css);
+
+	return closed ? null : (
+		<div classes={active ? themeCss.tab : null} hidden={!active}>
```
tomdye: please add `theme.variant()` to the root

src/tab-controller/index.tsx
```diff
@@ -20,84 +21,136 @@ export enum Align {
...
 	/** Orientation of the tab buttons */
 	alignButtons?: Align;
 	/** Custom aria attributes */
 	aria?: { [key: string]: string | null };
-	/** Called when a new tab button is clicked */
-	onRequestTabChange?(index: number, key: string): void;
-	/** Called when a tab close button is clicked */
-	onRequestTabClose?(index: number, key: string): void;
+	/** Index of the initial active tab. Defaults to 0. */
+	initialIndex?: number;
```
tomdye: Could this be the tab id rather than index?

mwistrand: Yes. Good catch.

PR #1229, opened 2020-03-31
Adding typeahead widget (#795)

### LINKED ISSUE #795: Typeahead Widget - using resources / data middleware
This widget should use a `text-input` and a `menu` to facilitate a typeahead.
This will differ from the combo / filtering select as it will not be provided with a full list of results to filter.

The typeahead should provide a callback when the search string has changed and accept an array of results to show in the menu / listbox. This will enable back end searches and other complex async functions to run. 

A custom row renderer can be used if the data returned does not match the default object or has specific display requirements.

rorticus:
**Type:** feature

**Description:**

Adding Typeahead widget!

Resolves #795

### DIFF
--- a/src/select/index.tsx
+++ b/src/select/index.tsx
@@ -112,7 +112,7 @@ export const Select = factory(function Select({
 	let valid = icache.get('valid');
 	const dirty = icache.get('dirty');
 	const { messages } = i18n.localize(bundle);
-	const { get, getOptions, isLoading } = data();
+	const { get, getOptions, isLoading, getTotal } = data();
 
 	if (required && dirty) {
 		const isValid = value !== undefined;
@@ -230,7 +230,7 @@ export const Select = factory(function Select({
 							close();
 						}
 
-						return isLoading(getOptions()) ? (
+						return getTotal(getOptions()) === undefined && isLoading(getOptions()) ? (
 							<LoadingIndicator key="loading" />
 						) : (
 							<div key="menu-wrapper" classes={themedCss.menuWrapper}>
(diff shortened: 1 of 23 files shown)


src/select/index.tsx
```diff
@@ -229,7 +230,9 @@ export const Select = factory(function Select({
 							close();
 						}
 
-						return (
+						return isLoading(getOptions()) ? (
```
tomdye, line 233: I think this should only show if the resource is loading and the current total is undefined. Otherwise this would show the loading indicator instead of results when the resource is fetching the next page etc.

src/typeahead/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+			case Keys.Up:
+				preventDefault();
+				if (!onOpen()) {
+					icache.set('activeIndex', (activeIndex - 1 + total) % total);
+				}
+				break;
+			case Keys.Enter:
+				preventDefault();
+
+				const allItems = get({ query: getOptions().query }) as ListOption[];
```
tomdye, line 144: is the `get` result not already typed to `ListOption` due to the the generic passed when the middleware was created?

src/typeahead/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+						);
+					},
+					content: (toggleClosed) => {
+						function closeMenu() {
+							icache.set('focusNode', 'trigger');
+							toggleClosed();
+						}
+
+						return isLoading(getOptions()) ? (
+							<LoadingIndicator key="loading" />
```
tomdye, line 276: same comment here re. total = undefined.

Issue #1250, opened 2020-04-02
has('test') is false during tests, which breaks resize middleware mocks


maier49:
**Bug** 

When running tests, `has('test')` returns false. This results in shims not being properly wrapped, and as a result the resize and breakpoint middleware mocks fail when running in node trying to access a nonexistent element. 
Package Version: 

**Expected behavior:**
`has('test')` should. be `true` when running tests

**Actual behavior:**
It is not

PR #1251, opened 2020-04-02
Two column layout widget

### LINKED ISSUE #1222: 2 Column Layout
Create a two column layout widget with opinionated size defaults. 
- Should take two child renderers
   - leading
   - trailing
- accept bias argument for `leading` / `trailing` to determine which side takes up available space
- should collapse in a responsive manner

eg:

```tsx
<TwoColumn bias="trailing">
   {{
      leading: <div>leading content</div>,
      trailing: <div>trailing content</div>
   }}
</TwoColumn>
```

maier49:
**Type:** feature

**Description:**
Adds a two column layout widget. By default the columns get 50% of the available space, and the
layout switches to one column below 600px width. The breakpoint width can be set as a property, and a "bias" can be set to make the leading or trailing column expand to take up all available space. If a bias is set that column will also be the one that is shown when minimized.

The mock in the tests is a workaround for #1250
Resolves #1222

### DIFF
--- a/.dojorc
+++ b/.dojorc
@@ -53,6 +53,7 @@
 			"src/title-pane",
 			"src/tooltip",
 			"src/trigger-popup",
+			"src/two-column-layout",
 			"src/typeahead"
 		]
 	}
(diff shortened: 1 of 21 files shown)


maier49:
@tomdye The "small" column size is now 320px in the Dojo and material themes (both use the same grid size), and is themeable.

Issue #1243, opened 2020-04-01
Migration Guide - Matt W


tomdye:
Contribute to migration under /docs/V7-Migration-Guide.md
- [x] Tab controller (#1284)
- [x] checkbox (#1279)

PR #1279, opened 2020-04-02
Checkbox v7 migration doc updates

mwistrand:
**Type:** feature

**Description:**

Addresses part of #1243.

Update the v7 migration document with property changes to the checkbox widget, along with an example demonstrating a migration from v6 to v7.


tomdye:
this will need updates re. `label` after child renderer changes

PR #1281, opened 2020-04-02
Use child renderer for Radio label

### LINKED ISSUE #1262: Radio child renderers
**Enhancement** 
 - Radio label should be an optional child of type RenderResult

ericos:
**Type:** feature

**Description:**

Moves label property of Radio widget to an optional child of type RenderResult.

Resolves #1262 

Screen Shot 2020-04-02 at 10 28 34 AM

### DIFF
--- a/src/examples/src/widgets/radio/Labelled.tsx
+++ b/src/examples/src/widgets/radio/Labelled.tsx
@@ -4,11 +4,5 @@ import Radio from '@dojo/widgets/radio';
 const factory = create();
 
 export default factory(function LabelledRadioButton() {
-	return (
-		<Radio>
-			{{
-				label: () => 'Radio Button 1'
-			}}
-		</Radio>
-	);
+	return <Radio>Radio Button 1</Radio>;
 });
--- a/src/examples/src/widgets/radio/Labelled.tsx
+++ b/src/examples/src/widgets/radio/Labelled.tsx
@@ -1,8 +1,15 @@
 import { create, tsx } from '@dojo/framework/core/vdom';
 import Radio from '@dojo/widgets/radio';
+import Label from '@dojo/widgets/label';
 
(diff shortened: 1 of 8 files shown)


src/radio/index.tsx
```diff
@@ -42,15 +41,25 @@ export interface RadioProperties extends ThemeProperties, FocusProperties {
 	widgetId?: string;
 }
 
-const factory = create({ focus, theme }).properties<RadioProperties>();
+export interface RadioChild {
+	label?: () => RenderResult;
```
tomdye, line 45: this should just be a single child of type `RenderResult`, ie. a normal child. No need for the function or the object as nothing is injected.

src/radio/index.tsx
```diff
@@ -128,7 +138,7 @@ export const Radio = factory(function Radio({ properties, id, middleware: { focu
 					required={required}
 					secondary={true}
 				>
-					{label}
+					{label()}
```
tomdye, line 141: think this can just be `children()`

src/examples/src/widgets/radio/Labelled.tsx
```diff
@@ -4,5 +4,11 @@ import Radio from '@dojo/widgets/radio';
 const factory = create();
 
 export default factory(function LabelledRadioButton() {
-	return <Radio label={'Radio Button 1'} />;
+	return (
+		<Radio>
```
tomdye, line 8: I'd expect this to be `<Radio>Radio Button 1</Radio>`

ericos: Updated this per feedback. Simplified implementation to eliminate render function, etc.

ericos:
Updated this per discussion. Also updated radio group usage of the Radio widget.

PR #1288, opened 2020-04-02
Update tooltip to use a child renderer for content

### LINKED ISSUE #1274: Tooltip child renderers
**Enhancement** 

- content should be child of type `RenderResult`

maier49:
**Type:** feature

**Description:**
Updates the tooltip to use a child renderer for content, moves target element to a child renderer named `target`.
Resolves #1274

### DIFF
--- a/src/tooltip/index.tsx
+++ b/src/tooltip/index.tsx
@@ -16,7 +16,7 @@ export interface TooltipProperties {
 }
 
 export interface TooltipChildren {
-	target?: RenderResult;
+	trigger?: RenderResult;
 	content: RenderResult;
 }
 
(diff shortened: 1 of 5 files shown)


src/tooltip/index.tsx
```diff
@@ -1,22 +1,25 @@
...
-	/** Information to show within the tooltip */
-	content: DNode;
 	/** Determines if this tooltip is visible */
 	open?: boolean;
 	/** Where this tooltip should render relative to its child */
 	orientation?: Orientation;
 }
 
+export interface TooltipChildren {
+	target?: RenderResult;
```
tomdye, line 19: could we go with `trigger` please to mirror the trigger popup.

Issue #1242, opened 2020-04-01
Migration guide - Eric


tomdye:
Contribute to migration under /docs/V7-Migration-Guide.md
- [x] Global-event - #1289 
- [x] Button - #1289
- [x] Progress - #1278 
- [x] Radio - #1252

PR #1289, opened 2020-04-02
v7 Migration guide updates for button and global-event

ericos:
**Description:**

Removes button and global-event sections due to backwards compatibility. No changes needed.

Part of #1242


ericos:
> on second thoughts, i think we changed the event callbacks for button significantly.

Ah, good catch! I added Button back to the document.

PR #1290, opened 2020-04-02
feat: simpler AccordionPane

bitpshr:
**Type:** bug / feature

**Description:**

Resolves #???

### DIFF
--- a/src/accordion-pane/index.tsx
+++ b/src/accordion-pane/index.tsx
@@ -1,10 +1,14 @@
 import * as css from '../theme/default/accordion-pane.m.css';
 import * as titlePaneCss from '../theme/default/title-pane.m.css';
 import theme from '../middleware/theme';
-import { RenderResult, Theme, ThemeWithVariant } from '@dojo/framework/core/interfaces';
-import { TitlePaneProperties } from '../title-pane';
+import { RenderResult } from '@dojo/framework/core/interfaces';
+import { TitlePane, TitlePaneProperties, TitlePaneChildren } from '../title-pane';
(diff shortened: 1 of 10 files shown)


src/accordion-pane/index.tsx
```diff
@@ -63,24 +68,46 @@ export const AccordionPane = factory(function LoadingIndicator({
...
+						onOpen={onOpen(key)}
+						open={open(key)}
+						theme={localTheme}
+					>
+						{paneChildren()[0]}
+					</TitlePane>
+				</div>
+			);
+		})
+	);
```
bitpshr, line 107: @tomdye: The `AccordionPane` dynamically creates this `Pane` to return to its child renderer. This has to be dynamically created since it uses `onClose`, `onOpen`, and `open` functions that are local to `AccordionPane`. All works fine, except passing the composed theme to the child `TitlePane` on line 100. Everything looks correct, but `TitlePane`, for whatever reason, doesn't pick up and use the classes. Oddly, if I manually delete the `variant` property on that `localTheme` composed theme object, everything works.

Any thoughts on what I need to do differently?

tomdye: Paul I simply meant you'd create a new widget inside the accordion pane index.tsx called `Pane` that did the theme.compose onto the titlepane or simply wrapped them in a div or something. It wouldn't be injected into the renderer, the user would just import it from the file like we do with breadcrumb / breadcrumbSeparator. The user would still have to do the open / close checks etc via the callbacks, this was meant to simply remove the `theme` step.

PR #1295, opened 2020-04-02
feat: simpler AccordionPane

### LINKED ISSUE #1224: AccordionPane improvements
**Enhancement**

The `AccordionPane` expects a child renderer that returns `TitlePane` widgets, but these have to manually apply many renderer-supplied props. We should make `AccordionPane` return a `Pane` to each renderer instead of additional props, to make the use of `AccordionPane` easier.

Package Version: `7.0.0`

### LINKED ISSUE #1345: Allow title-pane to be controlled
## Proposal

## New Properties

- `open?: boolean`

bitpshr:
**Type:** feature

**Description:**

This pull request adds a new supporting `Pane` widget that's exported from the `Accordion` class, meant to be used in place of raw `TitlePane` children. This mitigates the need to pass a `theme` to the child renderer to align with the latest widget conventions.

Resolves #1224 
Resolves #1345


.dojorc
```diff
@@ -6,7 +6,7 @@
 	"build-widget": {
 		"prefix": "dojo",
 		"widgets": [
-			"src/accordion-pane",
+			"src/accordion",
```
bitpshr, line 9: `AccordionPane` was renamed `Accordion` throughout, as this felt more appropriate now that a `Pane` is also exported.

src/title-pane/index.tsx
```diff
@@ -15,6 +15,8 @@ export interface TitlePaneProperties {
 	headingLevel?: number;
 	/** If true the pane is opened and content is visible initially */
 	initialOpen?: boolean;
+	/** Explicitly control TitlePane */
+	open?: boolean;
```
tomdye, line 19: you should be using `initialOpen` here to control it, not `open`.

bitpshr: @tomdye per our discussion, this should be ready for a final review as-is. Thanks in advance.

src/title-pane/index.tsx
```diff
@@ -55,19 +57,12 @@ export const TitlePane = factory(function TitlePane({
...
 		onOpen,
+		open: openProp,
 		theme: themeProp
 	} = properties();
 	const { content, title } = children()[0];
 
-	let open = icache.get('open');
-	let performTransition = false;
-	const existingInitialOpen = icache.get('initialOpen');
-	if (existingInitialOpen !== initialOpen) {
```
tomdye, line 65: I believe you'd still need the `initialOpen` equality check in here would you not?

PR #1310, opened 2020-04-03
Select: moved label to child renderer

### LINKED ISSUE #1266: Select child renderers
**Enhancement** 
- label should be child of type `RenderResult`

jellison:
**Type:** feature

**Description:**

Moves `label` from property to child renderer

Resolves #1266

### DIFF
--- a/src/select/index.tsx
+++ b/src/select/index.tsx
@@ -41,8 +41,6 @@ export interface SelectProperties {
 	disabled?: boolean;
 	/** Sets the helper text of the input */
 	helperText?: string;
-	/** The label to show */
-	label?: RenderResult;
 	/** Boolean to indicate if field is required */
 	required?: boolean;
 	/** Callabck when valid state has changed */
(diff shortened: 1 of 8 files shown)


src/select/index.tsx
```diff
@@ -96,7 +95,7 @@ export const Select = factory(function Select({
 		resource,
 		transform
 	} = properties();
-	const [itemRenderer] = children();
+	const { itemRenderer, label } = children()[0] || ({} as SelectChildren);
```
tomdye, line 98:
```suggestion
	const [{ itemRenderer, label } = { itemRenderer = undefined, label = undefined }] = children();
```

src/select/index.tsx
```diff
@@ -53,7 +51,9 @@ export interface SelectProperties {
 
 export interface SelectChildren {
 	/** Custom renderer for item contents */
-	(properties: ItemRendererProperties): RenderResult;
+	itemRenderer?(properties: ItemRendererProperties): RenderResult;
```
tomdye, line 54: could you rename this to `items` please? I believe this is more in line with what we've named things elsewhere, ie. `radios` / `checkboxes` etc for those rendered in the groups.

Issue #1313, opened 2020-04-03
PasswordInput theme not working


jellison:
#### Bug

PasswordInput theme isn't working correctly. Likely due to a theme compose prefix without a composes in the dojo theme to re-apply the original styles as well as the changes.

### Actual

![image](https://user-images.githubusercontent.com/41762295/78373309-719fd380-7598-11ea-9612-7bb4444d40b5.png)

### Expected

![image](https://user-images.githubusercontent.com/41762295/78373355-7fedef80-7598-11ea-9b4d-f42d0747cf7e.png)

PR #1322, opened 2020-04-03
Theme compose and Password input fixes

tomdye:
**Type:** bug 

**Description:**

Fixes password-input styles by fixing theme.compose to work properly with theme variants.

Resolves https://github.com/dojo/widgets/issues/1313


src/middleware/theme.tsx
```diff
@@ -69,6 +73,17 @@ export const theme = factory(function({ middleware: { coreTheme }, properties })
 					{} as ClassNames
 				);
 				baseTheme = { ...baseTheme, ...prefixedCss };
+
+				if (isThemeWithVariant(theme)) {
```
maier49, line 77: Does the `ThemeWithVariants` case need any additional consideration?

PR #1327, opened 2020-04-06
Pagination: automatically remove sibling links as space permits

### LINKED ISSUE #1239: Enhance Pagination to dynamically adjust sibling links
**Enhancement** 

Enhance the Pagination widget to automatically remove sibling links when there isn't space to include them. `siblingCount` effectively becomes a "maximum" sibling count. `siblingCount` of undefined would include as many sibling links as can fit.

jellison:
DRAFT: some unit tests are missing an implementation

**Type:** feature

**Description:**

Pagination widget now responsive by automatically removing sibling links when space is insufficient to render `siblingCount` links.

When `siblingCount` is unset, all siblings are included which will fit the available space.

Resolves #1239 

##### Dojo theme
image

##### Material theme
image

### Mid-sized

image

### Min-sized

image

### DIFF
--- a/src/pagination/index.tsx
+++ b/src/pagination/index.tsx
@@ -4,13 +4,8 @@ import i18n from '@dojo/framework/core/middleware/i18n';
 import theme from '@dojo/framework/core/middleware/theme';
 import dimensions from '@dojo/framework/core/middleware/dimensions';
 import resize from '@dojo/framework/core/middleware/resize';
-import { RenderResult } from '@dojo/framework/core/interfaces';
-import {
-	Resource,
-	createResource,
-	ResourceQuery,
-	DataTemplate
-} from '@dojo/framework/core/resource';
+import { RenderResult, Theme, ThemeWithVariant } from '@dojo/framework/core/interfaces';
+import { Resource, createResource } from '@dojo/framework/core/resource';
 import global from '@dojo/framework/shim/global';
 
 import Icon from '../icon';
@@ -19,28 +14,6 @@ import { ListOption } from '../list';
 
 import bundle from './Pagination.nls';
 import * as css from '../theme/default/pagination.m.css';
-
-export function createMemoryTemplate<S = void>({
-	filter
-}: { filter?: (query: ResourceQuery[], v: S) => boolean } = {}): DataTemplate<S> {
-	return {
-		read: ({ query }, put, get) => {
-			let data: any[] = get();
-			const filteredData = filter && query ? data.filter((i) => filter(query, i)) : data;
-			put(0, filteredData);
-			return { data: filteredData, total: filteredData.length };
-		}
-	};
-}
-
-export function createMemoryResourceWithData<S = any>(data: S[]) {
-	const memoryTemplate = createMemoryTemplate<S>();
-
-	return {
-		resource: () => createResource(memoryTemplate),
-		data
-	};
-}
 
 export interface PaginationProperties {
 	/** The initial page number */
@@ -69,9 +42,41 @@ interface PaginationCache {
 	initialPage?: number;
 	initialPageSize?: number;
 	currentPage: number;
+	nodeWidths: { prev: number; next: number; current: number; sibling: number };
 	pageSize: number;
 	pageSizes: number[];
 	pageSizesResource: { resource: () => Resource; data: ListOption[] };
+	theme?: Theme | ThemeWithVariant;
+}
+
+function createMemoryResourceWithData<S = any>(data: S[]) {
+	return {
+		resource: () =>
+			createResource({
+				read: (_, put, get) => {
+					let data: any[] = get();
+					put(0, data);
+					return { data, total: data.length };
+				}
+			}),
+		data
+	};
+}
+
+function getRenderedWidth(dnode: RenderResult, wrapperClass?: string): number {
+	if (dnode === undefined) {
+		return 0;
+	}
+
+	const r = renderer(() => dnode);
+	const div = global.document.createElement('div');
+	div.className = wrapperClass || '';
+	div.style.position = 'absolute';
+	global.document.body.appendChild(div);
+	r.mount({ domNode: div, sync: true });
+	const dimensions = div.getBoundingClientRect();
+	global.document.body.removeChild(div);
+	return dimensions.width;
 }
 
 const icache = createICacheMiddleware<PaginationCache>();
@@ -115,31 +120,17 @@ export default factory(function Pagination({
 		);
 	}
 
+	if (theme.get() !== icache.get('theme')) {
+		icache.set('theme', theme.get());
+		icache.delete('nodeWidths');
+	}
+
 	const page = icache.getOrSet('currentPage', 1);
 	const pageSize = icache.getOrSet('pageSize', 10);
 	const pageSizesResource = icache.get('pageSizesResource');
 
-	const getRenderedWidth = (dnode: RenderResult) => {
-		if (dnode === undefined) {
-			return 0;
-		}
-
-		const r = renderer(() => dnode);
-		const div = global.document.createElement('div');
-		div.className = theme.variant() || '';
-		div.style.position = 'absolute';
-		global.document.body.appendChild(div);
-		r.mount({ domNode: div, sync: true });
-		const dimensions = div.getBoundingClientRect();
-		global.document.body.removeChild(div);
-		return dimensions.width;
-	};
-
-	// If provided, use `siblingCount` as an upper bound for sibling links
-	const leadingCount = siblingCount ? Math.min(siblingCount, page - 1) : page - 1;
-	const trailingCount = siblingCount ? Math.min(siblingCount, total - page) : total - page;
-
-	const prevLink = leadingCount ? (
+	const showPrev = page > 1;
+	const prevLink = (
 		<button
 			type="button"
 			key="prev"
@@ -155,11 +146,10 @@ export default factory(function Pagination({
 			</div>
 			<div classes={classes.label}>{messages.previous}</div>
 		</button>
-	) : (
-		undefined
 	);
 
-	const nextLink = trailingCount ? (
+	const showNext = page < total;
+	const nextLink = (
 		<button
 			type="button"
 			key="next"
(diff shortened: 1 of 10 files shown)


src/pagination/index.tsx
```diff
@@ -12,17 +20,27 @@ import { ListOption } from '../list';
 import bundle from './Pagination.nls';
 import * as css from '../theme/default/pagination.m.css';
 
-function createMemoryTemplate<S = void>(): DataTemplate<S> {
+export function createMemoryTemplate<S = void>({
+	filter
+}: { filter?: (query: ResourceQuery[], v: S) => boolean } = {}): DataTemplate<S> {
 	return {
-		read: (_, put, get) => {
+		read: ({ query }, put, get) => {
```
tomdye, line 27: what made you add this in? I don't believe you're using the filter or query at all are you?

src/pagination/index.tsx
```diff
@@ -53,20 +71,27 @@ interface PaginationCache {
 	currentPage: number;
 	pageSize: number;
 	pageSizes: number[];
-	pageSizesData: ListOption[];
+	pageSizesResource: { resource: () => Resource; data: ListOption[] };
```
tomdye, line 74: I think this should stay as `ListOption[]`, there's no need to expose the resource usage to the consumer of this pagination widget

src/pagination/index.tsx
```diff
@@ -84,105 +109,167 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
...
+			createMemoryResourceWithData(pageSizes.map((ps) => ({ value: ps.toString() })))
+		);
 	}
 
 	const page = icache.getOrSet('currentPage', 1);
 	const pageSize = icache.getOrSet('pageSize', 10);
-	const pageSizesData = icache.get('pageSizesData');
+	const pageSizesResource = icache.get('pageSizesResource');
+
+	const getRenderedWidth = (dnode: RenderResult) => {
```
tomdye, line 122: this should be outside of the factory so it's only created once

src/pagination/index.tsx
```diff
@@ -84,105 +109,167 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
...
+		</div>
+	);
 
-	const leading: WNode[] = [];
-	const leadingCount = Math.min(siblingCount, page - 1);
-	for (let i = 1; i <= leadingCount; i++) {
-		const linkTo = page - i;
+	let availableWidth =
+		containerWidth -
+		getRenderedWidth(prevLink) -
```
tomdye, line 197: all of these sizes should really be saved in an `icache` and not calculated every time, just on first render.

src/pagination/index.tsx
```diff
@@ -84,105 +109,167 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
...
+		getRenderedWidth(current) -
+		getRenderedWidth(nextLink);
 
-		leading.unshift(
+	for (let i = 1; i <= Math.max(leadingCount, trailingCount); i++) {
```
tomdye, line 201: Feels to me that this should just calculate what there's room for and render it, this trial and error approach will be costly. Should just calc and store the width of elements without buttons, and the width of a button then on resize, calculate how many fill fit, perhaps use a threshold of 90% available space? When calculating the width of the button up front you could use the largest page number as it's label, then you can always be sure they'll fit.

src/pagination/index.tsx
jellison: No, I just copied it from examples.

src/pagination/index.tsx
jellison: I had it outside, but it needs `theme.variant()` to render properly. I could just pass that in for each case.

src/pagination/index.tsx
jellison: The intention was to only create the source when necessary (ie, when the `pageSizeData` had changed).

src/pagination/index.tsx
jellison: I'll try it. I worry there will still be scenarios where the rendered width changes, such as changing themes in Parade (although I can't think of any others).

tomdye: you could clear the widths in the icache when the theme changes

src/pagination/index.tsx
jellison: I think I've incorporated correct resource usage, but I'm still storing the resource in the cache. I'll let you resolve this conversation if you agree.

src/pagination/index.tsx
```diff
@@ -51,22 +42,61 @@ interface PaginationCache {
...
 	currentPage: number;
+	nodeWidths: { prev: number; next: number; current: number; sibling: number };
 	pageSize: number;
 	pageSizes: number[];
-	pageSizesData: ListOption[];
+	pageSizesResource: { resource: () => Resource; data: ListOption[] };
+	theme?: Theme | ThemeWithVariant;
+}
+
+function createMemoryResourceWithData<S = any>(data: S[]) {
```
tomdye, line 52: This was used as a helper in the examples but realistically you should be creating a single resource outside of your render and then passing the data to it. You shouldn't be creating a whole new resource each time.

So here, just do

```ts
const template = {
				read: (_, put, get) => {
					let data: any[] = get();
					put(0, data);
					return { data, total: data.length };
				}
			};

const resource = createResource(template);
```

Then when you use it, pass it as
```
resource={{ resource: () => resource, data }}
```

The helper for the first part will be coming very soon via framework, today hopefully

src/pagination/index.tsx
```diff
@@ -84,105 +114,158 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
 
 	if (pageSizes !== undefined && pageSizes !== icache.get('pageSizes')) {
 		icache.set('pageSizes', pageSizes);
-		icache.set('pageSizesData', pageSizes.map((ps) => ({ value: ps.toString() })));
+		icache.set(
+			'pageSizesResource',
+			createMemoryResourceWithData(pageSizes.map((ps) => ({ value: ps.toString() })))
```
tomdye, line 119: shouldn't be doing this here, see my comment above

src/pagination/index.tsx
```diff
@@ -84,105 +114,158 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
...
+			</div>
+			<div classes={classes.label}>{messages.next}</div>
+		</button>
+	);
+
+	const leadingSiblings: RenderResult[] = [];
+	const trailingSiblings: RenderResult[] = [];
+	const containerWidth = dimensions.get('links').size.width;
+	if (containerWidth) {
+		const nodeWidths = icache.getOrSet('nodeWidths', () => ({
```
tomdye, line 174: I think this would be better as individual icache entries.
Is it possible to just get the widget width when rendered with zero pages, and the width of a single page link and use those to do the maths? This seems like a fair few things to be checking. 
It also appears that it's duplicating the render code for the buttons / pages etc which could easily get out of sync.

src/pagination/index.tsx
```diff
@@ -84,105 +114,158 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
...
+
+		const maxLeading = siblingCount ? Math.min(siblingCount, page - 1) : page - 1;
+		const maxTrailing = siblingCount ? Math.min(siblingCount, total - page) : total - page;
+		let numSiblings = 0;
+
+		for (let i = 1; i <= Math.max(maxLeading, maxTrailing); i++) {
+			const showLeading = i <= maxLeading;
+			const showTrailing = i <= maxTrailing;
+
+			if (maxSiblings >= numSiblings + +showLeading + +showTrailing) {
```
tomdye, line 202: this is quite hard to follow

src/pagination/index.tsx
```diff
@@ -84,105 +114,158 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
...
+				if (showLeading) {
+					leadingSiblings.unshift(
+						<button
+							type="button"
+							key={`numberedLink-${page - i}`}
```
tomdye, line 207: perhaps create `previousPageNumber` and `nextPageNumber` variables?

src/pagination/index.tsx
```diff
@@ -84,105 +114,158 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
...
+							classes={[classes.numberedLink, classes.link]}
+							onclick={(e) => {
+								e.stopPropagation();
+								icache.set('currentPage', page - i);
+								onPage && onPage(page - i);
+							}}
+						>
+							{(page - i).toString()}
```
tomdye, line 215:
```
{`${previousPage}`}
```

src/pagination/index.tsx
```diff
@@ -84,105 +114,158 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
...
+							type="button"
+							key={`numberedLink-${page + i}`}
+							classes={[classes.numberedLink, classes.link]}
+							onclick={(e) => {
+								e.stopPropagation();
+								icache.set('currentPage', page + i);
+								onPage && onPage(page + i);
+							}}
+						>
+							{(page + i).toString()}
```
tomdye, line 233:
```
{`${nextPage}`}
```

src/pagination/index.tsx
jellison: The difficulty is that "zero page width" changes for the first and last pages because there are no previous and next links, respectively.

src/pagination/index.tsx
```diff
@@ -84,105 +114,158 @@ export default factory(function Pagination({ middleware: { theme, icache, i18n }
...
 	if (pageSizes !== undefined && pageSizes !== icache.get('pageSizes')) {
 		icache.set('pageSizes', pageSizes);
-		icache.set('pageSizesData', pageSizes.map((ps) => ({ value: ps.toString() })));
+		icache.set(
+			'pageSizesResource',
+			createMemoryResourceWithData(pageSizes.map((ps) => ({ value: ps.toString() })))
+		);
+	}
+
+	if (theme.get() !== icache.get('theme')) {
```
tomdye, line 123: I think you should be able to use `diffProperty` here for this.
Could you try:

```ts
diffProperty(‘theme’, (current, next) => {
     if (current.theme !== next.theme) { 
      // recalculate here?
    }
});
```

src/pagination/index.tsx
jellison: I'll clean up the whole loop and we can revisit.

src/pagination/index.tsx
jellison: Got it, thanks!

src/pagination/index.tsx
jellison: Good call. I should use that more.

PR #1351, opened 2020-04-07
Controlled text area

### LINKED ISSUE #1342: Allow text area to be controlled
## Proposal

## New Properties

- `value?: string`

maier49:
**Type:** feature

**Description:**
Allows a `value` property to be specified to control a text area component.
Resolves #1342

### DIFF
--- a/src/text-area/index.tsx
+++ b/src/text-area/index.tsx
@@ -171,9 +171,8 @@ export const TextArea = factory(function TextArea({
 	} = properties();
 
 	let { value } = properties();
-	const hasValue = typeof value !== 'undefined';
 
-	if (!hasValue) {
+	if (value === undefined) {
 		value = icache.get('value');
 		const existingInitialValue = icache.get('initialValue');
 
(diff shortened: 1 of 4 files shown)


src/text-area/index.tsx
```diff
@@ -169,13 +170,18 @@ export const TextArea = factory(function TextArea({
 		onValidate
 	} = properties();
 
-	let value = icache.get('value');
-	const existingInitialValue = icache.get('initialValue');
+	let { value } = properties();
+	const hasValue = typeof value !== 'undefined';
```
tomdye, line 174: I don't think this will ever not be true due to the default in the destructuring.
How about using the same approach I did in text-input?

https://github.com/dojo/widgets/pull/1357/files#diff-be70a03acb1add410d9f0b3bd7817872R172-R184

maier49: That default in the destructuring is in a different function, btw, the diff is just trying to deceive you.

PR #1329, opened 2020-04-07
Docs for select/native-select

maier49:
**Type:** feature

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code matches the [style guide](https://github.com/dojo/framework/blob/master/STYLE.md)
* [x] Unit tests are included in the PR
* [ ] For new widgets, an entry has been added to the `.dojorc`
* [ ] For new widgets, `theme.variant()` is added to the root domnode
* [ ] Any widget variant uses `theme.compose` like [this](https://github.com/dojo/widgets/issues/847)
* [ ] WidgetProperties are exported

**Description:**
Migration guide for native select and select
Resolves #1240

PR #1353, opened 2020-04-08
Optionally controlled native select

### LINKED ISSUE #1335: Allow native-select to be controlled
## Proposal

## New Properties

- `value: string`

ericos:
**Type:** feature

**Description:**

Will need to update migration guide PR #1329 

Allows native-select to be controlled. Adds `value` property.

Resolves #1335 

Screen Shot 2020-04-07 at 10 36 18 PM

### DIFF
--- a/src/native-select/index.tsx
+++ b/src/native-select/index.tsx
@@ -18,6 +18,8 @@ export interface NativeSelectProperties {
 	onValue?(value: string): void;
 	/** The initial selected value */
 	initialValue?: string;
+	/** A controlled value */
+	value?: string;
 	/** Options to display within the menu */
 	options: MenuOption[];
 	/** Property to determine if the input is disabled */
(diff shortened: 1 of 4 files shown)


src/native-select/index.tsx
```diff
@@ -121,7 +131,7 @@ export const NativeSelect = factory(function NativeSelect({
 					}}
 					classes={themedCss.select}
 				>
-					{!initialValue && <option key="blank-option" value="" />}
+					{value === undefined && <option key="blank-option" value="" />}
```
maier49, line 134: This changes the behavior. Previously is no `initialValue` was provided then this blank value would always display, even after selecting a value. Now it is only displayed until you select a value and then goes away.

tomdye: fair point, please can you fix this @ericos

PR #1358, opened 2020-04-08
Fix pointer for Calendar in material theme

tomdye:
**Type:** bug / feature

**Description:**

For material calendar theme:
- adds cursor: pointed where appropriate
- disable user selection
- remove fieldset borders
- center text in dates

### DIFF
--- a/src/theme/material/calendar.m.css
+++ b/src/theme/material/calendar.m.css
@@ -43,13 +43,15 @@
 	text-align: center;
 	position: relative;
 	z-index: 0;
-}
-
-.date {
 	cursor: pointer;
 	display: inline-flex;
 	justify-content: center;
 	align-items: center;
+}
+
+.monthFields,
+.yearFields {
+	border: none;
 }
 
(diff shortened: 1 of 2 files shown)


src/theme/material/calendar.m.css
```diff
@@ -41,6 +43,15 @@
...
 	z-index: 0;
+	cursor: pointer;
+	display: inline-flex;
+	justify-content: center;
+	align-items: center;
+}
+
+.monthFields,
+.yearFields {
+	border: none;
```
maier49, line 54: Padding and margin should be set to 0 here too I believe. The years were 4 to a row when the resets were applied

ericos:
Per the Material guidelines, there should probably be hover styles (6 in diagram below) on the Material calendar. I can create another issue if needed.

mio-staging_mio-components_1584058305895_assets_0B54x0KDcs5U-d0dJMFg2OTdGWjg_picker-elements-desktop

PR #1362, opened 2020-04-08
Optionally control radio group

### LINKED ISSUE #1337: Allow radio group to be controlled
## Proposal

## New Properties

- `value?: string[]`

maier49:
**Type:** feature

**Description:**
Adds back a `value` property to allow optionally controlling the radio group.

Resolves #1337


tomdye:
Is there any way to keep a single value passed down to the middleware? ie. resolve the value / initial value before hand?

maier49:
@tomdye I don't think there is, for the same reason we need two properties in the first place. We need to know whether the value is an `initialValue` that should only be respected when it changes or a `value` that should override local changes.

tomdye:
Thanks for the explanation @maier49

PR #1363, opened 2020-04-08
Optionally control select

### LINKED ISSUE #1339: Allow select to be controlled
## Proposal

## New Properties

- `value?: string[]`

maier49:
**Type:** feature

**Description:**
Adds an optional `value` property to select
Resolves #1339

### DIFF
--- a/src/examples/src/widgets/select/Controlled.tsx
+++ b/src/examples/src/widgets/select/Controlled.tsx
@@ -12,7 +12,7 @@ const options = [
 
 const resource = createMemoryResourceWithData(options);
 
-export default factory(function Basic({ middleware: { icache } }) {
+export default factory(function Controlled({ middleware: { icache } }) {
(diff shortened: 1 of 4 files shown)


src/examples/src/widgets/select/Controlled.tsx
```diff
@@ -0,0 +1,33 @@
...
+const factory = create({ icache });
+const options = [
+	{ value: 'cat', label: 'Cat' },
+	{ value: 'dog', label: 'Dog' },
+	{ value: 'fish', label: 'Fish' }
+];
+
+const resource = createMemoryResourceWithData(options);
+
+export default factory(function Basic({ middleware: { icache } }) {
```
ericos, line 15:
```suggestion
export default factory(function Controlled({ middleware: { icache } }) {
```
👯‍♂

PR #1366, opened 2020-04-08
Make range-slider optionally controlled

### LINKED ISSUE #1338: Allow rangeSlider to be controlled
## Proposal

## New Properties

- `value?: RangeValue`

ericos:
**Type:** feature

**Description:**

Adds `value` prop to `RangeSlider` widget to allow the option for it to be controlled.

Resolves #1338

### DIFF
--- a/docs/V7-Migration-Guide.md
+++ b/docs/V7-Migration-Guide.md
@@ -517,7 +517,7 @@ Latest example can be found on [widgets.dojo.io/#widget/radio/overview](https://
   - No longer supported
 #### Changes in behaviour
 
-The range slider widget now internally manages its own value by default and changed via the `initialValue` property. The range slider can be fully controlled using the `value` property.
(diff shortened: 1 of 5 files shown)


docs/V7-Migration-Guide.md
```diff
@@ -515,7 +517,7 @@ Latest example can be found on [widgets.dojo.io/#widget/radio/overview](https://
   - No longer supported
 #### Changes in behaviour
 
-The range slider widget now internally manages its own value. The value can be changed via the `initialValue` property, and updates are exposed via the `onValue` callback.
+The range slider widget now internally manages its own value by default and changed via the `initialValue` property. The range slider can be fully controlled using the `value` property.
```
maier49, line 520:
```suggestion
The range slider widget now internally manages its own value by default and can be changed via the `initialValue` property. The range slider can be fully controlled using the `value` property.
```

PR #1373, opened 2020-04-13
TS forward compatibility

### LINKED ISSUE #1372: Add support for TypeScript 3.8.x and potentially 3.9.x
**Enhancement** 

Widgets should be compatible with 3.8 and the latest snapshot of 3.9

maier49:
**Type:** feature

**Description:**
Fixes issues found when building and running tests with later versions of TS. Seemed to work fine when being used as a library. Tested using the example app.
Resolves #1372


tomdye:
Should this be updating TS version in `package.json` too?

agubler:
@tomdye no, the plan is to be forwards compatible but not mandate users to update typescript.

PR #1374, opened 2020-04-13
Adding multi-select-typeahead widget (#1011)

### LINKED ISSUE #1011: Multi select typeahead
# Multi select typeahead

We should create a multi-select typeahead widget which incorporates both the typeahead widget and the `chip` widget to create a composite widget.
This will allow users to make multiple selections from a typeahead menu represented as closable chips.
This should be similar to the way that you enter email addresses into the `to` field of an email and be material themable. An example of this can be seen [here](https://material.angularjs.org/latest/demo/chips).

rorticus:
**Type:** bug / feature

**Description:**

Adding `MultiSelectTypeahead` widget for selecting multiple values! 

Untitled
Untitled2

Resolves #1011 , #1346


src/multi-select-typeahead/index.tsx
```diff
@@ -0,0 +1,263 @@
...
+	value?: string[];
+	/** Property to determine if the input is disabled */
+	disabled?: boolean;
+	/** Property to determine how many items to render. Defaults to 6 */
+	itemsInView?: number;
+	/** The name of the input */
+	name?: string;
+	/** Placement of the select menu; 'above' or 'below' */
+	position?: PopupPosition;
+	/** Placement of the selected values. Default is 'bottom' */
```
tomdye, line 35: default is inline, not bottom

src/multi-select-typeahead/index.tsx
```diff
@@ -0,0 +1,263 @@
...
+						}
+					}}
+				>
+					{label}
+				</Label>
+			)}
+			<Typeahead
+				key="typeahead"
+				theme={theme.compose(
+					css,
```
tomdye, line 186: should this have a prefix

tomdye: Is this the correct order? I thought it was `baseCss, overrideCss`

src/multi-select-typeahead/index.tsx
```diff
@@ -0,0 +1,263 @@
...
+				values.length > 0 ? themeCss.hasValue : null,
+				focused ? themeCss.focused : null,
+				label ? themeCss.hasLabel : null
+			]}
+		>
+			{label && (
+				<Label
+					focused={focused}
+					active={active}
+					theme={theme.compose(
```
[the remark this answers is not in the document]
rorticus: I wasn't sure. I thought we only used a prefix if we have multiples of the same widget (i.e., multiple text inputs could be distinguished by prefix). That being said, there are a number of widgets in here (label, chip, typeahead, icon), so i can see prefixing them just to keep the classes unique.

src/multi-select-typeahead/index.tsx
rorticus: Nope! Good catch. Flipped all the theme.composes

src/multi-select-typeahead/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+	position?: PopupPosition;
+	/** Placement of the selected values. Default is 'inline' */
+	placement?: 'inline' | 'bottom';
+}
+
+export interface MultiSelectTypeaheadChildren {
+	/** Adds a <label> element with the supplied text */
+	label?: RenderResult;
+	/** Custom renderer for item contents */
+	items?: (properties: ItemRendererProperties) => RenderResult;
```
tomdye, line 43: could you rename this to `item` as it's a single item renderer

PR #1375, opened 2020-04-13
Optionally controlled checkbox group

### LINKED ISSUE #1331: Allow checkbox group to be controlled
## Proposal

## New Properties

- `value: string[]`

ericos:
**Type:** feature

**Description:**

Adds `value` property and allows Checkbox Group to be controlled.

Resolves #1331

### DIFF
--- a/src/checkbox-group/middleware.tsx
+++ b/src/checkbox-group/middleware.tsx
@@ -25,6 +25,14 @@ export const checkboxGroup = factory(({ middleware: { icache } }) => {
 				);
 				icache.set('initial', initialValue);
 			}
+		} else {
+			icache.set(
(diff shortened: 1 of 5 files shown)


src/checkbox-group/middleware.tsx
```diff
@@ -39,7 +41,13 @@ export const checkboxGroup = factory(({ middleware: { icache } }) => {
 
 		return (key: string) => ({
 			checked(checked?: boolean) {
-				const values = icache.getOrSet('values', {});
+				const values =
```
tomdye, line 44: could this be done when it's checked to see if `value` is `undefined` rather than every time `checked` is called?

PR #1384, opened 2020-04-15
V7 migration guide improvements

### LINKED ISSUE #1140: Migration guide: v7
A full migration guide will be needed for widgets that have been updated for the upcoming version 7 release. Although we have brand new documentation thanks to `@dojo/parade` much of the migration guide will be explaining the changes required (with examples) for upgrading  from version 6 to version 7.

ericos:
**Type:**  feature

**Description:**

**In Progress** -- Adds some missing information to the v7 migration guide. Formatting updates.

Resolves #1140

### DIFF
--- a/docs/V7-Migration-Guide.md
+++ b/docs/V7-Migration-Guide.md
@@ -455,28 +455,91 @@ Latest example can be found at [widgets.dojo.io/#widget/label/overview](https://
 
 ---
 
-### List
-- Renamed from `ListBox`
+### ListBox
+- Rewritten and renamed to `List`
+- Now uses a `resource` and the `data` middleware.
 
 #### Property changes
 ##### Additional Mandatory Properties
-- `foo: string;`
-	- this prop does x
+- `onValue(value: string)`
+	- this is the callback on value change.
+- `resource`
+	- this is the data resource for the List to use to generate the list items
+- `transform`
+	- a transform is required for the List as it uses a typed data middleware.
+	- a `defaultTransform` is exported from `List`
 
 ##### Changed properties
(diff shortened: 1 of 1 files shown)


tomdye:
@ericos looks like you added a fair few new-lines to this document. Was that intentional? I don't think they should be between the imports etc in the examples.
I also think that as this is a migration guide we should be writing it in the perspective of a v6 user. ie. Talk about the change to the original widget rather than how the new differs from the old.
If you could fix up the spacing, we can get the other edits merged in.

ericos:
> @ericos looks like you added a fair few new-lines to this document. Was that intentional? I don't think they should be between the imports etc in the examples.
> I also think that as this is a migration guide we should be writing it in the perspective of a v6 user. ie. Talk about the change to the original widget rather than how the new differs from the old.
> If you could fix up the spacing, we can get the other edits merged in.

Ah, hadn't pushed my latest changes that cleaned up the extra spacing. Just pushed that up with a few additional fixes.

docs/V7-Migration-Guide.md
```diff
@@ -382,185 +427,274 @@ The `Header` no longer auto-collapses action items into a `SlidePane`. Instead,
...
+##### v7 Example
+```tsx
+<Label valid={false}>Invalid Label</Label>
+```
+
+Latest example can be found at [widgets.dojo.io/#widget/label/overview](https://widgets.dojo.io/#widget/label/overview)
+
+---
 
+### List
```
ericos, line 458: List content needs to be filled out once all changes to that widget are complete.

PR #1388, opened 2020-04-15
Do not animate initial render of TitlePane

### LINKED ISSUE #1386: Titlepane animates on initial render
**Bug** 

When accordions / titlepanes perform their initial render they animate open / closed according to the value they'd been passed. This looks a bit jarring and they should really just render initially in the proposed state.

example:
https://widgets.dojo.io/#widget/accordion/overview

maier49:
**Type:** bug

**Description:**
Stores `'open'` or `'closed'` to depending on whether the pane is initially open or closed. Once a new value is achieved (by being passed or internally toggled) this stored value will be changed to `true`. If the value is `true` then the `contentTransition` class is applied.

Resolves #1386


maier49:
@tomdye Sometimes (like in our accordion pane example) the render gets called more than once so you end up seeing the transition anyway if you do it that way.

tomdye:
@maier49 this appears to still be broken for material theme

PR #1389, opened 2020-04-15
Updating List item renderer to not include ListItem

rorticus:
**Type:** bug / feature

**Description:**

- Moving `ListItem` and `MenuItem` to `List`
- Removed `ListItem` and `MenuItem` encapsulation from custom renderer in list
- Updated types of widgets that use List item renderer
- Updated examples


src/multi-select-typeahead/index.tsx
```diff
@@ -225,35 +234,40 @@ export const MultiSelectTypeahead = factory(function MultiSelectTypeahead({
...
-										classes={{
-											'@dojo/widgets/icon': {
-												icon: [themeCss.selectedIcon]
-											}
-										}}
-									/>
-								) : null}
-								{item.label || item.value}
-							</div>
+							<ListItem {...props}>
```
tomdye, line 251: wouldn't you pass you `selected` value to the `ListItem` here too?

PR #1397, opened 2020-04-16
Floating action button

### LINKED ISSUE #1382: Create Floating-Action-Button widget (FAB).
**Enhancement** 

Material / Android make frequent use of FAB. We should consider adding this to Dojo with corresponding theme implementations.

Inspiration:
- https://material-components.github.io/material-components-web-catalog/#/component/fab
- https://material-ui.com/components/floating-action-button/
- https://material.io/components/buttons-floating-action-button

maier49:
**Type:** feature

**Description:**
Adds a FAB component. The FAB component also provides a FAB icon that applies the proper font size for the icon so that it can be used alongside a label in the extended version without having to override the fontsize for the label or introduce complexity to child renderers.
Resolves #1382

### DIFF
--- a/src/theme/dojo/variants/default.m.css
+++ b/src/theme/dojo/variants/default.m.css
@@ -68,4 +68,6 @@
 	--avatar-size-large: 56px;
 
 	--tab-width: calc(var(--grid-base) * 16);
+
+	/* Floating action button */
 }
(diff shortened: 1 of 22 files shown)


src/theme/dojo/variants/default.m.css
```diff
@@ -68,4 +68,6 @@
 	--avatar-size-large: 56px;
 
 	--tab-width: calc(var(--grid-base) * 16);
+
+	/* Floating action button */
```
ericos, line 72: Remove?

tomdye, line 72:
```suggestion

	/* Floating action button */
```
```suggestion
```

PR #1409, opened 2020-04-21
Speed dial

### LINKED ISSUE #1383: Create speed-dial widget using FAB
**Enhancement** 

We should create a speed-dial style widget. This would rely on a FAB button being implemented and would display further options when hovered or clicked. This could be used in such implementations as a `share` menu

Inspirations:
- https://material-ui.com/components/speed-dial/
- https://material.io/components/buttons-floating-action-button/#types-of-transitions

Blocked by: #1382

maier49:
**Type:** feature

**Description:**
Adds a speed-dial widget that displays several related FABs when moused over. Also includes a change to FAB to support a "small" size needed for the speed dial. 
Resolves #1383


src/examples/src/widgets/floating-action-button/Small.tsx
```diff
@@ -0,0 +1,12 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import FloatingActionButton, { Icon } from '@dojo/widgets/floating-action-button';
+
+const factory = create();
+
+export default factory(function Small() {
+	return (
+		<FloatingActionButton size="small">
```
tomdye, line 8: can you wrap these in an `Example` widget please as per the other examples?

src/theme/default/speed-dial.m.css
```diff
@@ -0,0 +1,54 @@
...
+.action1 {
+}
+/* Class applied to the third action when closing, and the third to last when opening */
+.action2 {
+}
+/* Class applied to the fourth action when closing, and the fourth to last when opening */
+.action3 {
+}
+/* Class applied to the fifth action when closing, and the fifth to last when opening */
+.action4 {
```
tomdye, line 53: does this mean you're limited to 5 actions?

tomdye: can this not just be an `action` class and the user could use an `n-th` selector perhaps?

maier49: It doesn't limit you to 5 actions, this is a ramping animation delay to give the effect of the actions rolling out. After the 6th action (the first action has no delay) they will all just use the `action5` class. The order is then reversed when closing.

tomdye:
Closing in favour of https://github.com/dojo/widgets/pull/1456

PR #1420, opened 2020-04-22
fix: support custom TypeAhead values

### LINKED ISSUE #1416: TypeAhead throws when enter is pressed and there is no match
**Bug** 

When using a TypeAhead, then I type in a word not in the resource data and hit Enter I get an exception

Package Version: widgets 7.alpha.4 / framework alpha.16

**Code**

https://widgets.dojo.io/#widget/typeahead/overview

**Expected behavior:**

I would think that `onValue` should receive the value w/ some indicator whether it's a value not in the list and some way to accept/clear the value if it's valid. I'm not exactly sure what the general behavior should be, just that sometimes I might want to use an arbitrary value and sometimes not.

**Actual behavior:**

image

bitpshr:
**Type:** bugfix

**Description:**

This pull request updates the `TypeAhead` widget so that `onValue` is called if a user explicitly hits "enter" after typing in a custom value. The `onValue` callback now receives a flag indicating whether or not the value is custom so consumers can take (or not take) appropriate action.

Resolves #1416 

**Preview:**

preview

### DIFF
--- a/src/typeahead/index.tsx
+++ b/src/typeahead/index.tsx
@@ -1,32 +1,33 @@
-import { create, tsx } from '@dojo/framework/core/vdom';
-import { PopupPosition } from '@dojo/widgets/popup';
 import { RenderResult } from '@dojo/framework/core/interfaces';
-import { createICacheMiddleware } from '@dojo/framework/core/middleware/icache';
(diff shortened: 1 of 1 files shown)


bitpshr:
Superseded by #1432.

PR #1431, opened 2020-04-27
fix: simplify form examples

bitpshr:
**Type:** bug

**Description:**

This pull request simplifies all form examples (and by side effect, fixes styling on mobile) by removing the middle name field. After checking each example out, the middle name field wasn't demonstrating any behavior not demonstrated elsewhere, and both the boilerplate code and the example visuals were getting a bit cramped with four fields, three of which were now on the same row.

References #1419

**Preview:**

preview


bitpshr:
Superseded by https://github.com/dojo/widgets/pull/1437.

PR #1432, opened 2020-04-27
Allow typeahead to take free text

### LINKED ISSUE #1416: TypeAhead throws when enter is pressed and there is no match

tomdye:
**Type:** feature

**Description:**

- Adds `strict` property to `Typeahead` and `ChipTypeahead`.
- Allows non-resource values to be entered.
- Adds examples
- Adds tests

Resolves #1416

### DIFF
--- a/src/typeahead/index.tsx
+++ b/src/typeahead/index.tsx
@@ -145,8 +145,13 @@ export const Typeahead = factory(function Typeahead({
 
 	function callOnValue(value: string) {
 		const { onValidate, onValue, required } = properties();
-		let valid = required ? true : undefined;
+		const lastValue = icache.get('lastValue');
 
+		if (lastValue === value) {
+			return;
+		}
+
+		let valid = required ? true : undefined;
 		if (required && !value) {
 			valid = false;
 		}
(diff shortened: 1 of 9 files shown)


src/typeahead/index.tsx
```diff
@@ -170,17 +187,27 @@ export const Typeahead = factory(function Typeahead({
...
+						if (!disabled) {
+							icache.set('value', activeItem.value);
+							onClose();
+							callOnValue(activeItem.value);
+						}
+					} else {
+						if (strict) {
+							icache.set('valid', false);
+							const { onValidate } = properties();
+							onValidate && onValidate(false);
```
bitpshr, line 206: I sort of feel like if we're in strict mode and a user hits "enter" and no values are found, we should just no-op rather than failing validation.

tomdye: I've added an `if` so that this only happens if it's required

PR #1434, opened 2020-04-27
Updating chip typeahead to not select values on loss of focus

rorticus:
**Description:**

When the Typeahead loses focus, if not in strict mode, it's value gets passed out on the onValue callback. In the ChipTypeahead, this means that if you type something in, and then click outside the typeahead, the value in the text box gets added as a chip. This can be confusing if you type something in and then click on Chip thats part of the same control.

I'm not sure if this is the right fix or not, but this simply ignores values from the Typeahead if they happened from a blur event.

### DIFF
--- a/src/typeahead/index.tsx
+++ b/src/typeahead/index.tsx
@@ -293,13 +293,14 @@ export const Typeahead = factory(function Typeahead({
 								onBlur={() => {
 									const { onBlur } = properties();
 
+									onBlur && onBlur();
+
 									if (!strict) {
 										const value = icache.getOrSet('value', '');
 										callOnValue(value);
 									}
 
 									closeMenu();
-									onBlur && onBlur();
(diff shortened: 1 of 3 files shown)


src/typeahead/index.tsx
```diff
@@ -293,13 +293,14 @@ export const Typeahead = factory(function Typeahead({
 								onBlur={() => {
 									const { onBlur } = properties();
 
+									onBlur && onBlur();
+
 									if (!strict) {
 										const value = icache.getOrSet('value', '');
 										callOnValue(value);
```
tomdye, line 300: surely it's this part you want to remove @rorticus ? so it doesn't `callOnValue` when the typeahead `blur`s?

rorticus:
Don't think this is quite what we want. Closing to get something better.

PR #1435, opened 2020-04-28
Example should start `checked`

sebilasse:
**Type:** bug 

**Description:**

Example description says
> Sample checkbox that starts checked

But checked was `false`


tomdye:
good catch, thank you

PR #1437, opened 2020-04-28
feat: responsive form fields

bitpshr:
**Type:** feature

**Description:**

This pull request updates the `FormGroup` and `FormField` widgets so they have a minimum width and wrap in a responsive manner based on the size of their parent.

Reference #1419

**Preview:**

preview

### DIFF
--- a/src/theme/material/form.m.css
+++ b/src/theme/material/form.m.css
@@ -5,6 +5,7 @@
 .groupRoot {
 	display: flex;
 	flex-direction: row;
+	flex-wrap: wrap;
 	left: calc(-1 * var(--spacing));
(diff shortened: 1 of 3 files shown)


src/theme/material/form.m.css
```diff
@@ -30,4 +31,5 @@
 .fieldRoot {
 	flex: 1;
 	padding: 0 var(--spacing);
+	min-width: calc(20 * var(--spacing));
```
maier49, line 34: Is 16px * 20 the right min size? This collapses the three text input row to two rows even at full width for me.

bitpshr: Ah, good catch. Updated.

PR #1454, opened 2020-05-05
Snap List Scroll via calculated scroll top attribute

### LINKED ISSUE #1450: Snap scroll to items on the List
# Enhancement

When scrolling through the list,  it would be great if the scroll could snap to the nearest item.
Similar effect can be seen here: https://codepen.io/maxakohler/full/JBjROd

tomdye:
**Type:** feature

**Description:**

Implements scroll snap via calculated scroll top height. 
The intention was to accomplish this via css `scroll-snap` but this does not appear to work as we are manually setting both `scrollTop` and `translate-y` to facilitate our virtual scrolling window.
Raising this mainly to get feedback on how this feels / performs, it may be that it's too 'janky' and we simply remove it.

Feedback appreciated.

Resolves #1450

### DIFF
--- a/src/list/index.tsx
+++ b/src/list/index.tsx
@@ -602,8 +602,14 @@ export const List = factory(function List({
 			scrollTop={scrollTop}
 			onscroll={(e) => {
 				const newScrollTop = (e.target as HTMLElement).scrollTop;
-				if (scrollTop !== newScrollTop) {
-					icache.set('scrollTop', newScrollTop);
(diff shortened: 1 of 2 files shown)


tomdye:
Example here: https://dojowidgets-git-fork-tomdye-scroll-snap-list.dojo1.now.sh/#widget/list/example/itemrenderer

tomdye:
Closing as this breaks keyboard navigation and is very jarring when used with a track pad and intertia scrolling.

PR #1459, opened 2020-05-07
fix: wrap list item result text

### LINKED ISSUE #1457: List item text should not wrap
**Enhancement** 

Currently, when the list item text is too long it will wrap, we should change this to use an ellipsis instead.

Example of existing behavior:
Screenshot 2020-05-05 at 17 58 41

bitpshr:
**Type:** bugfix

**Description:**

This pull request adds list item text wrapping to the Dojo and Material themes.

Resolves #1457 

**Preview:**

preview

### DIFF
--- a/src/theme/dojo/list-item.m.css
+++ b/src/theme/dojo/list-item.m.css
@@ -4,6 +4,9 @@
 	color: var(--color-text-primary);
 	font-family: var(--font-family);
 	cursor: pointer;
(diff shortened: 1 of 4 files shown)


tomdye:
@bitpshr you need to do `menu-item` too I believe

bitpshr:
@tomdye good catch.

menu-item

PR #1460, opened 2020-05-07
Star Rating Widget

tomdye:
**Type:** new widget

**Description:**

Adds a rate widget with customisable icons

### DIFF
--- a/src/rate/index.tsx
+++ b/src/rate/index.tsx
@@ -1,11 +1,9 @@
 import { focus } from '@dojo/framework/core/middleware/focus';
 import { create, tsx } from '@dojo/framework/core/vdom';
-
-// import { formatAriaProperties } from '../common/util';
 import theme from '../middleware/theme';
 import * as css from '../theme/default/rate.m.css';
+import * as fixedCss from './rate.m.css';
 import * as baseCss from '../common/styles/base.m.css';
-// import Label from '../label';
 import RadioGroup from '../radio-group';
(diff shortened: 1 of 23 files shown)


src/rate/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+	/* The form name for this rate widget */
+	name?: string;
+	/* Flag to indicate if half stars should be used */
+	allowHalf?: boolean;
+	/* Flag to set the readonly state */
+	readOnly?: boolean;
+}
+
+export interface RateChildren {
+	label?: RenderResult;
```
maier49, line 30: If we make a new parade release, jsdoc comments on these fields will be shown in a `children` table. Do we want to start adding them generally?

maier49 (approved):
Looks good to me. Just had one question about child interface comments going forward

src/rate/tests/Rate.spec.tsx
```diff
@@ -0,0 +1,651 @@
...
+import Rate from '../index';
+import RadioGroup from '../../radio-group';
+
+import * as fixedCss from '../rate.m.css';
+import * as css from '../../theme/default/rate.m.css';
+import * as baseCss from '../../common/styles/base.m.css';
+
+import { Icon } from '../../icon';
+
+const noop = () => {};
```
msssk, line 16: this is so ubiquitous that it's exported by `common/tests/support/test-helpers`

src/rate/index.tsx
tomdye: good shout

PR #1462, opened 2020-05-11
WIP: Upgrade usages of Dojo Resources

agubler:
**Type:** feature

**Description:**

Upgrades widget resources usages for the latest proposed version.

**Currently WIP, requires changes reviewing, merged and released from framework**


src/list/index.tsx
```diff
@@ -404,7 +377,9 @@ export const List = factory(function List({
 					'item'
 				)}
 			>
-				<LoadingIndicator />
+				<div styles={{ height: '26px', paddingTop: '13px' }}>
```
agubler, line 380: @tomdye Can you suggest the correct way to do this? The loading indicator at the moment does not display correctly as it does not match the height of the calculated item height.

src/list/index.tsx
```diff
@@ -309,75 +310,47 @@ export const List = factory(function List({
...
-		const pages = [];
-		const loading = [];
-		getOrRead(getOptions());
-		const total = getTotal(getOptions());
-		if (total) {
-			for (let i = 0; i < Math.min(total - start, count); i++) {
+		const metaInfo = meta(options(), true);
+		if (metaInfo && metaInfo.total) {
+			const pages: number[] = [];
+			for (let i = 0; i < Math.min(metaInfo.total - start, count); i++) {
```
agubler, line 335: I shouldn't need to loop through simply to work out the pages required.

src/list/index.tsx
tomdye: I reckon you could just set the height to the calculated `itemHeight` ?

Issue #1223, opened 2020-03-30
Widget 7 Patterns


tomdye:
# Widget 7 Patterns

## Children and Child renderers

If a node is rendered inside a widget, it should be passed as a child rather than via a child renderer. This child could be any `RenderResult`. 

If the widget doesn’t need to determine where the children are rendered, or when they are rendered in a single location, then normal children are fine.

```tsx
<ParentWidget>
	<ChildWidget />
</ParentWidget>

<ParentWidget>
	<ChildWidget />
	<ChildWidget />
	<ChildWidget />
</ParentWidget>

<ParentWidget>
	Hello, World
</ParentWidget>
```

When a widget needs to inject functions or properties into the child widgets, a child function should be used:

```tsx
<ParentWidget>
	{ (foo) => <ChildWidget foo={foo('a')} /> }
</ParentWidget>
```

When multiple children are accepted and are to be placed in different locations by the parent widget, a child object should be used. This approach is use for example in our `Card` widget as it accepts content for different sections of the card and renders them in the appropriate locations with wrapping styles / classes. The child object can contain a mix of both `RenderResult` and functions that return a `RenderResult`.

```tsx
<ParentWidget>{{
	foo: (foo) => <ChildWidget foo={foo('a')} />,
	bar: <span>bar</span>,
	baz: 'hello, world'
}}</ParentWidget>
```

## Partial Control

Widgets should work out of the box wherever possible and be easy to use without needing to be fully controlled. As a result, we have re-written most of our widgets to use the `icache` middleware and maintain their own state. 
Form input widgets should accept an `initialValue` which can be used to control the `value` but does not need to be set each time an input calls its `onValue` callback.
In the case of widgets that hide / show such as dialogs or sidepane, control over the `open` and `requestClose` properties will still need to be maintained by the parent.

The lower level inputs, text-input, radio and checkbox are the exceptions to the "uncontrolled" approach as they are generally used as building blocks for more specialised widgets. For example, text-input allows you to control validation whereas the email-input, which utilises a text-input internally, is uncontrolled and manages its own validation state.
In the same fashion, checkboxes and radio buttons are fully controlled but we anticipate them to be used most frequently via checkbox-group and radio-group, both of which are partially controlled and follow the same `initialValue` pattern as other input widgets.

```tsx
// controlled input
let value = 'foo';
<TextInput value={value} onValue={(newValue) => { value = newValue; invalidator(); }} />

// uncontrolled input
<TextInput initialValue={'foo'} />
[...]

PR #1464, opened 2020-05-12
Initial Readme updates

tomdye:
**Type:** feature

**Description:**

Initial work on the widgets readme.
- Adds outline for upcoming sections
- Removes deprecated info re. class extension
- Changes examples refs to parade
- Change themes repo refs 
- Moves icon info to icon readme
- Removes list of widget readme
- Updates licence year to 2020

Work towards https://github.com/dojo/widgets/issues/1223

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,8 @@ These widgets are built using Dojo's widget authoring system [(@dojo/framework/c
 
 - [Usage](#usage)
 - [Features](#features)
+- [Examples](#examples)
 - [Conventions](#conventions)
-  - [Icons](#icons)
-  - [Coding conventions](#coding-conventions)
 - [How Do I Contribute?](#how-do-i-contribute)
     - [Installation](#installation)
     - [Testing](#testing)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -9,9 +9,8 @@ These widgets are built using Dojo's widget authoring system [(@dojo/framework/c
 
 - [Usage](#usage)
 - [Features](#features)
+- [Examples](#examples)
```
jellison, line 12: What if we have two top-level headings targets towards consumers and the other targeted towards contributors. Something like `Using Widgets` and `Contributing`? As a consumer, I usually stop reading once I've reached the part of the README that pertains to contributors.

README.md
```diff
@@ -50,112 +49,158 @@ This allows our [`dojo cli`](https://github.com/dojo/cli) build tooling to make
...
-- All widgets are fully themeable. Example themes are available in the [@dojo/themes](https://github.com/dojo/themes) repository.
+- All widgets are fully themeable. Example themes are available in the [/theme](https://github.com/dojo/widgets/tree/master/src/theme) folder.
 
 - All widgets support internationalization (`i18n`)
 
-## Widgets
+## Examples
 
-Live examples of current widgets are available in the [widget showcase](https://dojo.github.io/examples/widget-showcase/).
+Live examples of current widgets are available at [widgets.dojo.io](https://widgets.dojo.io ).
```
jellison, line 58: Now that we have Parade, I think we should expand this to `Documentation` and point out that we have examples, API docs, theme docs, etc all available here. I mean, it is a list of widgets, but it's so much more.

README.md
```diff
@@ -50,112 +49,158 @@ This allows our [`dojo cli`](https://github.com/dojo/cli) build tooling to make
...
+- reading properties
 
-[Checkbox/Toggle](src/checkbox/README.md)
+### Event callbacks
+- defining callbacks
+- passing callbacks
+- using event callbacks
 
-[ComboBox](src/combobox/README.md)
+### Control patterns
```
jellison, line 72: This new section is super useful to everyone, not just consumers. What if this were moved up nearer the top to give consumers a view of the philosophy of how widgets have been developed.

PR #1466, opened 2020-05-12
Tree Widget

### LINKED ISSUE #1214: Tree widget
Similar to 

https://ant.design/components/tree/
https://github.com/react-component/tree/

jellison:
**Type:** feature

**Description:**

Tree widget

Resolves #1214 

image


src/tree/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+import { createICacheMiddleware } from '@dojo/framework/core/middleware/icache';
+import theme from '@dojo/framework/core/middleware/theme';
+import { RenderResult } from '@dojo/framework/core/interfaces';
+import Icon, { IconType } from '../icon';
+
+import * as css from '../theme/default/tree.m.css';
+import Checkbox from '../checkbox';
+
+export interface TreeNode {
+	content: string | RenderResult;
```
tomdye, line 11: We shouldn't be passing `RenderResult` as widget properties, they should only be children

src/examples/src/widgets/tree/Basic.tsx
```diff
@@ -0,0 +1,36 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import Tree from '@dojo/widgets/tree';
+
+const factory = create();
+
+export default factory(function Basic() {
+	const nodes = [
```
tomdye, line 7: I think we need to flatten this and look at working with IDs.

tomdye: something like `{id: string, content: string, children: string[]}[]`  or track the parent ID

src/examples/src/widgets/tree/Basic.tsx
```diff
@@ -0,0 +1,36 @@
...
+		{
+			content: 'parent 1',
+			expanded: true,
```
tomdye, line 10: I think that things like `expanded` / `selected` etc should be moved out of the data, perhaps array of ids allowing for both controlled / partially controlled scenario?

src/tree/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+	}
+	if (node.expanded !== undefined && node.expanded !== lastNode.expanded) {
+		icache.set('expanded', node.expanded);
+	}
+
+	const expanded = icache.getOrSet('expanded', node.expanded || false);
+	const checked = icache.getOrSet('checked', node.checked || false);
+	const isLeaf = !node.children || node.children.length === 0;
+	const isSelected = selectedNode && selectedNode === node;
+	const spacers = new Array(level).fill(<div classes={classes.spacer} />);
```
tomdye, line 135: this isn't going to work in IE11 without importing the array shim

src/tree/index.tsx
jellison: Replaced with `value: string` and a custom renderer via child on `Tree`.

src/examples/src/widgets/tree/Basic.tsx
jellison: I took the opposite approach: `parent?: string`. I think this better matches relational structures. Also used `value: string` as that is consistent with `ListOption`.

src/examples/src/widgets/tree/Basic.tsx
jellison: Implemented as `expandedNodes?: string[]` and `selectedNotes?: string[]`

src/tree/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+		onCheck,
+		onExpand
+	} = properties();
+	const [itemRenderer] = children();
+
+	// Merge controlled properties with our internally-tracked ones
+	diffProperty(
+		'expandedNodes',
+		(
+			{ expandedNodes: current }: TreeNodeProperties,
```
matt-gadd, line 204: if you pass `properties` to the second param of diff property you do not have to manually type this.

src/tree/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+			{ expandedNodes: next }: TreeNodeProperties
+		) => {
+			if ((current && current.includes(node.id)) || (next && next.includes(node.id))) {
```
matt-gadd, line 207: if you're going to use `node` here you need to take it off the properties passed into `diffProperty` not the main render block, because `diffProperty` is executed at a different point (before the render) so `node` can be stale.

src/tree/index.tsx
jellison: Good tip!

src/tree/index.tsx
jellison: Good point.

samends:
Thanks for your work @jellison! We're continuing the work on this widget in this PR https://github.com/dojo/widgets/pull/1516

PR #1469, opened 2020-05-14
Fixed broken Material tab theme

### LINKED ISSUE #1468: Closeable Tabs show `close` text in material theme
**Bug** 

In the material theme, text that reads `close` is visible next to the close icon.
This text should not be visible.

It would appear that the current implementation places the text inside a button then uses an `after` pseudo selector to add the icon then hides the text via `font-size: 0`. 
This should be replaced by a simple `Icon` widget usage with `altText` text reading `close`.

Screenshot 2020-05-13 20 28 15

jellison:
**Type:** bug

**Description:**
Replaced tab "close" with Icon:

Dojo Theme:
image

Material Theme:
image

Resolves #1468

### DIFF
--- a/src/theme/material/tab-container.m.css
+++ b/src/theme/material/tab-container.m.css
@@ -33,7 +33,7 @@
 	composes: nativeResetRoot from './button.m.css';
 	position: absolute;
 	top: 50%;
-	transform: translateY(-50%) scale(0.7);
+	transform: translateY(-50%);
 	right: 5px;
 	cursor: pointer;
 	padding: 1px 3px;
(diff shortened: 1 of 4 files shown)


src/theme/material/tab-container.m.css
```diff
@@ -33,20 +33,13 @@
 	composes: nativeResetRoot from './button.m.css';
 	position: absolute;
 	top: 50%;
-	transform: translateY(-50%);
+	transform: translateY(-50%) scale(0.7);
```
tomdye, line 36: could a `size` property be specified on the `Icon` instead perhaps?

jellison: I considered that, but the Dojo theme doesn't need it. For whatever reason, the Material Icon is just a little big.

PR #1470, opened 2020-05-18
Stepped wizard widget

### LINKED ISSUE #1125: Stepped Wizard Widget
We should create a stepped wizard allowing for progress through a task / journey to be displayed.
Drawing inspiration from :
- https://ant.design/components/steps/
- https://material.angular.io/components/stepper/overview

maier49:
**Type:** feature

**Description:**
Adds a stepped wizard widget with horizontal and vertical display modes
Resolves #1125


tomdye:
I've taken a look at the examples in parade and I think this could be made more compatible with custom elements by simply passing `<Step>` widgets that include the title/subtitle etc. I think that the `StepWizard` widget itself can handle the rendering / clicking etc of the avatars thus meanign there's no need to use a render function.

Something like

```tsx
<Wizard onStep={(index: number) => {}} steps={3} initalStep={1}>
	<Step>{
		title: 'foo',
		subtitle: <span>foo subtitle</span>
	}</Step>
	<Step>{
		title: 'bar',
		subtitle: <span>bar subtitle</span>
	}</Step>
</Wizard>
```

As per the speed-dial, the wizard can just iterate through it's children.
What do you think about renaming this to `Wizard` / `Step` also?

maier49:
@tomdye The naming makes sense. I'll make those changes

maier49:
@tomdye Updated 👍

Since we're rendering the avatar/icon ourselves I also made some changes to explicitly handle an error state as we're not passing it to a render function.

tomdye:
@maier49 can you disable the user interaction as well? ie. if the wizard steps are to be controlled by next / prev buttons or by a different process.
Also, I think the styling might need some work. It doesn't look quite right just yet with the text wrapping and no differentiation between title/subtitle etc.

maier49:
@tomdye I added a `clickable` property that defaults to `true` that turns off the pointer styles and ignores clicks if set to `false`. I also made another pass and I think the text and icons should be much closer to the ant version in terms of weight/size/opacity.

maier49:
@tomdye I've reworked the API, added some basic styling to the default theme, and commented the classes. Could you be any more specific on what styles need updates?

src/examples/src/widgets/wizard/Basic.tsx
```diff
@@ -0,0 +1,95 @@
...
+	]);
+	return (
+		<Example>
+			<Wizard
+				steps={steps}
+				onStep={(stepIndex) => {
+					icache.set('activeStep', stepIndex);
+					icache.set(
+						'steps',
+						steps.map((_, index) => {
```
tomdye, line 24: won't this overwrite any existing `errror` states?

src/wizard/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const factory = create({ theme, dimensions, resize }).properties<WizardProperties>();
+
+export default factory(function Wizard({
+	properties,
+	children,
+	middleware: { theme, dimensions, resize }
+}) {
+	const classes = theme.classes(css);
+	const { direction = 'horizontal', onStep, steps, clickable = true } = properties();
```
tomdye, line 71: do you think clickable should be the default?

src/wizard/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+	resize.get('root');
+	const width = dimensions.get('root').size.width;
+
+	const stepNodes = children();
+	const forceVertical = width < 200 * stepNodes.length;
```
tomdye, line 76: where does `200` come from here and what is the reason behind forcing it to be vertical?

src/wizard/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+	const stepWrappers = stepNodes.map((step, index) => [
+		<div classes={classes.stepIcon}>
+			<Avatar
+				theme={theme.compose(
+					avatarCss,
+					css,
+					'stepAvatar'
```
tomdye, line 84: could you make this a single word, perhaps just `avatar`

src/wizard/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+				)}
+				outline={Boolean(steps[index].status !== 'inProgress')}
+			>
+				{steps[index].status === 'complete' ? (
```
tomdye, line 88: could you make this a switch to avoid the double ternary?

src/wizard/tests/unit/Wizard.spec.tsx
```diff
@@ -0,0 +1,237 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+const { it, describe } = intern.getInterface('bdd');
+const { assert } = intern.getPlugin('chai');
+import testHarness from '@dojo/framework/testing/harness/harness';
```
tomdye, line 4: Do you think you could try using the new test renderer?
docs: https://next.dojo.io/learn/testing/introduction#using-assertions

`import renderer, { assertion } from '@dojo/framework/testing/renderer';`

tomdye: Ant recently re-wrote the `List` tests using the new test renderer, would be a good place to start looking: https://github.com/dojo/widgets/blob/master/src/list/tests/List.spec.tsx

src/examples/src/widgets/wizard/Basic.tsx
maier49: Yes, but there aren't any in this basic example

tomdye: fair enough, just thinking about the best thing to be showing users is all, it's not a bit issue

maier49: Fair enough, I can add an if to return `error` if there is one. No harm in doing so

src/wizard/index.tsx
maier49: That's reasonable, I'll flip this

src/wizard/index.tsx
maier49: Forcing it to be vertical is basically copying Ant design's approach to responsiveness, except they only do it when viewed on "mobile" (via emulator at least), otherwise the steps just become squished and ugly/unusable. 200 px is just about how much space a step with a label, subtitle, and description needs to look reasonable. Could allow that limit to be overridden with a property?

Also open to any other suggestions on making this properly responsive

src/wizard/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+				content = <Icon type="checkIcon" />;
+				break;
+			case 'error':
+				content = <Icon type="closeIcon" />;
+				break;
+			default:
+				content = String(index + 1);
+		}
+		return {
+			content: [
```
tomdye, line 98: How about using `<virtual>` here instead?

src/wizard/index.tsx
maier49: I've removed all responsive rendering logic from the widget. For now this will be the responsibility of the parent

tomdye (approved):

PR #1489, opened 2020-06-16
Allow resizing two column layout

### LINKED ISSUE #1479: Add resize capability to the two-column layout
## Enhancement 

In order to provide split pane capability to dojo 7 (which is currently missing) we should add column resize functionality to our [two column layout widge](https://github.com/dojo/widgets/blob/master/src/two-column-layout/index.tsx)t. This should be relatively straight forward and can use the approach seen here in this POC: https://github.com/tomdye/dojo-resize-toolbar/blob/master/src/widgets/SplitPane.tsx

The initial size should remain that which is used by the two-column layout presently.

### Proposed new properties

```ts
resize?: boolean // determines if the columns should be resizeable
```

maier49:
**Type:** feature

**Description:**
Adds the ability to resize the two column layout
Resolves #1479

### DIFF
--- a/src/theme/default/two-column-layout.m.css
+++ b/src/theme/default/two-column-layout.m.css
@@ -20,10 +20,13 @@
 	align-items: center;
 	justify-content: center;
 	cursor: col-resize;
-	transition: all var(--short-animation-duration) var(--short-animation-easing);
+	transition: all var(--short-animation-duration) var(--short-animation-easing)
+		var(--animation-delay);
 }
 
-.divider:hover {
-	border-left: 5px solid var(--component-color);
-	border-right: 5px solid var(--component-color);
+.divider:hover,
+.divider:focus,
+.divider:active {
+	border-left: 5px solid var(--component-color-hover);
+	border-right: 5px solid var(--component-color-hover);
 }
--- a/src/theme/default/two-column-layout.m.css
+++ b/src/theme/default/two-column-layout.m.css
@@ -4,3 +4,7 @@
 }
 .small {
 }
+.divider {
+}
+.thumb {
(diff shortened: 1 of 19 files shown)


src/theme/default/two-column-layout.m.css
```diff
@@ -4,3 +4,7 @@
 }
```
tomdye, line 4: Can you add some comments to these classes please and initial styling so that it works without a theme

tomdye:
@maier49 I've noticed that the mouse and the resize bar get out of sync when the scroll hits it's min / max. Do you think you could take a look at that.
With regards to the thumb for resizing, I'm not sure it _is_ the best way forward. Looks like others, including this simple react example use a border-left/right that animate in on hover instead? This would likely work better with other themes too and is less obtrusive.

 http://react-split-pane-v2.surge.sh/?SimpleExample

maier49:
@tomdye I think I've fixed that issue. It was because I was adding to delta to the actual element size instead of the cached size. So past the limit it wouldn't track those changes really and then when moving back it would shrink the element.

I've also updated the styles.

tomdye:
@maier49 do you think you could add a longer delay to the wider drag bar transition and also ensure that it stays visible when the bar is focused/active as wall as when hovered. Could you also increase the z-index of the divider line so that it shows above the wider hover border.

maier49:
@tomdye We don't have a delay on transitions anywhere else so I added a new variable for that, set it to .5s. I changed the hover style to also include focus/active states. And there isn't really a relevant z-index because it's just the background-color vs border-colors of the but I did adjust it so that the line is still visible.

PR #1494, opened 2020-06-23
Resource wrapper widget

### LINKED ISSUE #1481: Resource support for Custom Elements
**Bug**
At the moment it's impossible to use any widget that depends on a resource as a Custom Element, because the only way to create/pass a resource is via middleware from the parent. For Custom Elements we may need a special wrapper to make this more "dom" like.

maier49:
**Type:** bug

**Description:**
Adds a `Resource` widget which is intended to be used only as a custom element to wrap widgets that expect a resource. It can be seen in action here wrapping a list component:
https://resource-widget.vercel.app/

The changes to other widgets were needed because when rendering as a custom element the child widget renders once before the parent can edit its properties, which was accessing an undefined object in widgets that use a resource.
Resolves #1481

### DIFF
--- a/.dojorc
+++ b/.dojorc
@@ -43,6 +43,7 @@
 			"src/radio-group",
 			"src/raised-button",
 			"src/range-slider",
+			"src/resource",
 			"src/result",
 			"src/select",
 			"src/slide-pane",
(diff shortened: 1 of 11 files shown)


matt-gadd:
@maier49 this seems a bit intrusive on widget authoring. I couldn't really tell how a user would pass a resource template from html either? is this something we can potentially solve at the framework/element wrapper level without a wrapping widget?

maier49:
Closing in favor of attempting to resolve this directly in framework
https://github.com/dojo/framework/pull/815

PR #1502, opened 2020-07-11
add zh-CN bundle

### LINKED ISSUE #1503: Add zh-CN language bundles
Add zh-CN language bundles to dojo widgets that use i18n

xiaohulu:
**Type:** bug / feature

**Description:**

Add `zh-CN` and `zh-TW`  bundles.

Resolves #1503


tomdye:
Thanks for doing this @xiaohulu :)

xiaohulu:
@tomdye My pleasure :)

But codecov check failed.

tomdye:
not a problem re. code cov, we should probably exclude the language bundles anyway. I asked you a question on discord re. changing the locale name from `zh-cn` to `zh` so that it covers a wider range of chinese locales. Would that be possible or is the translation you created only valid for `zh-cn` and not for the wider `zh` ?

xiaohulu:
@tomdye `zh-cn` for Simplified Chinese, `zh-tw` for Traditional Chinese.

for example https://github.com/ant-design/ant-design/tree/master/components/locale

image

tomdye:
Thank you for your work on this @xiaohulu, we came across a quirk of cldr data resolution when looking into this for you. It seems that although the lang is `zh-CN`, the locale bundle needs to be specified as `zh`. I have updated this for you and changed the package-lock file to point to a fix we've made in `webpack-contrib`. This should now work with `zh-TW` as desired. Really appreciate your contribution

PR #1506, opened 2020-07-14
Fix aria attributes

maier49:
**Type:** bug

**Description:**
The changes to aria attribute types for dojo/framework#735 caused some type errors here and this fixes them.


src/button/index.tsx
```diff
@@ -80,7 +80,7 @@ export const Button = factory(function Button({
 			id={idBase}
 			focus={focus.shouldFocus()}
 			name={name}
-			type={type}
+			type={type as any}
```
tomdye, line 83: I'm not sure that this should be an `as any`?

src/calendar/tests/unit/DatePicker.spec.tsx
```diff
@@ -115,7 +115,7 @@ const expectedMonthPopup = function(open: boolean) {
 		<div
 			id=""
 			key="month-grid"
-			aria-hidden={`${!open}`}
+			aria-hidden={!open ? 'true' : 'false'}
```
tomdye, line 118: could this ternary be reversed so that you don't use the negative?

src/calendar/tests/unit/DatePicker.spec.tsx
```diff
@@ -138,7 +138,7 @@ const expectedYearPopup = function(
 	return (
 		<div
 			key="year-grid"
-			aria-hidden={`${!open}`}
+			aria-hidden={!open ? 'true' : 'false'}
```
tomdye, line 141: same comment re. negative

src/range-slider/index.tsx
```diff
@@ -155,9 +155,9 @@ export const RangeSlider = factory(function RangeSlider({
 	const getInputProperties = (isSlider1: boolean) => ({
 		...formatAriaProperties(aria),
 		'aria-describedby': isSlider1 ? minLabelId : maxLabelId,
-		'aria-invalid': valid === false ? 'true' : null,
+		'aria-invalid': (valid === false ? 'true' : undefined) as 'true' | undefined,
```
tomdye, line 158: is the `as` needed here?

src/tab-container/index.tsx
```diff
@@ -181,7 +181,7 @@ export const TabContainer = factory(function TabContainer({
 	];
 
 	let alignClass;
-	let orientation = 'horizontal';
+	let orientation: 'horizontal' | 'vertical' = 'horizontal';
```
tomdye, line 184: Should we create a type for this?

src/tab-container/tests/TabContainer.spec.tsx
```diff
@@ -61,8 +61,8 @@ const reverseOrientationTemplate = assertionTemplate(() => (
 
 const tabButtonProperties = {
 	'aria-controls': 'test-tab-0',
-	'aria-disabled': 'false',
-	'aria-selected': 'true',
+	'aria-disabled': 'false' as 'false',
```
tomdye, line 64: is this really necessary? Why does `false` need to be cast to `false`?

src/button/index.tsx
maier49: Yeah I'll change it to `as 'submit' | 'reset' | 'button'`. Should have commented on this but I was wondering what we thought about whether the `'menu'` type should be adopted into the framework level or if we it is correct to treat this as an exception considering there's no browser support for `type="menu"`.

tomdye: why does it need to be cast at all? is the `property` not typed already? regarding `menu`, is it a property we currently support? if not, we should add it to our button type typings

src/range-slider/index.tsx
maier49: Unfortunately, since TS doesn't infer the correct type here. I've changed it to use the exported `AriaAttributes` interface explicitly though.

src/tab-container/index.tsx
maier49: I changed it to use the existing property on the `AriaAttributes` interface

src/tab-container/tests/TabContainer.spec.tsx
maier49: Unfortunately, since TS doesn't infer the correct type here. I've changed it to use the exported `AriaAttributes` interface explicitly though.

maier49:
@tomdye If we want to use the `AriaAttributes` interface then this can't be merged until we have a version of framework that includes that. Additionally, if we want to support booleans to avoid the need to cast strings that would require a change to vdom functionality as well as types. So in the meantime I've addressed the feedback and converted this to a draft.

agubler:
Superseded by https://github.com/dojo/widgets/pull/1569

PR #1507, opened 2020-07-14
Updated popup widget to have positions left and right

### LINKED ISSUE #1414: Enhance Popups - More Positions and Better Mobile Support
**Enhancement**

Several widgets are currently based on the `Popup` or `TriggerPopup`, which are limited to `above`/`below` placements. This makes building robust, mobile-aware - widgets more difficult.

I propose these base widgets be enhanced in this way:
- Expand positioning to support a plethora of positions like https://ant.design/components/popconfirm/.
- Refactor the incoming `position` property to be absolute (as opposed to a recommendation). The base widgets should automatically find the best position unless a `position` is explicitly set. Currently, composite widgets try to override positioning (with relative or absolute positioning) to achieve a proper fit. This change will allow composite widgets to trust the positioning and focus on implementing support for possible positions.
- Automatically turn positional popups into "modals" for mobile, similar to https://material.io/components/dialogs#alert-dialog (horizontally & vertically centered).
- (Optional) Allow the consuming widget to specify which positions it wants to support (maybe, `['top', 'bottom', 'modal']` is all they want, but still want to take advantage of the internal "best fit" positioning).

samends:
**Type:** Feature

**Description:**
Adds the positions of `left` and `right` on the popup widget. When the popup doesn't fit to on the left it will default to right and vice versa with right.

Resolves #1414


src/popup/index.tsx
```diff
@@ -53,10 +56,19 @@ export const Popup = factory(function({
...
+	const widthOfScreen = document.documentElement.clientWidth;
 
 	const willFit = {
 		below: yTop + wrapperDimensions.size.height <= bottomOfVisibleScreen,
-		above: yBottom - wrapperDimensions.size.height >= topOfVisibleScreen
+		above: yBottom - wrapperDimensions.size.height >= topOfVisibleScreen,
+		left:
+			leftStart - wrapperDimensions.size.width >= 0 &&
+			yTop + wrapperDimensions.size.height / 2 <= bottomOfVisibleScreen &&
+			yBottom - wrapperDimensions.size.height / 2 >= topOfVisibleScreen,
```
maier49, line 67: Should this be 
```
(yTop + yBottom)/2 + wrapperDimensions.size.height / 2 <= bottomOfVisibleScreen &&
(yTop + yBottom)/2 - wrapperDimensions.size.height / 2 >= topOfVisibleScreen,
```

Right now I think it's calculating it as if the popup is cut in half with half on top of the trigger and half below it.

samends: Yea that's how I have it, so it looks like this
Screen Shot 2020-07-14 at 1 51 37 PM

I was following this design from the framework that was referenced in the original issue https://ant.design/components/popconfirm/. There was an idea of `bottom-right` and top-right` being added as well.

src/popup/index.tsx
```diff
@@ -19,12 +19,14 @@ export interface BasePopupProperties {
...
 
 export interface PopupProperties extends BasePopupProperties {
-	/** The X position on the page where the popup should render */
-	x: number;
 	/** The Y position on the page where the bottom of the popup should be if rendering "above" */
 	yBottom: number;
 	/** The Y position on the page where the popup should start if rendering "below" */
 	yTop: number;
+	/** The position on the page where the popup should start if rendering "right" */
+	rightStart: number;
```
tomdye, line 27: how about making these xStart / xEnd to reflect the yBottom / yTop properties. These would identify the left and right boundaries of the popup trigger / positioning.

src/popup/index.tsx
tomdye: I think that in the most case, top / bottom will be used and only left / right would be used for things like sub-menus or tooltips etc which I believe should be centred.

tomdye: Something to consider is that when placing left / right i think you only need to check the space to the left or right, not the vertical placement. When doing the actual placement though, it should be centred BUT should be bound by the top of the screen.

samends: When placing left and right there is a chance of the placement going off screen, which is why I was checking the vertical here, why do you think it doesn't need to be checked?

samends: Also what do you mean by being bound by the top of the screen?

tomdye: if you had, for example, a large sub menu coming off of the right side of another menu item you wouldn't want it rendered off of the top of the page, you'd fix it's upper bound to that of the visible window.

src/popup/index.tsx
```diff
@@ -82,6 +88,31 @@ export const Popup = factory(function({
 				wrapperStyles.top = `${yTop}px`;
 			}
 		}
+
+		if (position === 'left' || position === 'right') {
+			const triggerHeight = yTop - yBottom;
+			if (yBottom - wrapperDimensions.size.height / 2 >= topOfVisibleScreen) {
```
tomdye, line 94: could you use `Math.min` here?

samends: I used `Math.max` instead

tomdye: yeah, that one lol

PR #1509, opened 2020-07-15
Combine Calendar bundles and remove customisation properties

### LINKED ISSUE #1508: Calendar widget should not expose properties for overriding labels outside of i18n
**Fix** 

In the current `Calendar` widget, a user can override labels, month names and weeknames via properties.
These should be overridden via the `i18nBundle` property only. We should therefore remove these properties.

This will be a breaking change.

tomdye:
**Type:** feature

**Description:**

- Moves the calendar specific `common` i18n bundle values into the calendar bundle as this is the only place they're used
- Removes the customisation `properties` outside of the `i18nBundle` functionality
- Fixes label etc usage throughout the widgets used by `Calendar`
- Fixes tests
- Updates README

Resolves #1508


tomdye:
Closing to be re-raised with the prior nls combination pr

PR #1516, opened 2020-07-21
Tree widget

### LINKED ISSUE #1214: Tree widget

samends:
**Type:** Feature

Adds tree widget to widget project

Resolves #1214


src/tree/index.tsx
```diff
@@ -0,0 +1,382 @@
...
+	const [itemRenderer] = children();
+	const classes = theme.classes(css);
+	const isActive = node.id === activeNode;
+	const isSelected = node.id === selectedNode;
+	const expanded = expandedNodes.includes(node.id);
+	const checked = checkedNodes.includes(node.id);
+	const isDisabled = disabledNodes && disabledNodes.includes(node.id);
+	const isExpandable = node.hasChildren;
+	const spacers = new Array(depth);
+	fill(spacers, <div classes={classes.spacer} />);
```
tomdye, line 337: I think that this should be possible by simply putting a `margin-left`or `padding-left` on each of the nested `ol` nodes. We shouldn't be creating `spacer` divs like this.

src/tree/index.tsx
```diff
@@ -0,0 +1,382 @@
...
+
+/*******************
+ * Tree
+ *******************/
+
+export interface TreeNodeOption {
+	id: string;
+	parent: string;
+	value: string;
+	hasChildren: boolean;
```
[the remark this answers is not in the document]
samends: The `hasChildren` is mandatory in the current implementation. It used to show whether a node is expandable without having to make a query to find out if the node has children. The would be useful async query calls.

src/tree/index.tsx
```diff
@@ -0,0 +1,382 @@
...
+				nodes = [...nodes, ...createNodeFlatMap(node.id)];
+			}
+		});
+		return nodes;
+	}
+
+	function onKeyDown(event: KeyboardEvent) {
+		event.stopPropagation();
+		const nodes = createNodeFlatMap();
+		const activeIndex = nodes.findIndex((node) => node.id === activeNode);
```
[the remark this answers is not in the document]
samends: Oh are the nodes in resources flatted? I'll look into this

src/tree/index.tsx
```diff
@@ -0,0 +1,382 @@
...
+	parent: string;
+	value: string;
+	hasChildren: boolean;
+}
+
+export interface TreeProperties {
+	checkable?: boolean;
+	selectable?: boolean;
+	checkedNodes?: string[];
+	expandedNodes?: string[];
```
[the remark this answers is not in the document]
samends: Where else are you using `initialExpanded` or `initialChecked` in the widgets project?

src/tree/index.tsx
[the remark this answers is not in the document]
samends: The node flat map is also used further down to return the next node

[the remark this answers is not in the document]
samends: Looks like `find` does not give you the node index of a flattened array

src/tree/index.tsx
samends: I believe this was done as the further nested the node are, the more spacing they need. So the styling needed  correlate with the node depth.

src/tree/index.tsx
```diff
@@ -0,0 +1,382 @@
...
+
+import { Keys } from '../common/util';
+import Icon from '../icon';
+import Checkbox from '../checkbox';
+import { ListItem } from '../list';
+
+import * as css from '../theme/default/tree.m.css';
+import LoadingIndicator from '../loading-indicator';
+
+/*******************
```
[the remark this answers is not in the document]
samends: maybe that was an older convention? Took it out

src/tree/index.tsx
tomdye: given that each list will be within a parent list it should work fine

src/tree/index.tsx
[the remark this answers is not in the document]
tomdye: not expanded / checked necessarily, but we use the that as the partial control pattern for other widgets
for example `value` and `initialValue` here: https://github.com/dojo/widgets/blob/master/src/native-select/index.tsx#L21

tomdye:
@samends can you change the `glitch` to return the total for each of the queries rather than the total of all of the items please?

samends:
@tomdye I took out the pagination part in the `glitch` so the total is simply the length of the total returned items

src/examples/src/widgets/tree/Checkable.tsx
```diff
@@ -0,0 +1,66 @@
...
+	}
+];
+const template = createMemoryResourceTemplate<TreeNodeOption>();
+
+export default factory(function Checkable({ id, middleware: { resource } }) {
+	return (
+		<Example>
+			<Tree
+				initialExpanded={(id) => {
+					console.log('initial expanded', id);
```
tomdye, line 56: why is this a function and not just an array?

src/tree/index.tsx
```diff
@@ -0,0 +1,400 @@
...
+
+export interface TreeProperties {
+	checkable?: boolean;
+	selectable?: boolean;
+	checkedNodes?: string[];
+	expandedNodes?: string[];
+	selectedNode?: string;
+	disabledNodes?: string[];
+	parentSelection?: boolean;
+	initialExpanded?(id: string): void;
```
tomdye, line 32: Why are initialExpanded and initialChecked functions when the controlled versions of those are simply string arrays? They should both be string arrays.

src/examples/src/widgets/tree/Checkable.tsx
tomdye, line 56: shouldn't have console.logs in here anymore

src/tree/index.tsx
```diff
@@ -0,0 +1,400 @@
...
+		if (checked) {
+			hasBeenChecked === false && initialChecked && initialChecked(n);
+			icache.set('hasBeenChecked', true);
+		}
+		icache.set('checkedNodes', checkedNodes);
+		onCheck && onCheck(n, checked);
+	}
+
+	function expandNode(id: string) {
+		expandedNodes.push(id);
```
tomdye, line 141: it's better not to save references to icache values to be unsed inside callbacks etc as they can become stale and result in errors.
In the case of changing the `expandedNodes` cache we should be doing something like the below in order to ensure the latest values are used. Please ensure the same is done elsewhere.

```tsx
icache.set('expandedNodes', (currentExpanded) => [ ...currentExpanded, id ]);
```

tomdye: I think we need to be careful here with auto-expanding when the user has supplied an `expandedNodes` controlling property. If this is present then we should not be updating the cache here and visually expanding the node, instead should be simply calling out to the `onExpand` callback and reacting to the new property change as with all other properties when the parent updates the `expandedNodes` property. I think this may be the case with other controlled properties within this PR.

tomdye (changes requested):

src/tree/index.tsx
```diff
@@ -0,0 +1,422 @@
...
+	checkable?: boolean;
+	selectable?: boolean;
+	checkedNodes?: string[];
+	expandedNodes?: string[];
+	initialChecked?: string[];
+	initialExpanded?: string[];
+	selectedNode?: string;
+	disabledNodes?: string[];
+	parentSelection?: boolean;
+	onSelect?(id: string): void;
```
tomdye, line 34: I think this should probably be `value` / `initialValue` / `onValue` to mirror other widgets and the `List`

src/tree/index.tsx
```diff
@@ -0,0 +1,422 @@
...
+	return mapNodeTree();
+});
+
+interface TreeNodeProperties {
+	checkable: boolean;
+	selectable: boolean;
+	activeNode?: string;
+	selectedNode?: string;
+	checkedNodes: string[];
+	disabledNodes: string[];
```
tomdye, line 334: can you please drop `Nodes` naming from these

src/tree/index.tsx
```diff
@@ -0,0 +1,422 @@
...
+	expandedNodes: string[];
+	parentSelection?: boolean;
+	node: TreeNodeOption;
+	onActive(node: string): void;
+	onSelect(node: string): void;
+	onCheck(node: string, checked: boolean): void;
```
tomdye, line 340: If this is used when a tree node is created then a smple `onCheck` and `onExpand` calback should be enough, the parent knows which node its rendering, this is true for all of these callbacks that return a node / key

src/tree/index.tsx
```diff
@@ -0,0 +1,422 @@
...
+	selectable?: boolean;
+	checkedNodes?: string[];
+	expandedNodes?: string[];
+	initialChecked?: string[];
+	initialExpanded?: string[];
+	selectedNode?: string;
+	disabledNodes?: string[];
+	parentSelection?: boolean;
+	onSelect?(id: string): void;
+	onCheck?(id: string, checked: boolean): void;
```
tomdye, line 35: do you think `onCheck` and `onExpand` might be better reflecting their `checked` and `expanded` counterparts by simply calling the callback with the array of `checked` / `expanded` nodes? This would simplify the callback by providing symmetry with the value and reflect what ant-design does.

src/tree/index.tsx
```diff
@@ -0,0 +1,422 @@
...
+}
+
+export interface TreeProperties {
+	checkable?: boolean;
+	selectable?: boolean;
+	checkedNodes?: string[];
+	expandedNodes?: string[];
+	initialChecked?: string[];
+	initialExpanded?: string[];
+	selectedNode?: string;
```
tomdye, line 31: can you drop the `Nodes` naming from these please?

src/tree/index.tsx
```diff
@@ -0,0 +1,422 @@
...
+	}
+	return mapNodeTree();
+});
+
+interface TreeNodeProperties {
+	checkable: boolean;
+	selectable: boolean;
+	activeNode?: string;
+	selectedNode?: string;
+	checkedNodes: string[];
```
tomdye, line 333: i don't think that a tree node should know anything bout the selected nodes, active nodes etc, it should simply be passed a boolean if it's active / selected / checked etc. Please correct me if I am wrong.

tomdye (changes requested):

src/tree/index.tsx
samends: I feel `Nodes` here would be helpful for the user in understanding what this field is for, in addition, it makes the code a bit more understandable as there are other variables name `checked` and selected` that are used for other values.

src/tree/index.tsx
samends: hm I feel it depends on how the user of this widget wants to structure their code. By passing `checked` and `expanded` the user wouldn't have to have to keep track of these values in the parent widget, they would also have to use `find` if they wanted to find out which node was just checked. Overall I feel we can update this to pass `checked` and `expanded`

src/tree/index.tsx
samends: yup I agree here, it seems the only reason these values are being passed is to calculate `selected` `checked` etc

src/tree/index.tsx
samends: Not sure I'm clear on this, do you not want `onCheck` and `onExpand` to not pass node values?

src/tree/tests/unit/Tree.spec.tsx
```diff
@@ -0,0 +1,659 @@
...
+
+import { Keys } from '../../../common/util';
+import { stubEvent } from '../../../common/tests/support/test-helpers';
+import Tree, { TreeNode, TreeNodeOption } from '../../index';
+import * as css from '../../../theme/default/tree.m.css';
+
+import { simpleTree } from './TreeData.mock';
+import LoadingIndicator from '../../../loading-indicator';
+
+const noop = () => {};
```
msssk, line 21: this is exported from `common/tests/support/test-helpers`

src/tree/index.tsx
samends: Hm @tomdye how would this be different than `onCheck` and `onSelect`?

src/tree/tests/unit/Tree.spec.tsx
samends: done 👍

src/tree/index.tsx
```diff
@@ -0,0 +1,415 @@
...
+}
+
+export interface TreeProperties {
+	checkable?: boolean;
+	selectable?: boolean;
+	checked?: string[];
+	expanded?: string[];
+	initialChecked?: string[];
+	initialExpanded?: string[];
+	selectedNode?: string;
```
tomdye: I think that this should be renamed to `value` to align with the `onValue` callback

src/tree/index.tsx
```diff
@@ -0,0 +1,415 @@
...
+		if (!controlledCheckedNodes) {
+			if (checked) {
+				icache.set('checked', (currentChecked) => [...currentChecked, id]);
+			} else {
+				icache.set('checked', (currentChecked) =>
+					currentChecked ? currentChecked.filter((n) => n !== id) : []
+				);
+			}
+		}
+		onCheck && onCheck(icache.get('checked') || []);
```
tomdye, line 140: I don't think this is going to work because you're only going to update the `checked` if it's not controlled so the listener in the controlled scenario will only ever get the previous checked value.
How about you just set the icache without the check and then call the `onCheck` with the result. There's no need to check if it's controlled at this point as you'r already doing it in the next render function and thus deciding if you should use the icache value or the `checked` value you've been passed in properties. The same is true for expand / collapse functionality. You should always be updating the `icache` and calling back with the latest.

src/tree/index.tsx
```diff
@@ -0,0 +1,415 @@
...
+								node.hasChildren && classes.leaf,
+								selectable && classes.selectable,
+								node.id === selectedNode && classes.selected
+							]}
+						>
+							<TreeNode
+								activeNode={activeNode}
+								checkable={checkable}
+								selectable={selectable}
+								checkedNodes={controlledCheckedNodes || checkedNodes}
```
tomdye, line 288: I thought you were changing the `TreeNodeProperties` to simply be passed a `boolean` for `checked` / `selected` / `disabled` etc? is there a reason why it needs to know about all of the checked nodes etc rather than just about itself?

src/tree/index.tsx
```diff
@@ -0,0 +1,415 @@
...
+						<li
+							classes={[
+								classes.node,
+								node.hasChildren && classes.leaf,
+								selectable && classes.selectable,
+								node.id === selectedNode && classes.selected
+							]}
+						>
+							<TreeNode
+								activeNode={activeNode}
```
tomdye, line 285: same comment as the other properties, this should likely just be `active: boolean`

src/tree/index.tsx
```diff
@@ -0,0 +1,415 @@
...
+	const classes = theme.classes(css);
+	const defaultRenderer = (n: TreeNodeOption) => n.value;
+	const [itemRenderer] = children();
+
+	const activeNode = icache.get('activeNode');
+	const selectedNode = icache.get('selectedNode');
+	const controlledExpandedNodes = icache.get('controlledExpandedNodes');
+	const controlledCheckedNodes = icache.get('controlledCheckedNodes');
+	const expandedNodes = icache.getOrSet('expanded', []);
+	const checkedNodes = icache.getOrSet('checked', []);
```
tomdye, line 118: you should be able to determine if you're using the `icache` or properties to the controlled values and simply set them as `consts` here so that you don't need the `if controlledCheckedNodes` etc statements in the rest of the render function.

tomdye: something like the below, taken from `text-input`.

``` tsx
let { value } = properties();
const { initialValue } = properties();

if (value === undefined) {
		value = icache.get('value');
		const existingInitialValue = icache.get('initialValue');

		if (initialValue !== existingInitialValue) {
			icache.set('value', initialValue);
			icache.set('initialValue', initialValue);
			value = initialValue;
		}
	}
```

tomdye (changes requested):

src/tree/index.tsx
samends: If you set the icache here there would be no difference between `initialChecked` and `controlledCheckedNodes`. The way I pictured the difference between `initialChecked` and `controlledCheckedNodes` was that the user would be in control of the when the boxes get clicked, if you set the `icache` here the `Tree` would update the checkbox without the user explicitly saying so

src/tree/index.tsx
samends: I've updated these to be `booleans`

tomdye (approved):

PR #1540, opened 2020-08-27
Add FileUploader widget

### LINKED ISSUE #1129: Uploader widget
TBD

msssk:
**Type:** feature

**Description:** add `FileUploader` widget

Resolves #1129


tomdye:
I added comments to the issue re. the separation of the two widgets. I don't think this will need a child renderer at all as the file-list version of the file upload widget would be opinionated in it's appearance. The only child would likely be the `label` as per other widgets. You should be able to use `composes` in your css to utilise the existing button styles for each of the themes also.

src/file-uploader/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+import * as fileUploadInputCss from '../theme/default/file-upload-input.m.css';
+import * as fileUploadInputFixedCss from '../file-upload-input/styles/file-upload-input.m.css';
+
+export interface FileUploaderChildren {
+	label?: RenderResult;
+}
+
+export interface FileUploaderProperties extends FileUploadInputProperties {
+	/** Custom validator used to validate each file */
+	customValidator?: (file: File) => ValidationInfo | void;
```
tomdye, line 25: I'm not sure about this pattern, other widgets set the `valid` property when they want to control the validation and use the the `onValue` function to trigger said validation.

src/file-uploader/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+	/** The files to render in the widget (controlled scenario) */
+	files?: FileWithValidation[];
```
tomdye, line 28: Seems odd that the validation is being mixed into the files for all of these callbacks. Would have expected `onValue` etc and `files` to only take / pass `File[]` params.

PR #1550, opened 2020-10-07
feat: DraggableList v2

### LINKED ISSUE #1429: Drag and drop list
**Enhancement** 

We should implement a drag-and-drop list widget. Details TBD.

Package Version: latest

bitpshr:
**Type:** bug / feature

**Description:**

This pull request augments the existing resource-powered `List` widget by adding the ability to drag-and-drop reorder items. This work implements a completely different approach to that in #1547, the latter of which did not build upon `List` or use resources.

Resolves #1429

**Preview:**

preview

### DIFF
--- a/.dojorc
+++ b/.dojorc
@@ -24,6 +24,7 @@
 			"src/context-popup",
 			"src/date-input",
 			"src/dialog",
+			"src/draggable-list",
 			"src/email-input",
 			"src/floating-action-button",
 			"src/form",
--- a/.dojorc
+++ b/.dojorc
@@ -24,7 +24,6 @@
 			"src/context-popup",
 			"src/date-input",
 			"src/dialog",
-			"src/draggable-list",
 			"src/email-input",
 			"src/floating-action-button",
 			"src/form",
--- a/src/examples/src/widgets/list/Draggable.tsx
+++ b/src/examples/src/widgets/list/Draggable.tsx
@@ -1,45 +1,33 @@
-import { create, tsx } from '@dojo/framework/core/vdom';
-import List, { ListOption } from '@dojo/widgets/list';
 import icache from '@dojo/framework/core/middleware/icache';
-import Example from '../../Example';
 import {
 	createMemoryResourceTemplate,
 	createResourceMiddleware
 } from '@dojo/framework/core/middleware/resources';
+import { create, tsx } from '@dojo/framework/core/vdom';
+import List, { ListOption } from '@dojo/widgets/list';
+
+import Example from '../../Example';
 
 const resource = createResourceMiddleware();
 const factory = create({ icache, resource });
 
-const animals = [
-	{ value: 'cat' },
-	{ value: 'dog' },
-	{ value: 'mouse' },
-	{ value: 'rat' },
-	{ value: 'cat' },
-	{ value: 'dog' },
-	{ value: 'mouse' },
-	{ value: 'rat' },
-	{ value: 'cat' },
-	{ value: 'dog' },
-	{ value: 'mouse' },
-	{ value: 'rat' }
-];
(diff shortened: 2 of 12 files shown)


src/theme/material/list-item.m.css
```diff
@@ -17,4 +18,30 @@
 .active {
 	border: none;
 	background: var(--mdc-theme-on-surface);
+	height: 48px;
+	box-shadow: none;
+}
+
+.collapsed {
```
tomdye, line 25: is it possible to remove the `important` usage here by adding a `.root` selector before `.collapsed` instead?

src/theme/dojo/list-item.m.css
```diff
@@ -9,12 +9,13 @@
...
+	min-height: calc(var(--grid-base) * 5.5);
 	display: flex;
 }
 
 .active {
 	border: 1px solid var(--color-highlight-border);
+	transition: none !important;
 }
 
 .selected {
```
tomdye, line 21: I think that the dojo theme should have a `movedUp` / `movedDown` class as well as the sizes may change / differ from the default styling.

.dojorc
```diff
@@ -24,6 +24,7 @@
 			"src/context-popup",
 			"src/date-input",
 			"src/dialog",
+			"src/draggable-list",
```
tomdye, line 27: I don't believe this is needed anymore

src/examples/src/widgets/list/Draggable.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {
+	createMemoryResourceTemplate,
+	createResourceMiddleware
+} from '@dojo/framework/core/middleware/resources';
+
+const resource = createResourceMiddleware();
+const factory = create({ icache, resource });
+
+const animals = [
+	{ value: 'cat' },
```
tomdye, line 14: Could you remove the repetition from these values please so that the active states etc still work correctly.
It would also be good to demo the same functionality with a paginated list, perhaps generate a larger data set.

bitpshr:
Thanks for the review @tomdye, should be ready for another pass when there's time.

tomdye:
@bitpshr there appears to be some weirdness with the list now where it scrolls down when the mouse is over the bottom few items? Also noticed that if you scroll down a few pages and hold the mouse on an item to drag it, it immediately scrolls back upto the top of the list.

bitpshr:
Thanks for another pass @tomdye. The issue with the list scrolling to the top was due to faulty `activeIndex` management and should now be fixed. The issue with the list auto-scrolling when hovering over the last visible item was prexisting, and appears to be the result of shoddy `scrollTop` management. I'm happy to look into this issue separately from this pull request to keep reviews clean.

tomdye:
@bitpshr re. shoddy scroll top, i think it may be due to the height of items being calculated incorrectly once a theme has been applied and thus the overall height  / items-in-view being wrong.

src/theme/material/list-item.m.css
```diff
@@ -17,4 +18,30 @@
...
+	height: 0 !important;
```
tomdye, line 26: Were you able to try removing these `importants`?

tomdye (commented):
PR is looking great. Do you think the draggable items should perhaps be visibly different and have an indicator that they are draggable? could you also add `cursor: move` perhaps?

bitpshr:
Thanks @tomdye, drag icon with cursor styling added. Note that I didn't add `cursor: move` to the whole item (even though the whole item is actually draggable), as this made it seem as though items were not also selectable. After some testing, it felt like the whole item being draggable, coupled with a drag-styled icon with `cusor: move`, provides the right combination of usability and intuitive UI. Happy to revisit this as-needed though.

tomdye (approved):

PR #1561, opened 2020-11-17
Fix the filtering behaviour of resource widgets to use visible label

### LINKED ISSUE #1562: Typeahead, Select and List widget use value property to match input instead of label property
**Bug**

The Typeahead-, Select and List widgets operate on resources that provide ListOption items,
consisting of a "value" and "label" property. Currently the Typeahead widget (and apparently
Select and List as well) will always match the value, not the label, and there is no
configuration option for it.

In real world applications, matching the value very rarely makes sense. Objects in selectable
list usually have an identity property that is different from the textual representation, and the
user wants to type ahead to what he sees in the list.

A real workaround needs a complete custom resource, with its own finder and filter that can work
with value objects and strings. The other workaround - using the labels as values only works
when the labels are unique. Both workarounds are really ugly. IMO, this needs a fix that cannot
wait for Dojo 8.

Workaround in case of dire need:

Change the value field of your ListOptions from your id to an actual value object that contains
an id and the label.

Copy the code of memoryTemplate in resources.ts, modify the defaultFilter roughly as follows
(I eliminated the outer loop):

```javascript
  // at the start of your filter function

  let value = query['value'];
  if (typeof value !== 'string')
    // "name" property here, needs to be the property of your value object that contains the label.
    value = value.name;
  }

  const queryKey = "label";
  const itemValue = item[queryKey];
```

Note the workaround implies that the "value" property of the Typeahead widget will contain
a value object that represents the selection, rather than just a id.

Also note you need your resource implementation to make sure "===" between value objects
works as an equality comparison (so the same id needs to be represented by the same
value object instance. Also you need to use a different type for the ListOption where you set the
type of value to "any" instead of "string".

Package Version: @dojo/widgets 7.0.3 (same in master)

**Code**

Use the basic example from the widgets reference, code sandbox link:
https://codesandbox.io/s/github/dojo/widgets/tree/master/src/examples?fontsize=14&initialpath=%23/widget/list/sandbox/basic%3Ftheme%3Ddojo&module=/src/widgets/list/Basic.tsx&file=/src/widgets/list/Basic.tsx:421-512

Replace
```javascript
  const animals = [{ value: 'cat' }, { value: 'dog' }, { value: 'mouse' }, { value: 'rat' }];
```
with
```javascript
  const animals = [{ value: '4711', label: 'cat' }, { value: '4712', label: 'dog' }, { value: '4713', label: 'mouse' }, { value: '4714', label: 'rat' }];
```
**Expected behavior:**

The Typeahead shows the labels instead of the values, typing an input should match the label property.

**Actual behavior:**

The Typeahead shows the labels instead of the values, but typing an input matches the "value" property.

agubler:
**Type:** bug

**Description:**

All resource based list widgets should filter/search by the label not the "hidden" value of the item. Updates examples to be more realistic, i.e. have different values for `value` and `label` so incorrect behaviour is not masked.

Resolves #1562

### DIFF
--- a/src/chip-typeahead/index.tsx
+++ b/src/chip-typeahead/index.tsx
@@ -268,7 +268,7 @@ export const ChipTypeahead = factory(function ChipTypeahead({
 						return (
 							<ListItem {...props} selected={selected}>
 								<div classes={[themeCss.item, selected ? themeCss.selected : null]}>
-									{item.label || item.value}
+									{item.label}
 								</div>
 							</ListItem>
 						);
--- a/src/chip-typeahead/index.tsx
+++ b/src/chip-typeahead/index.tsx
@@ -153,7 +153,7 @@ export const ChipTypeahead = factory(function ChipTypeahead({
 
 								options.splice(index, 1);
 								icache.set('value', options.map((option) => option.value));
-
+								icache.set('options', options);
 								onValue && onValue(options);
 
 								focus.focus();
--- a/src/chip-typeahead/index.tsx
+++ b/src/chip-typeahead/index.tsx
@@ -162,8 +162,10 @@ export const ChipTypeahead = factory(function ChipTypeahead({
 			>
 				{{
 					label: selected
-						? selected(value, option && option.label)
-						: (option && option.label) || value
+						? selected(value, option ? option.label : value)
+						: option
+						? option.label
+						: value
 				}}
 			</Chip>
 		);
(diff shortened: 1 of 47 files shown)


src/chip-typeahead/index.tsx
```diff
@@ -52,11 +52,12 @@ export interface ChipTypeaheadChildren {
 		props: ListItemProperties & MenuItemProperties
 	) => RenderResult;
 	/** Custom renderer for selected items */
-	selected?: (value: string, label?: string) => RenderResult;
+	selected?: (value: string, label: string) => RenderResult;
```
agubler, line 55: Should we change this to `selected?: (options: ListOption) => RenderResult`

tomdye: I think `ListOption` contains `divider` / `disabled` so probably not?

tomdye (approved):

PR #1566, opened 2020-11-18
Update widgets to latest framework v7 patch release

agubler:
**Type:** bug / feature

**Description:**

Changes to ensure that assertion templates can target wrapped nodes within functional children mean that we need to update the Tree tests to actually test the children (which is correct regardless).

Also fixes the types for a harness abstraction.

### DIFF
--- a/src/tree/index.tsx
+++ b/src/tree/index.tsx
@@ -53,8 +53,7 @@ const factory = create({ theme, icache, diffProperty, focus, resource })
 	.properties<TreeProperties>()
 	.children<TreeChildren | undefined>();
 
-export default factory(function({
-	id,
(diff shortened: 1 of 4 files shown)


src/tree/index.tsx
```diff
@@ -53,8 +53,7 @@ const factory = create({ theme, icache, diffProperty, focus, resource })
 	.properties<TreeProperties>()
 	.children<TreeChildren | undefined>();
 
-export default factory(function({
-	id,
+export default factory(function Tree({
```
agubler, line 56: Add name to the function as per best practice

tomdye (approved):

PR #1596, opened 2020-12-01
Add z-index to SpeedDial

### LINKED ISSUE #1591: The z-index of the Speed Dial is not high enough
**Bug** 
The SpeedDial appears underneath a Card when used together.

speed dial z-index

agubler:
**Type:** bug

**Description:**

The speed dial should have a high z-index to ensure that it exists above standard application elements.

Resolves #1591


src/speed-dial/styles/speed-dial.m.css
```diff
@@ -0,0 +1,3 @@
+.root {
+	z-index: 301;
```
tomdye, line 2: popup's use the 400 range, do you think this should be the same?

tomdye (approved):

PR #1597, opened 2020-12-01
Support capability to indicate widgets should inherit css vars for the variant from a parent

### LINKED ISSUE #1586: Is the material theme for a snackbar correct?
**Bug** 
The current material theme for a snackbar looks like this:
snackbar

Are we sure this is correct? the contrast on the text is terrible?

agubler:
**Type:** feature

**Description:**

Proof of concept that introduces a property from the widget `theme` middleware to prevent applying the `theme.variant` class to the root meaning that widgets will instead inherit the variable properties from a parent. This allows variables to be partially overridden falling back to the standard set that are applied either at the root of the composite widget or the root of the application.

Unfortunately there was a clash of the property name with the `Avatar` widget that also used `variant` for a different purpose, this property has been renamed `type` and therefore makes this a breaking changes.

Also introduces a new widget `ActionButton` that sets the `variant` property to inherit that should be used for actions throughout the application, allowing parent widgets such as the `SnackBar` to control certain variables as required.

Resolves #1586 (when using the new `ActionButton`)

### DIFF
--- a/src/theme/material/variants/default.m.css
+++ b/src/theme/material/variants/default.m.css
@@ -3,7 +3,7 @@
 
 	/* variables taken from @material/theme/dist/mdc.theme.css */
 	--mdc-theme-primary: #6200ee;
-	--mdc-theme-dark-primary: #bb86fc;
+	--mdc-theme-primary-on-dark: #bb86fc;
 	--mdc-theme-secondary: #018786;
 	--mdc-theme-background: #fff;
 	--mdc-theme-surface: #fff;
--- a/src/theme/material/variants/default.m.css
+++ b/src/theme/material/variants/default.m.css
@@ -3,6 +3,7 @@
 
 	/* variables taken from @material/theme/dist/mdc.theme.css */
 	--mdc-theme-primary: #6200ee;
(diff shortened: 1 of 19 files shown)


matt-gadd (approved):

src/theme/material/variants/default.m.css
```diff
@@ -3,6 +3,7 @@
 
 	/* variables taken from @material/theme/dist/mdc.theme.css */
 	--mdc-theme-primary: #6200ee;
+	--mdc-theme-dark-primary: #bb86fc;
```
tomdye, line 6: could you name this `--mdc-theme-primary-on-dark` perhaps?

agubler: @tomdye why? Is that the convention?

tomdye: edited above

agubler: @tomdye thanks for updated the comment.

agubler: @tomdye updated.

tomdye (approved):

PR #1599, opened 2020-12-01
Make the form api typing more ergonomic

### LINKED ISSUE #1582: Better typing for the Form child renderer
**Enhancement** 

At the moment for a child renderer the user has to import the form middleware (which is odd in itself as not being used as a middleware), to pass a generic to it for the function to be typed. This is cumbersome and I doubt many users would discover they need to do this. Example below:

```tsx
<Form>{
({ field }: FormMiddleware<MyFieldsInterface>) => {
    const firstName = field('firstName');
    const lastName = field('lastName');
}
}</Form>
```

it would be nicer if we could change the api so importing a type was unnecessary, and perhaps even mandating the generic passed by defaulting it to never? Example below:

```tsx
<Form>{
(fields) => {
    const { firstName, lastName } = fields<MyFieldsInterface>();
}
}</Form>
```

agubler:
**Type:** feature

**Description:**

Overload the form middleware children function API to be a function factory that returns the API - this means that consumers do not need to import the `FormMiddleware` interface to get the typings, instead use the factory function and pass the field interface.

```ts
const { field } = form<Fields>();
```

Resolves #1582

### DIFF
--- a/src/examples/src/widgets/form/ActionForm.tsx
+++ b/src/examples/src/widgets/form/ActionForm.tsx
@@ -1,7 +1,6 @@
 import { createICacheMiddleware } from '@dojo/framework/core/middleware/icache';
 import { create, tsx } from '@dojo/framework/core/vdom';
 import Form, { FormField, FormGroup } from '@dojo/widgets/form';
-import { FormMiddleware } from '@dojo/widgets/form/middleware';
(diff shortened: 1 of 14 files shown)


tomdye (approved):

tomdye:
Nice 👍

PR #1601, opened 2020-12-01
Add disabled and readonly to date input and make name optional

### LINKED ISSUE #1581: `name` property is required on `DateInput` but not on any other input widgets
**Bug**

The `name` property is required on the `DateInput` but not on any of the other input widgets which is strange. It should be consistent across the type of widgets.

### LINKED ISSUE #1583: Add disabled property for the DateInput widget
**Enhancement**

Right now the `DateInput` widget has no `disabled` or `readOnly` property. It could be helpful for developers using this widget to have an option to disable this widget as needed.

agubler:
**Type:** bug

**Description:**

Adds `disabled` and `readOnly` props to the `DateInput` and makes `name` optional.

Resolves #1583
Resolves #1581


agubler:
@tomdye I've added `disabled` and `readOnly` props to the date input that get passed to the underlying `TextInput`, let me know if there should be a class/style applied at all other than passing those props through to the `TextInput`

tomdye (approved):
Looks good to me, text input should style correctly off of the properties, only thing you / we might want to do is to change the date icon color too perhaps?

tomdye (approved):

PR #1605, opened 2020-12-04
Consistently pass theme related props to widgets

agubler:
**Type:** bug

**Description:**

Widgets that compose other widgets should pass all theme related properties down, such as `theme`, `classes` and `variant`. Also we should consistently use the `theme` middleware from widgets across the codebase and not mix and match between that and the core middleware from framework.


tomdye (approved):
Everything seems to be in order to me

PR #1642, opened 2021-01-17
Convert widgets to use dojo 8 resources

agubler:
**Type:** feature

**Description:**

Converts all resource based widgets to use the latest version of the resource from dojo/framework


tomdye (approved):
Everything appears to be working as expected

PR #1658, opened 2021-01-29
Add required property to DateInput widget

### LINKED ISSUE #1644: Add required property to DateInput
**Bug** 

The `DateInput` widget doesn't have the `required` property. This would be useful when using this widget in a form where the date is required

Package Version: 7.0.6

samends:
**Type:** Feature

**Description:**

Adds required property to `DateInput` widget

Resolves #1644


src/examples/src/widgets/date-input/Required.tsx
```diff
@@ -0,0 +1,24 @@
+import { create, tsx } from '@dojo/framework/core/vdom';
+import DateInput from '@dojo/widgets/date-input';
+import icache from '@dojo/framework/core/middleware/icache';
+import Example from '../../Example';
+
+const factory = create({ icache });
+
+export default factory(function Controlled({ middleware: { icache } }) {
```
tomdye, line 8: Can you rename this to `Required` please

tomdye (changes requested):
@samends it doesn't look to me that the error state is being passed to the label / input border etc here when it is invalid due to this new  `requiredField`. Please see attached picture of current appearance vs expected.

Screenshot 2021-02-03 at 11 26 04
Screenshot 2021-02-03 at 11 28 22

src/examples/src/widgets/date-input/Required.tsx
samends: yup!

samends:
Hey @tomdye check again, I updated the `DateInput` widget to use the `valid` property for the error messages instead of the `helperText` property

tomdye:
@samends looks good when the widget is set to required, however - should the date input be green to indicate it's valid in it's standard non-validated use case? Looks like widgets such as the numberInput do not do this.

samends:
hmm yea your right, looks like that's the standard `TextInput` behavior, let me check in on this

samends:
Hey @tomdye I updated the `DateInput` widget so it doesn't show the green outline as default

tomdye:
@samends i think the problem now is that when the widget _is_ valid after user input it is not displaying the green outline either. My comment was only that it shouldn't be there on first load. ie. before the widget becomes dirty and is interacted with.

tomdye:
@samends it looks like when it's `required` and there is no value it's not showing as invalid and when it's not required, ie a standard date input and there is a value, it's showing as valid when it's clicked or when it has a value. Can you clarify what you expect the behaviour to be in the latter case please - do we expect a non-validated widget to show as visibly valid? 
Screenshot 2021-03-01 at 09 22 08
Screenshot 2021-03-01 at 09 22 19

samends:
Hey @tomdye I've updated this PR so the `DateInput` widget is in line with existing input widget behavior. There was two things that were found I wanted to clarify:

- It seems the `TextInput` widget doesn't show the success outline when the input is required, it only shows the success outline when the `pattern` property is used, did we also want to not show the success outline for the `DateInput` widget?

- When the `TextInput` widget is required, when the input focuses or is blurred without any input it doesn't show an invalid error, right now the `DateInput` also has this behavior. Did we want to show an invalid message when a required input is focused/blurred without any input?

For both of these issues I could do a follow up PR to make these behaviors consistent, let me know!

tomdye:
The behaviour seems to match other inputs now as you suggested @samends . I don't think there's any need for a follow up PR just yet, better we work out the correct consistent approach across the board first as it will need to be implemented for all our widgets consistently.

tomdye (approved):
Thanks for seeing this through @samends

PR #1663, opened 2021-02-09
Support actions in the header section of the header card

agubler:
**Type:** feature

**Description:**

Adds support for aligning "actions" on the right of a header in the card. This has only been added to the `header-card` which is the card to use if you want support with the header layout.


agubler:
@tomdye some questions that have arisen when working on on the header actions?

* Why is there a minimum width for buttons in the dojo theme?
* Why is there a minimum width for list/menu items in material? It looks like it comes from the material theme
* If there is a minimum width on list items, for popup menu's should we be able to support aligning the popup to the right or of the trigger instead of the left?

PR #1667, opened 2021-02-20
add text-input autocomplete hook

### LINKED ISSUE #1610: Autocomplete messes with floating label styles like those in Material
**Bug** 

Screenshot 2020-12-07 at 13 34 52

Apparently this can be fixed with a transition to trigger js via the ` :-webkit-autofill` pseudo (e.g https://codedaily.io/tutorials/69/Animated-Input-Label-with-Chrome-Autofill-Detection-in-React).

aciccarello:
**Type:** bug

**Description:**
Use keyframes to trigger an event when 

Resolves #1610

### DIFF
--- a/src/text-input/index.tsx
+++ b/src/text-input/index.tsx
@@ -106,6 +106,7 @@ interface TextInputICache {
 	dirty: boolean;
 	value?: string;
 	initialValue?: string;
+	autofilled?: boolean;
 }
 
 const factory = create({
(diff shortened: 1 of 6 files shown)


tomdye (approved):
Looks good to me, do you think it should also call `onValue` when the input is autocompleted?

aciccarello:
Good question @tomdye. I think that would make sense to add.

aciccarello:
@tomdye I remembered that the value isn't available at this point (for security reasons) so there wouldn't be any value to pass in the callback

tomdye:
@aciccarello fair enough re. value not being available. I recall a similar issue with autocomplete on a sign in form needing to enable a sign in button. This PR is already approved.

PR #1668, opened 2021-03-03
SpeedDial position property

### LINKED ISSUE #1571: SpeedDial does not have a positioning property
**Bug** 
The first thing someone wants to do with the speed dial is position it (usually in the bottom right hand corner). There is currently no option for this

aciccarello:
**Type:** feature

**Description:**
This adds a property to quickly fixed-position the speed dial either in the bottom right or bottom center. In those cases it will default to a direction of "up".

The floating-action-button previously had no background, but this is problematic when displaying over other content so I removed that style.

### Open questions:
* [ ] Should the direction attribute be allowed at all with positioning?
* [ ] Are these the positions that we want?
* [ ] Does the button style change need to be called out as a breaking change?

Resolves #1571

## Screenshots
localhost_9999_(iPad) (3)
localhost_9999_(iPad) (2)
localhost_9999_(iPad) (6)
localhost_9999_(iPad) (4)
localhost_9999_(iPad) (5)


src/examples/src/config.tsx
```diff
@@ -1659,6 +1660,11 @@ export const config = {
 					module: DirectionSpeedDial,
 					title: 'With a direction'
 				},
+				{
+					filename: 'Position',
```
matt-gadd, line 1664: can you make this example use the iframe and sizing so it correctly displays in the example area?

```ts
  sandbox: true,
  size: 'large'
````

matt-gadd (changes requested):
I think the positional change should apply to the floating action button too (given it's the singular version of a speed dial).

tomdye:
@aciccarello do you think this should have a bottom-left option too? Also how about putting the example in a large sandbox so that the action button does not bleed out of the example area?

src/examples/src/config.tsx
tomdye: same catch as me - but more helpful 👍

aciccarello:
@matt-gadd I can add this to the FAB as a followup PR

matt-gadd (approved):

aciccarello:
I've updated styles to use variables for offset position and `translateX`/`translateY` to center the position following some learning from the FAB PR (#1670). Also improved the example so that unsupported directions can't be used with the position.

matt-gadd (approved):

PR #1676, opened 2021-03-06
Calculate the size of rows correctly

agubler:
**Type:** bug

**Description:**

Enables calculating the size of rows correctly for material, this removes the fixed height of the row items and changes to a min height which enables custom list items. Also it splits the padding and the other styles from the list root so the list root can be applied when doing the offscreen render as it contains styles that affect the size (from the material styles).

In future the list should not affect the list item styles at all.

Resolves #???


tomdye (approved):

src/examples/src/config.tsx
```diff
@@ -302,8 +302,8 @@ export const config = {
 	name: '@dojo/widgets',
 	home: 'src/examples/README.md',
 	themes: [
-		{ label: 'dojo', theme: dojoTheme },
 		{ label: 'material', theme: materialTheme },
+		{ label: 'dojo', theme: dojoTheme },
```
aciccarello, line 306: Was this reorder intentionally committed? I dug into the commits because I was afraid I had done it 😅

agubler: Weird I didn't get a notification for this comment! I did change it on purpose, an executive decision because i was always looking at material over dojo in the majority of cases

PR #1677, opened 2021-03-08
Offer option to fill the available height with items when using the list

agubler:
**Type:** feature

**Description:**

If would be nice if the list could offer a mode that would fill the available space (height) with as many items that fit, instead of having to provide a number of items to render. The is issue is, that the list itself does the list item height calculation so as a user we'd have to duplicate that to calculated how many items could be visible in the space.

The other option would be to pass a "height" property that gives the list the height and then uses that to determine how many items should be in view.


tomdye (approved):
great stuff

PR #1681, opened 2021-03-10
fix icon button positioning

### LINKED ISSUE #1594: Icon Button is not positioned correctly
**Bug** 
Especially bad for material theme. See below: 

Screenshot 2020-11-30 at 13 11 31

https://widgets.dojo.io/#widget/icon/example/iconbutton

aciccarello:
**Type:** bug

**Description:**
Vertically aligns the label of a button so that icons and text are aligned and adds a 8px gap between text and icons. Where the button variants overrode the label styles (mostly in Dojo theme), it was necessary to compose them back in. This uses flex gap which isn't currently supported by Safari (though it is in Technical Preview). I figured the style wasn't strictly necessary so a slight degradation seemed acceptable.

[Can I Use flex gap usage](https://caniuse.com/flexbox-gap)

The example was updated to include more button types and to use small sized icons which fit better and should be recommended when used with the material theme.

Resolves #1594

Icon Button example page


tomdye:
@aciccarello it looks like material has a specific `icon-button` with different classes used to align the label and icon appropriately - https://material.io/develop/web/components/buttons/icon-buttons
It also seems that for the dojo theme, adding `display: flex; justify-content: center; align-items: center;` aligns and centers the icon / label appropriately.

Do you think it might make sense to have a separate `IconButton` ? That way we could make the `label` optional and thus allow for buttons containing only an icon also. With this approach, in the material case, the appropriate `icon-button` css classes could be used.

aciccarello:
@tomdye I'm confused how material integrates that with the different variants. On the design spec [they show](https://material.io/components/buttons#anatomy) it being used with different variants. How do they have a single `icon-button`?

material design diagram with different button variants with icons

Is an icon-only icon a priority? If so I can work on that styling (which would be mostly reducing the button padding).

aciccarello:
Ah, the icon-button is only for icon-only buttons. It's taller (48px instead of 36px) and square. For a button with a label we should stick with what we have. If we want an icon-only button I'd recommend adding an icon-button widget separately.

> **Note:** For buttons with both icons and text, use the mdc-button component. For more information, see the mdc-buttondocs.

- [Material Web Icon Button Description](https://material.io/develop/web/components/buttons/icon-buttons)

I'm working on a margin solution which doesn't use flex gap and also supports the adjustments for the margin between the button border and the icon by passing in the icon separately.

tomdye:
@aciccarello If we changed the button's children api to take `icon` and `label`, both of which were `RenderResults` then we could do the appropriate wrapping and also handle the case where only an `icon` was passed and thus handle the sizing etc correctly. Would mean a breaking change though unless we allowed either / or. ie.

```
<Button>Button Label</Button>
<Button>{{ label: 'button label' }}</Button>
<Button>{{ icon: <Icon type='add'/> }}</Button>
<Button>{{ label: 'button label', icon: <Icon type='add'/>}}</Button>
```

Might be worth having a discussion about in the dojo room. I think the breaking change is the better approach personally as we're moving onto dojo 8 and it's a simple enough change, but the fallback to the previous api may be acceptable.

tomdye (approved):
Looks good to me @aciccarello , the examples all work well with the correct spacing etc

PR #1683, opened 2021-03-10
fix material card header margin

### LINKED ISSUE #1588: Margin looks incorrect for header on Material Card
**Bug** 
The margin looks incorrect for header on Material Card?

margin card

aciccarello:
**Type:** bug

**Description:**
Allows material card title margin styles if it is the first element. To do so we need to check that there isn't a media element or header before the title.

Resolves #1588

material card title example
material card media and title example


tomdye (approved):
Thought I’d already approved this.

PR #1686, opened 2021-03-16
feat: SpeedDial action label support

### LINKED ISSUE #1572: SpeedDial does not have the capability to show labels
**Enhancement** 

It would be nice to be able to have text labels along with an icon in the Speed Dial where an icon may not be descriptive enough. See: https://material-ui.com/components/speed-dial/#persistent-action-tooltips

bitpshr:
**Type:** feature

**Description:**

This pull request adds label support for `SpeedDial` actions.

preview

Resolves #1572


src/examples/src/config.tsx
```diff
@@ -1665,31 +1666,46 @@ export const config = {
 			overview: {
 				example: {
 					filename: 'Basic',
-					module: BasicSpeedDial
+					module: BasicSpeedDial,
+					sandbox: true,
+					size: 'medium'
```
bitpshr, line 1671: Applicable `SpeedDial` examples were sandboxed so that fixed positions work correctly. They were previously broken and / or used hardcoded widths and heights.

src/examples/src/widgets/speed-dial/Basic.tsx
```diff
@@ -11,13 +11,6 @@ export default factory(function Basic({ middleware: { icache } }) {
 	return (
 		<Example>
 			<SpeedDial>
-				<Action
-					onClick={() => {
-						icache.set('action', 'Mailing');
-					}}
-				>
-					<Icon type="mailIcon" />
-				</Action>
```
bitpshr, line 20: Since we're using a sandboxed example with a default "medium" size, one action was removed from each example so things fit without scrolling and without explicit heights.

src/examples/src/widgets/speed-dial/Position.tsx
```diff
@@ -11,18 +11,11 @@ type Directions = 'right' | 'left' | 'up' | 'down';
 export default factory(function Direction({ middleware: { icache } }) {
 	const position: SpeedDialPositions = icache.getOrSet('position', 'bottom-right');
 	const supportedDirections = getSupportedDirections(position);
-	let direction: Directions | undefined = icache.get('direction');
+	let direction: Directions | undefined = icache.getOrSet('direction', 'left');
```
bitpshr, line 14: Since the most common use case for the `SpeedDial` is at the bottom right of the screen, the default `position` and `direction` were updated to reflect this (along with applicable examples, as seen here.)

src/speed-dial/index.tsx
```diff
@@ -7,17 +7,37 @@ import * as fabCss from '../theme/default/floating-action-button.m.css';
...
+
 export interface ActionProperties {
 	/* On click callback for the action */
 	onClick(): void;
 	/* Title text for the action */
 	title?: string;
+	/* Static label for the action */
+	label?: string;
+	/* Orientation for the label */
+	labelOrientation?: LabelOrientation;
```
bitpshr, line 26: Note that `label` was intentionally chosen as an independent prop to `title`, the idea being that `title` is meant for a11y-specific titling and may be different than visual helper text (e.g. `label` could say "Click here!" while the `title` is "Click here for deals" for screen readers. If we want to instead consolidate these, that's easy to do as well.

tomdye (approved):

tomdye:
My only comment is that it might be more appropriate to use an `label` node for the label than a `div`. At the very least, it likely needs some aria attributes to link it to the button etc.

tomdye (changes requested):
Took a second glance after looking back through dojo chats at the button changes, I think it would be best if the children were consistent.

src/speed-dial/index.tsx
```diff
@@ -34,10 +66,36 @@ export const Action = actionFactory(({ properties, children, middleware: { theme
...
+				break;
+			case LabelOrientation.left:
+			default:
+				orientationClass = themedCss.labelLeft;
+				break;
+		}
+
+		return (
+			<div classes={[themedCss.labelContainer, orientationClass]}>
+				<label classes={themedCss.label}>{label}</label>
```
tomdye, line 93: I think this might need to use an aria attribute or `for` to link to the fab's `id`

src/speed-dial/index.tsx
```diff
@@ -33,11 +66,46 @@ export const Action = actionFactory(({ properties, children, middleware: { theme
 			}}
 			classes={classes}
 			variant={variant}
+			aria={
+				label
+					? {
+							'aria-describedby': labelId
```
tomdye, line 72: the `aria` property shouldn't be prefixed by `aria` as this is done automatically. You can see the result in the browser `aria-aria-describedby`.
Screenshot 2021-03-30 at 14 49 51.
Could you add a specific `Action` test to capture this please? Currently it's untested.

tomdye (changes requested):

tomdye:
It looks like most of the innards of `Action` are untested atm, do you think it would be worth adding some tests around those?

tomdye:
@bitpshr I think it's still worth adding a test here for the action render otherwise the label functionality is entirely untested, hence why the `aria-aria` bug got through.

PR #1690, opened 2021-03-19
add leading and trailing to listitem widget

### LINKED ISSUE #1570: The List has no ListItem widgets out of the box
**Enhancement** 

It would be nice not to have to write a load of list items manually in my app. Here is the link to materials list items: https://v3.material-ui.com/demos/lists/#interactive we should implement as many as possible.

aciccarello:
**Type:** feature

**Description:**
Changes ListItem widget to take leading, trailing, and primary child properites. If a RenderResult is passed, it is rendered as the primary content.

Resolves #1570 

list item renderer example page with dojo theme
list item renderer example page with material theme

### DIFF
--- a/src/examples/src/widgets/list/ListItemRenderer.tsx
+++ b/src/examples/src/widgets/list/ListItemRenderer.tsx
@@ -8,15 +8,12 @@ import {
 	createResourceTemplate,
 	createResourceMiddleware
 } from '@dojo/framework/core/middleware/resources';
-import Button from '@dojo/widgets/button';
-import * as exampleCss from './ListItemRenderer.m.css';
 
 const resource = createResourceMiddleware();
 const factory = create({ icache, resource });
 const template = createResourceTemplate<typeof states[0]>('value');
 
 export default factory(function ListItemRenderer({ id, middleware: { icache, resource } }) {
-	const listItemClasses = { '@dojo/widgets/list-item': { height: [exampleCss.rowHeight] } };
 	return (
 		<Example>
 			<List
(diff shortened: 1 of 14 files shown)


src/examples/src/widgets/list/ListItemRenderer.tsx
```diff
@@ -0,0 +1,105 @@
...
+} from '@dojo/framework/core/middleware/resources';
+import Button from '@dojo/widgets/button';
+import * as exampleCss from './ListItemRenderer.m.css';
+
+const resource = createResourceMiddleware();
+const factory = create({ icache, resource });
+const template = createResourceTemplate<typeof states[0]>('value');
+
+export default factory(function ListItemRenderer({ id, middleware: { icache, resource } }) {
+	const listItemClasses = { '@dojo/widgets/list-item': { height: [exampleCss.rowHeight] } };
```
matt-gadd, line 19: can you remove these style overrides please 😬, if it doesn't look right when removed then i'd say the current default behaviour isn't good enough.

src/examples/src/widgets/list/ListItemRenderer.tsx
```diff
@@ -0,0 +1,105 @@
...
+										primary: label,
+										trailing: (
+											<Button
+												onClick={() =>
+													alert(
+														`You clicked ${value}. This doesn't select it.`
+													)
+												}
+												classes={{
+													'@dojo/widgets/button': {
```
matt-gadd, line 63: remove overrides.

src/examples/src/widgets/list/ListItemRenderer.tsx
```diff
@@ -0,0 +1,105 @@
...
+					template: template({ id, data: states }),
+					transform: { value: 'value', label: 'value' }
+				})}
+				onValue={(value) => {
+					icache.set('value', value);
+				}}
+				itemsInView={8}
+			>
+				{({ value, label }, props) => {
+					switch (value) {
```
matt-gadd, line 33: perhaps this would be better now with just 3 lists that pass a single item? I don't really get the relevance of having the entire state list and selecting portions of it?

src/examples/src/widgets/list/ListItemRenderer.tsx
```diff
@@ -0,0 +1,105 @@
...
+							return (
+								<ListItem {...props} classes={listItemClasses}>
+									{{
+										leading: <Icon type="infoIcon" />,
+										primary: label,
+										trailing: (
+											<Button
+												onClick={() =>
+													alert(
+														`You clicked ${value}. This doesn't select it.`
```
matt-gadd, line 59: I don't really get the relevance of this? in fact my first thought was after i saw the alert was how do I select the item 😄? Is this example meant to be an exhaustive example of the List Item behaviour or?

matt-gadd (changes requested):

src/examples/src/widgets/list/ListItemRenderer.tsx
aciccarello: This was showing that click events in the trailing child don't bubble up to select the item. It is probably not needed in the example so I can remove it.

tomdye (approved):
looks like you addressed @matt-gadd review comments

Issue #1695, opened 2021-03-23
TextInput `Addon` approach re-work


tomdye:
The existing `Addon` usage for the `Text-Input` does not work properly with custom elements which is especially problematic when using `material` as it requires a class provided by `addon` for it's icon placement.
Further to the above, the `Addon` approach is not very user friendly and can be confusing.
I propose we remove it and instead, wrap the passed `RenderResult` in the addon class.

PR #1697, opened 2021-03-23
Remove text-input `Addon` widget

tomdye:
This PR removes the text-input `Addon` widget and makes it more user friendly.
You can now simply pass the leading / trailing dom node directly to the `TextInput`.

Resolves: #1695


src/theme/dojo/text-input.m.css
```diff
@@ -93,16 +93,17 @@
...
 
-.addonRoot {
+.leading,
+.trailing {
 	color: var(--color-text-faded);
 	flex: 0 0 auto;
 	font-size: inherit;
 	line-height: var(--line-height-base);
-	padding: 0 var(--grid-base);
+	/* padding: 0 var(--grid-base); */
```
samends, line 102: Did you mean to change this value here?

samends:
Looks good Tom! What was the original thought behind the `Addon` widget?

samends (approved):

src/theme/dojo/text-input.m.css
tomdye: need to remove that

src/text-input/tests/unit/TextInput.spec.tsx
```diff
@@ -606,12 +610,12 @@ registerSuite('TextInput', {
 					css.noLabel
 				])
 				.prepend('@inputWrapper', () => [leading]);
-			const h = harness(() => <TextInput>{{ leading }}</TextInput>);
+			const h = harness(() => <TextInput>{{ leading: <span>A</span> }}</TextInput>);
```
aciccarello, line 613: You could probably get away with just using a text node instead of a full span.
```suggestion
			const h = harness(() => <TextInput>{{ leading: 'A' }}</TextInput>);
```

src/examples/src/widgets/number-input/LeadingTrailing.tsx
```diff
@@ -11,8 +10,8 @@ export default factory(function Basic() {
 			<NumberInput>
 				{{
 					label: 'Number Input',
-					leading: <Addon>$</Addon>,
-					trailing: <Addon filled>MM</Addon>
+					leading: <span>$</span>,
```
aciccarello, line 13: I think we should show users that they can pass text here. For most use cases that should be sufficient.
```suggestion
					leading: '$',
```

src/text-input/tests/unit/TextInput.spec.tsx
tomdye: yup yup, you are correct thought wanted to show that you can add domnodes as well as text

src/examples/src/widgets/number-input/LeadingTrailing.tsx
tomdye: agreed, though I have shown a mix of uses.

tomdye:
> Looks good Tom! What was the original thought behind the `Addon` widget?

I believe it was to make the padding optional but this can be done better via theming etc now. The previous approach was awkward for custom elements.

tomdye:
@aciccarello comments addressed

aciccarello (approved):

PR #1699, opened 2021-03-23
Trigger onValidate when inital value is set on Select widget

### LINKED ISSUE #1653: Select isn't validating initial value
`Select` is reporting false validity when using an initial value.

https://codesandbox.io/s/distracted-meadow-2tdj1?file=/src/main.tsx

Click the select box to open it and then closing it will cause validity to report true.

samends:
**Type:** bug

**Description:**
`Select` widget was not validating when there was an initial value set. Pending unit test

Resolves #1653

### DIFF
--- a/src/select/index.tsx
+++ b/src/select/index.tsx
@@ -139,7 +139,7 @@ export const Select = factory(function Select({
 		data
 	} = get(options(), { read, meta: true });
 
-	if (required && dirty) {
+	if ((required && dirty) || value !== undefined) {
 		const isValid = value !== undefined;
 		if (isValid !== valid) {
 			icache.set('valid', isValid);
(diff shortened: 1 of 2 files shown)


src/select/index.tsx
```diff
@@ -139,7 +139,7 @@ export const Select = factory(function Select({
 		data
 	} = get(options(), { read, meta: true });
 
-	if (required && dirty) {
+	if ((required && dirty) || value !== undefined) {
```
aciccarello, line 142: The extra parenthesis shouldn't be necessary
```suggestion
	if (required && dirty || value !== undefined) {
```

aciccarello (approved):

tomdye (changes requested):
I'm not sure this is right as its validating selects even when they are not required and do not require validation.

samends:
@tomdye this was added so that if there's an initial value set, the `Select` widget will validate it

tomdye:
@samends I believe that any passed `onValidate` function should be called but if `required` is not set to true it should not validate internally.

tomdye (approved):

PR #1700, opened 2021-03-23
Support secondary text in ListItem widget

### LINKED ISSUE #1688: Support secondary text in ListItem widget
**Enhancement** 

The ListItem widget should be able to accept secondary text. In the material design theme, this would be shown below the primary text with a lighter shade font.

**Code**
```
<ListItem>
{{
  primary: 'Primary text',
  secondary: 'Secondary text'
}}
</ListItem>
```

**Expected behavior:**

[Secondary text list example from material design spec](https://material.io/components/lists)

This is split from #1570

aciccarello:
**Type:** bug / feature

**Description:**
This adds secondary text to the ListItem widget. Because secondar text often requires more height, I've replaced the single height css class with a range of sizes. This also allows different size leading/trailing content. It's a breaking change for anyone overriding that CSS class but should allow greater flexibility.

Resolves #1688

secondary list item material theme
secondary list item dojo theme
primary list item material theme
primary list item dojo theme


src/list/index.tsx
```diff
@@ -153,9 +158,25 @@ export const ListItem = listItemFactory(function ListItem({
...
-		? { primary: [firstChild, ...otherChildren] }
-		: firstChild;
+	const {
+		leading = undefined,
+		primary,
+		secondary = undefined,
+		trailing = undefined
+	} = isRenderResult(firstChild) ? { primary: [firstChild, ...otherChildren] } : firstChild;
+	const size = height || (secondary ? 'medium' : 'small');
+	const sizeStyle =
```
tomdye, line 168: Im not sure we should be setting a size at all, the list items should adapt to the size of their content and render appropriately. In any case we certainly shouldn't be using a massive ternary such as this.

tomdye (changes requested):
@aciccarello It seems to me that this would be better implemented as different List Item types. ie. Avatar Item, Two Line Item etc. Otherwise we're just going to end up with a large number of different properties when we could have had easy to use / reason with widgets.
It would also be good to have an example of each of these item types.

PR #1703, opened 2021-03-24
fix: disabled input backgrounds

bitpshr:
**Type:** bug

**Description:**

This pull request consolidates disabled backgrounds between `Select` and other inputs. A few different underlying issues caused the difference in background color: `Select` used an unnecessary explicit background color, and `Input` incorrectly applied background styles.

Resolves #1703

### DIFF
--- a/src/theme/material/select.m.css
+++ b/src/theme/material/select.m.css
@@ -28,7 +28,6 @@
 
 .disabled {
 	composes: mdc-select--disabled from '@material/select/dist/mdc.select.css';
-	background-color: var(--mdc-disabled-color);
 }
 
 .required {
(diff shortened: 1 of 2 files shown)


src/theme/material/select.m.css
```diff
@@ -28,7 +28,6 @@
 
 .disabled {
 	composes: mdc-select--disabled from '@material/select/dist/mdc.select.css';
-	background-color: var(--mdc-disabled-color);
```
bitpshr, line 31: I wasn't sure why we ever had this to begin with, as `mdc-select--disabled` provides the correct disabled background already, which is the same as `Input`.

src/theme/material/text-input.m.css
```diff
@@ -18,11 +18,11 @@
 	composes: mdc-line-ripple from '@material/textfield/dist/mdc.textfield.css';
 }
 
-.root .wrapper {
+.root .wrapper:not(.disabled) {
```
bitpshr, line 21: This selector was clobbering the correct disabled background.

bitpshr:
Closing in favor of #1702.

PR #1704, opened 2021-03-24
fix: outline button dojo theme

### LINKED ISSUE #1682: Dojo theme button colors illegible
**Bug** 

Package Version: 

**Code**

**Expected behavior:**
Button background and text are different colors so text is legible.

**Actual behavior:**
### Dojo Outlined button in toggle selected state
Dojo outlined button filled blue with no text visible

### Dojo dark theme buttons
Dojo dark theme buttons are all black with only outlined button text visible

bitpshr:
**Type:** bug

**Description:**

This pull request fixes focus styling and pressed state styling for the `OutlinedButton` using the Dojo theme.

Resolves #1682

### DIFF
--- a/src/theme/dojo/outlined-button.m.css
+++ b/src/theme/dojo/outlined-button.m.css
@@ -16,6 +16,10 @@
 	font-family: var(--font-family);
 }
 
+.root:focus {
+	outline: none;
+}
+
 .iconOnly {
 	min-width: 0;
 }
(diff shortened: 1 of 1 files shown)


tomdye:
Screenshot 2021-03-25 at 15 01 04
I'm not sure this is correct as it now has no accent at all compared to what is on master and no hover / click styling.

tomdye:
I just realised that I was looking at the v7 widgets example and not the latest, so apologies for my last comment.

tomdye (approved):

PR #1705, opened 2021-03-24
fix: rename CardProperties.onAction to onContentClick

### LINKED ISSUE #1648: Card widget onAction is confusing
Card's `onAction` callback is really just an `onClick` handler for the entire card. This should either be removed or renamed.

bitpshr:
**Type:** feature

**Description:**

This pull request renames the `Card` prop `onAction` to `onContentClick`, since it more-accurately describes what the prop actually does.

Resolves #1648

### DIFF
--- a/src/card/index.tsx
+++ b/src/card/index.tsx
@@ -4,7 +4,7 @@ import * as css from '../theme/default/card.m.css';
 import theme from '../middleware/theme';
 
 export interface CardProperties {
-	onContentClick?: () => void;
+	onClick?: () => void;
 	mediaSrc?: string;
 	mediaTitle?: string;
 	square?: boolean;
@@ -27,7 +27,7 @@ const factory = create({ theme })
 export const Card = factory(function Card({ children, properties, middleware: { theme } }) {
 	const themeCss = theme.classes(css);
 	const {
-		onContentClick,
+		onClick,
 		mediaSrc,
 		mediaTitle,
 		square,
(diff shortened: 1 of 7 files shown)


tomdye:
I didn't even know there was a click handler for card content. As there aren't any more click handlers for the `Card`, do you think `onClick` would be more appropriate name wise? I think that was the suggestion in the issue too.

bitpshr:
@tomdye `onClick` seems wrong since this does not include the card header, which I'd expect as a consumer. I think we should either call it `onContentClick`, or remove it altogether.

bitpshr:
This pull request has been updated in the following ways:

1. `onClick` is used instead of `onContentClick`, and this applies to all content but the action buttons as per the Material guidelines. See Action Area 1 in [this image](https://lh3.googleusercontent.com/t6XT_g7O4FhNRAK-IhjO4C3yTaSOKbROxr-lamFv6GGE0fh0UJGv36QZ3nQixaZ-OM9jkP0KZFN36ubXIDHjFxOYQuP0IyOW-YGNzzk=w1064-v0) ([reference](https://material.io/components/cards#actions).)
1. Styling was improved for `header` content so it has proper padding.

tomdye:
@bitpshr could you add an example using the click handler please and add the `pointer: cursor` css class we discussed in the dojo call when an `onClick` callback as been passed.

matt-gadd (approved):

tomdye (approved):

PR #1706, opened 2021-03-24
Adds circular loader

### LINKED ISSUE #1584: Update animations for LoadingIndicator widget
**Enhancement** 

It would be helpful for a developer to have a loading widget that has a property value that would change the type of spinner animation as needed. The update could involve adding the property value of `type` to the `LoadingIndicator` widget that could have the basic values `linear` and `circular` as described here: https://material.io/components/progress-indicators#types as well as some others. Some examples I found are:

- https://spin.js.org/
- https://loading.io/css/
- https://ant.design/components/spin/

samends:
**Type:** feature

**Description:**
Adds circular loader to `LoadingIndicator` widget

Resolves #1584


package.json
```diff
@@ -69,6 +69,7 @@
     "@material/card": "5.1.0",
     "@material/checkbox": "5.1.0",
     "@material/chips": "5.1.0",
+    "@material/circular-progress": "6.0.0",
```
samends, line 72: version `5.1.0` doesn't exist for this package https://www.npmjs.com/package/@material/circular-progress

src/theme/default/loading-indicator.m.css
```diff
@@ -21,3 +29,50 @@
...
+/*  An indeterminate SVG (there are three of these in total). */
+.circleGraphic {
+}
+
+/* A tiny little sliver of an SVG circle used to patch a gap between the circle-left and the circle-right. */
+.gapPatch {
+}
+
+/* Small size for circular loader */
+.small {
```
samends, line 63: sizing matches material ui https://github.com/material-components/material-components-web/tree/master/packages/mdc-circular-progress#sizing

src/examples/src/widgets/loading-indicator/Circular.tsx
```diff
@@ -0,0 +1,29 @@
...
+import Example from '../../Example';
+
+const factory = create({ icache });
+
+export default factory(function Circular({ middleware: { icache } }) {
+	const active = icache.getOrSet('active', true);
+	return (
+		<Example>
+			<virtual>
+				<LoadingIndicator type={'circular-small'} active={active} />
```
agubler, line 14:
```suggestion
				<LoadingIndicator type="circular-small" active={active} />
```

src/loading-indicator/index.tsx
```diff
@@ -6,29 +6,113 @@ import { create, tsx } from '@dojo/framework/core/vdom';
 export interface LoadingIndicatorProperties extends ThemedProperties {
 	/** If the element is actively loading. Defaults to true */
 	active?: boolean;
+	/* Sets loader type to linear or circular */
```
agubler, line 9:
```suggestion
	/* Sets loader type to linear or circular, defaults to `linear` */
```

src/loading-indicator/index.tsx
```diff
@@ -6,29 +6,113 @@ import { create, tsx } from '@dojo/framework/core/vdom';
...
 
 export const LoadingIndicator = factory(function LoadingIndicator({
 	properties,
 	middleware: { theme }
 }) {
 	const classes = theme.classes(css);
-	const { active = true } = properties();
+	const { active = true, type = 'linear' } = properties();
+
+	const [loadingType, size = 'small'] =
```
agubler, line 24: Can we not do this, it's better to be more explicit/verbose 😄 

```tsx
const isLinear = type === 'linear';

if (isLinear) {
    return (
        // the existing loading indicator
    )
}

let sizeClass: string | undefined;
switch (type) {
	case 'circular-small':
		sizeClass = themeCss.small;
		break;
	case 'circular-large':
		spacingClass = themeCss.large;
		break;
	case 'circular-medium':
	default:
		sizeClass = themeCss.medium;
		break;
}

return (
    // The circular indicator using the sizeClass
)
```

src/loading-indicator/index.tsx
```diff
@@ -6,29 +6,113 @@ import { create, tsx } from '@dojo/framework/core/vdom';
...
+						classes.linear
+					]}
+					role="progressbar"
+				>
+					<div classes={classes.buffer} />
+					{active ? (
+						<div classes={[classes.bar, classes.primary]}>
+							<span classes={classes.inner} />
+						</div>
+					) : null}
```
agubler, line 45:
```suggestion
					{active && (
						<div classes={[classes.bar, classes.primary]}>
							<span classes={classes.inner} />
						</div>
					)}
```

src/loading-indicator/index.tsx
```diff
@@ -6,29 +6,113 @@ import { create, tsx } from '@dojo/framework/core/vdom';
...
+											r="8.75"
+											stroke-dasharray="54.978"
+											stroke-dashoffset="27.489"
+											stroke-width="2.5"
+										/>
+									</svg>
+								</div>
+							</div>
+						</div>
+					) : null}
```
agubler, line 112:
```suggestion
					{active && (
						<div classes={classes.circularContainer}>
							<div classes={classes.spinnerLayer}>
								<div classes={[classes.circleClipper, classes.circleLeft]}>
									<svg
										classes={classes.circleGraphic}
										viewBox="0 0 24 24"
										xmlns="http://www.w3.org/2000/svg"
									>
										<circle
											cx="12"
											cy="12"
											r="8.75"
											stroke-dasharray="54.978"
											stroke-dashoffset="27.489"
											stroke-width="2.5"
										/>
									</svg>
								</div>
								<div classes={classes.gapPatch}>
									<svg
										classes={classes.circleGraphic}
										viewBox="0 0 24 24"
										xmlns="http://www.w3.org/2000/svg"
									>
										<circle
											cx="12"
											cy="12"
											r="8.75"
											stroke-dasharray="54.978"
											stroke-dashoffset="27.489"
											stroke-width="2"
										/>
									</svg>
								</div>
								<div classes={[classes.circleClipper, classes.circleRight]}>
									<svg
										classes={classes.circleGraphic}
										viewBox="0 0 24 24"
										xmlns="http://www.w3.org/2000/svg"
									>
										<circle
											cx="12"
											cy="12"
											r="8.75"
											stroke-dasharray="54.978"
											stroke-dashoffset="27.489"
											stroke-width="2.5"
										/>
									</svg>
								</div>
							</div>
						</div>
					)}
```

src/theme/dojo/loading-indicator.m.css
```diff
@@ -1,4 +1,7 @@
 .root {
```
agubler, line 1: Should the dojo theme have the size classes with the appropriate CSS Vars?

src/theme/material/loading-indicator.m.css
```diff
@@ -1,4 +1,4 @@
-.root {
+.linear {
```
agubler, line 1: Should the material theme have the size classes with the appropriate CSS Vars?

agubler (changes requested):

src/loading-indicator/index.tsx
samends: Yea I was planning ahead too much, figured with that code we could add more types like `spin-small` without having to update the code. However over engineering for the future is thing

src/theme/dojo/loading-indicator.m.css
samends: I've added the `.small` `.medium` and `.large` classes to the dojo and material theme, however should the sizes differ from the default sizing?

agubler: As long as they use CSS variables and they are configured in the variants for dojo/material default and dark then they can be the same as the default sizes - that will allow users to create a custom variant and change them

src/theme/dojo/loading-indicator.m.css
```diff
@@ -7,6 +10,15 @@
 	width: 100%;
 }
 
+.circular {
+	border: calc(var(--grid-base) / 2) solid var(--color-background-faded);
+	border-radius: 50%;
+	border-top: calc(var(--grid-base) / 2) solid var(--color-background-inverted);
+	width: calc(var(--grid-base) * 6.25);
```
tomdye, line 17: I'm not sure that these sizes here etc are relevant to the grid size, ie. if the grid size changed would you expect the loader to change size?
Also, we should really refrain from fractional multipliers like these. Just hardcode the width / height appropriately  in this case.

samends: Removed the width/height here as they're not needed

tomdye:
@samends Looks like you've addressed Ant's concerns but I believe it would benefit from having the different sizes examples added to the circular indicator example please.

tomdye (approved):
Thanks for the extra examples 👍

PR #1708, opened 2021-03-25
Add vertical and horizontal layout Stacks

### LINKED ISSUE #1585: Add VStack and HStack components
**Enhancement** 

Add VStack and HStack components for laying widgets out horizontally and vertically. This widget has been argued against in the past because "it's 1 line of css in flex", but that ignores the fact that these components would be able to apply padding and margin specific to a theme for the user as well.

agubler:
**Type:** feature

**Description:**

Introduces a set of layout components that can be used to create vertical and horizontal stacks of elements or widgets. A special `Spacer` component can be used to customise spacing between nodes in the stack. As well as the `Spacer` widget a number of properties can be configured to tweak the layout of the nodes in the stack and the container node itself.

* `align`: Used to align the nodes in the stack for the opposite direction of the stack
* `padding`: Sets padding on the main container, small, medium, large
* `spacing`: Space between the stack nodes, small, medium, large
* `stretch`: Stretches the container across width and height

The CSS theme for spacing and padding can be controlled by setting the relevant themes' CSS variables.

Two specific widgets have been created for each of the supported directions, `HStack` and `VStack` but also the general `Stack` can be used with a `direction` property passed (either `vertical` or `horiztonal`)

```tsx
<Stack direction="vertical">
    <TextInput />
    <Spacer />
    <TextInput />
</Stack>

<Stack direction="horizontal">
    <TextInput />
    <Spacer />
    <TextInput />
</Stack>
```

Resolves #1585


src/examples/src/config.tsx
```diff
@@ -316,8 +320,18 @@ export const config = {
 		{ label: 'default', theme: { theme: {} } }
 	],
 	tests,
-	readmePath: (widget: string) => `src/${widget}/README.md`,
-	widgetPath: (widget: string, filename: string) => `src/${widget}/${filename || 'index'}.tsx`,
+	readmePath: (widget: string) => {
+		if (widget === 'vstack' || widget === 'hstack') {
```
tomdye, line 324: is this something we can raise as an issue against parade to allow for different readme paths per item

agubler: Don't need that anymore, removed.

tomdye (approved):
I think that numbers inside the boxes may make the examples easier to correlate to the code, but otherwise it looks great. Very useful, thanks!

matt-gadd (approved):

PR #1712, opened 2021-03-29
support more date formats and custom date parsers

### LINKED ISSUE #1563: DateInput widget not working with browsers set to date formats other than 'dd/mm/yyyy' or 'mm/dd/yyyy'
**Bug** 

The DateInput widget will by default adopt the date format the browser is currently set to. When your date format is "dd.mm.yyyy" like in many EU countries, the DateInput widget will format the date in that format, but it will show the validation notice: "Invalid date format".

What I found:

1) There is no extension point yet to add DateFormats.

2) The widget uses the "Intl" builtin for date formatting (that's why the date shows up in the proper format). But the widget uses its own regex validation parsing in date-utils. Since date-utils only supports ISO and "dd/mm/yyyy" (plus the US version of it), you have a mismatch whenever your browser is set to something else.

3) We didn't get it to work with ISO either - because you can't tell it to just use ISO, you would have to configure the browser to deliver ISO from the Intl api by default. 

For completeness sake: tested in Chrome/Safari on MacOS and Chrome/Firefox on Windows

Package Version: @dojo/widgets 7.0.3

**Code**

Set your browser to a date format like "dd.mm.yyyy" - anything that isn't 'dd/mm/yyyy' or 'mm/dd/yyyy' and open the widget test page: 

https://dojowidgets-git-list-typeahead-filtering-select-behaviour.dojo1.vercel.app/#widget/date-input/overview

**Expected behavior:**

Date is displayed in browser locale date format and validates.

**Actual behavior:**

Date is displayed in browser locale date format but does not validate.

DateInput

aciccarello:
**Type:** bug

**Description:**
Adds support for more date formats and the ability to pass a custom date parser. This resolves an issue with the date handling outside of a few select locales and adds support for several common date formats (as defined by `Intl.DateTimeFormat().format()`). For other locales or for more flexible date parsing (such as to support multiple formats at once), a custom date formatter/parser combo can be passed.

For locales where `Intl.DateTimeFormat().format()` produces something that is not supported, selecting a date from the calendar was resulting in an invalid date warning. This was because the Date value is formatted as a string using `DateTimeFormat` and later parsed. The fix was adding a check to the date-to-string formatter that avoids outputting un-parseable strings and defaults to ISO strings.

Resolves #1563

### DIFF
--- a/src/date-input/date-utils.tsx
+++ b/src/date-input/date-utils.tsx
@@ -14,18 +14,47 @@ const dateExpressions = () => {
 	];
 
 	const usCanary = Intl.DateTimeFormat().format(new Date(2018, 3, 3)); // April 3
-	if (usCanary === '4/3/2018') {
-		// US-only 'mm/dd/yyyy' format
-		tokens.push([
-			new RegExp(`^${components.month}\/${components.day}\/${components.year}$`),
-			{ month: 1, day: 2, year: 3 }
-		]);
-	} else {
-		// standard 'dd/mm/yyyy' format
-		tokens.push([
-			new RegExp(`^${components.day}\/${components.month}\/${components.year}$`),
-			{ month: 2, day: 1, year: 3 }
-		]);
+	switch (usCanary) {
+		case '4/3/2018':
+		case '04/03/2018':
+			// US-only 'mm/dd/yyyy' format
+			tokens.push([
+				new RegExp(`^${components.month}\/${components.day}\/${components.year}$`),
+				{ month: 1, day: 2, year: 3 }
+			]);
+			break;
+		case '3/4/2018':
+		case '03/04/2018':
+			// standard 'dd/mm/yyyy' format
+			tokens.push([
+				new RegExp(`^${components.day}\/${components.month}\/${components.year}$`),
+				{ month: 2, day: 1, year: 3 }
+			]);
+			break;
+		case '3.4.2018':
+		case '03.04.2018':
+			// standard 'dd.mm.yyyy' format
+			tokens.push([
+				new RegExp(`^${components.day}\\.${components.month}\\.${components.year}$`),
+				{ month: 2, day: 1, year: 3 }
+			]);
+			break;
+		case '3-4-2018':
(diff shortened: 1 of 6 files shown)


src/date-input/date-utils.tsx
```diff
@@ -14,18 +14,47 @@ const dateExpressions = () => {
...
+		case '03.04.2018':
+			// standard 'dd.mm.yyyy' format
+			tokens.push([
+				new RegExp(`^${components.day}\\.${components.month}\\.${components.year}$`),
+				{ month: 2, day: 1, year: 3 }
+			]);
+			break;
+		case '3-4-2018':
+		case '03-04-2018':
+			// standard 'dd.mm.yyyy' format
```
mwistrand, line 44: => `// standard 'dd-mm-yyyy' format`

src/date-input/date-utils.tsx
```diff
@@ -67,5 +96,11 @@ export function formatDateISO(date: Date | undefined) {
...
 
 export function formatDate(date: Date) {
-	return Intl.DateTimeFormat().format(date);
+	const formattedDate = Intl.DateTimeFormat().format(date);
+	// Check if is a supported date format
+	const parsedDate = parseDate(formattedDate);
+	if (parsedDate && parsedDate.valueOf() === date.valueOf()) {
+		return formattedDate;
+	}
+	return formatDateISO(date);
```
mwistrand, line 105: Warning: this function uses `String.prototype.padStart`, which is not supported in IE11. I'll raise an issue and this can handled in a separate PR.

mwistrand (approved):
Just a couple comments. I think adding support for a custom formatter and parser goes a long way to improve the component's usefulness, since developers can opt into using `@dojo/i18n` formatters/parsers when needed.

Would it be worth adding an example of using `@dojo/i18n` to the date input's parade examples?

src/date-input/date-utils.tsx
mwistrand: This is now fixed in `master`.

PR #1713, opened 2021-03-31
Feature/action groups - 1574

### LINKED ISSUE #1574: Actions (groups of buttons/icons) inconsistency
**Bug** 

A number of widgets have the concept of "Actions", they are mostly collections of buttons/icons. It would be nice if we had one consistent pattern for these. For example at the moment in SpeedDial there is a specific `Action` component:

```
<SpeedDial>
  <Action
    onClick={() => {
      icache.set('action', 'Mailing');
    }}
  >
    <Icon type="mailIcon" />
  </Action>
</SpeedDial>
```

yet for the `Header` actions are defined like so:

```
<Header>
  {{
    title: 'My App',
    actions: [
      <Link to="#foo">Foo</Link>,
      <Link to="#bar">Bar</Link>,
      <Link to="#baz">Baz</Link>
    ]
  }}
</Header>
```

KaneFreeman:
**Type:** feature

All widgets with "actions" have been switched to the same pattern established by Speed Dial, an internal action widget specialized to that widget. Most do not do anything beyond wrapping `ActionButton`, `Link` and `Icon` but create a basis for expanding upon later on.

**Description:**

Resolves #1574

### DIFF
--- a/src/card/index.tsx
+++ b/src/card/index.tsx
@@ -1,7 +1,29 @@
-import { tsx, create } from '@dojo/framework/core/vdom';
 import { RenderResult } from '@dojo/framework/core/interfaces';
-import * as css from '../theme/default/card.m.css';
+import { create, tsx } from '@dojo/framework/core/vdom';
+import ActionButton, { ActionButtonProperties } from '../action-button/index';
+import Icon, { IconProperties } from '../icon/index';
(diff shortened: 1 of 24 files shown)


agubler:
Closing for now while we determine the best / most consistent approach for actions within the lib

PR #1716, opened 2021-04-01
fix: correctly validate required TimePicker

### LINKED ISSUE #1714: When using the Form the TimePicker is reported as valid even when the form field is required
**Bug**

When using the form widget and configuring a field to be "required" it should the whole form should only be valid once the input has been populated. The `TimePicker` reports as valid and therefore the forms valid state is set to true even though the required form as not been populated.

Codesandbox example: https://codesandbox.io/s/practical-solomon-wdysh?file=/src/main.tsx

bitpshr:
**Type:** bug

**Description:**

This pull request fixes an issue with `TimePicker` that prevented required instances from failing validation.

Resolves #1714

### DIFF
--- a/src/time-picker/index.tsx
+++ b/src/time-picker/index.tsx
@@ -335,7 +335,7 @@ export const TimePicker = factory(function TimePicker({
 				}
 			}
 
-			isValid = validationMessages.length === 0;
+			isValid = isValid === undefined ? validationMessages.length === 0 : isValid;
 			icache.set('isValid', isValid);
 		}
 
@@ -436,7 +436,10 @@ export const TimePicker = factory(function TimePicker({
 									onValidate={(valid, message) => {
 										if (valid !== icache.get('inputValid')) {
 											icache.set('inputValid', valid);
-											icache.set('inputValidMessage', message);
+											if (icache.get('dirty')) {
+												icache.set('inputValidMessage', message);
+											}
+											callOnValue();
 										}
 									}}
 									onKeyDown={(key) => {
--- a/src/time-picker/tests/unit/TimePicker.spec.tsx
+++ b/src/time-picker/tests/unit/TimePicker.spec.tsx
@@ -65,32 +65,33 @@ const baseTemplate = (date?: Date) =>
 		);
 	});
 
-const buttonTemplate = assertionTemplate(() => {
-	return (
-		<div classes={css.input}>
-			<TextInput
-				disabled={undefined}
-				key="input"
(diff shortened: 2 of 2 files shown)


src/time-picker/index.tsx
```diff
@@ -335,7 +335,7 @@ export const TimePicker = factory(function TimePicker({
 				}
 			}
 
-			isValid = validationMessages.length === 0;
+			isValid = isValid === undefined ? validationMessages.length === 0 : isValid;
```
bitpshr, line 338: This was the root of the issue: The underlying `TextInput` updates `inputIsValid`, but then local validation and `validationMessages.length` ended up overwriting this before triggering `onValidate`.

tomdye:
@bitpshr should this be showing as visibly valid, ie. green, when set to required and with a value passed? It appears that it currently does not. Also, it looks like the dojo themed time picker now shows as valid / green as soon as it is rendered even though it's not supposed to.

bitpshr:
@tomdye: The Material `TimePicker` does not show any valid border in its valid state (and neither do any other inputs, like `TextInput`), so I think we're good there. The Dojo `TimePicker` always showed a green border when initially rendered, and that is not related to these changes, but I can take a look at that as part of this PR.

agubler (changes requested):
This works for me, I'd like to get @tomdye's approval and also a test to cover the changes?

bitpshr:
@agubler test added that fails on `master` and passes here, made as simple as possible to show the issue. cc sir @tomdye when you have a chance to sign off.

src/time-picker/tests/unit/TimePicker.spec.tsx
```diff
@@ -65,32 +65,33 @@ const baseTemplate = (date?: Date) =>
...
-				type="text"
-				variant={undefined}
-				classes={undefined}
-			>
-				{{ label: undefined, trailing: undefined }}
-			</TextInput>
-		</div>
-	);
-});
+const buttonTemplate = (required?: boolean) =>
```
tomdye, line 68: test looks good, but we've tried really hard to remove any need for these `getRender` / `getTemplate` functions via the addition of the assertion template so I'd rather we didn't add one here when you could simply to

```
const requiredTemplate = baseTemplate
			.setProperty('@input', 'required', true)
```

tomdye (approved):
Added a comment re. how `setProperty` should be used on the `baseTemplate`, once that's fixed please merge

PR #1719, opened 2021-04-06
feat: wizard content configuration

### LINKED ISSUE #1611: Wizard should actually be a Container too
*Enhancement** 
The current wizard behaviour is not very useful as it does not actually allow users to pass content, it should be more like our other containers. For example it's currently impossible to actually implement the material vertical stepper behaviour:

image

bitpshr:
**Type:** feature

**Description:**

This pull request adds custom content support to `Wizard` step configuration.

preview

Resolves #1611

### DIFF
--- a/src/wizard/index.tsx
+++ b/src/wizard/index.tsx
@@ -1,5 +1,5 @@
 import { create, tsx } from '@dojo/framework/core/vdom';
-import { DNode, RenderResult } from '@dojo/framework/core/interfaces';
+import { RenderResult } from '@dojo/framework/core/interfaces';
 import theme from '../middleware/theme';
 import Icon from '../icon';
 import * as css from '../theme/default/wizard.m.css';
@@ -41,7 +41,7 @@ export type StepStatus = 'pending' | 'inProgress' | 'complete' | 'error';
 
 const factory = create({ theme })
 	.properties<WizardProperties>()
-	.children<WizardChildren | RenderResult | undefined>();
+	.children<WizardChildren | RenderResult | RenderResult[] | undefined>();
 
 export default factory(function Wizard({ properties, children, middleware: { theme } }) {
 	const themedCss = theme.classes(css);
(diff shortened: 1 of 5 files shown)


src/wizard/index.tsx
```diff
@@ -25,16 +26,22 @@ export interface WizardProperties {
...
-	step?(status: StepStatus | undefined, index: number, step: Step): RenderResult;
+	step?: StepRenderer;
 }
 
 export type StepStatus = 'pending' | 'inProgress' | 'complete' | 'error';
 
 const factory = create({ theme })
 	.properties<WizardProperties>()
-	.children<WizardChildren | undefined>();
+	.children<WizardChildren | RenderResult | undefined>();
```
agubler, line 44: I think this needs to be `(RenderResult | RenderResult[])` otherwise going to run into problems with tsx

src/wizard/index.tsx
```diff
@@ -49,7 +56,15 @@ export default factory(function Wizard({ properties, children, middleware: { the
 		theme: themeProp
 	} = properties();
 
-	const [{ step: stepRenderer } = { step: undefined }] = children();
+	let stepRenderer: StepRenderer | undefined;
+	let stepContent: DNode[] | undefined;
+	const body = children();
+
+	if (body && body[0] && !isRenderResult(body[0])) {
```
agubler, line 63: This is quite confusing, i think it can be a little simplier.

```tsx
const [renderer] = children();

if (renderer && typeof renderer.step === 'function') {
    stepRenderer = renderer.step;
}
```

and then because of the condition in stepWrappers you know you can use `children()[index]` as if it was custom renderer it would have already returned.

I think that covers the two scenarios... I a little bit clearer i think.

agubler (changes requested):

src/wizard/index.tsx
bitpshr: `renderer.step` still throws in TS; that's why I was using `isRenderResult`. That's the only real difference here imo, but I will try to make it clearer.

bitpshr: Cleaned up in 47f4e52.

src/wizard/index.tsx
```diff
@@ -49,7 +56,12 @@ export default factory(function Wizard({ properties, children, middleware: { the
 		theme: themeProp
 	} = properties();
 
-	const [{ step: stepRenderer } = { step: undefined }] = children();
+	let stepRenderer: StepRenderer | undefined;
+	const [renderer] = children();
+
+	if (renderer && !isRenderResult(renderer)) {
```
agubler, line 62: Oh that's much cleaner!

agubler (approved):
I think the example should actually be wired to show the wizard working, at the moment this doesn't work when it is uncontrolled and i don't think we should give the impression that it can 😄

PR #1724, opened 2021-04-08
1609 - Do not set first item as active by default

### LINKED ISSUE #1609: Should the list automatically set the first selected item?
**Bug**

At the moment if you create a `List` the first option is always selected if no value is set. For the `Listbox` html element, if no value is set no option is highlighted which seems the more correct behaviour?

KaneFreeman:
**Type:** bug

**Description:** When no active index is set, no list item is selected instead of the previous behavior of defaulting to the first item.

Resolves #1609


src/list/index.tsx
```diff
@@ -726,17 +726,16 @@ export const List = factory(function List({
...
 	const inputText = icache.get('inputText');
 	const {
 		meta: { total = 0 }
 	} = get(options(), { meta: true, read });
 	if (inputText && inputText !== icache.get('previousInputText') && total) {
 		const items = get({ ...options(), offset: 0, size: total });
-		const first = items.slice(0, computedActiveIndex);
-		const second = items.slice(computedActiveIndex);
-		let foundIndex = computedActiveIndex;
+		const first = computedActiveIndex ? items.slice(0, computedActiveIndex) : [];
```
tomdye, line 736: if computed active index was `0` wouldn't this be falsey rather than selecting the first item?

src/list/index.tsx
```diff
@@ -807,7 +805,11 @@ export const List = factory(function List({
 			styles={rootStyles}
 			role={menu ? 'menu' : 'listbox'}
 			aria-orientation="vertical"
-			aria-activedescendant={`${idBase}-item-${computedActiveIndex}`}
+			aria-activedescendant={
+				computedActiveIndex !== undefined
```
tomdye, line 809: 👍 good catch

src/list/index.tsx
```diff
@@ -362,15 +362,21 @@ export const List = factory(function List({
 				if (event.metaKey || event.ctrlKey) {
 					setActiveIndex(total - 1);
 				} else {
-					setActiveIndex((computedActiveIndex + 1) % total);
+					setActiveIndex(
+						(computedActiveIndex !== undefined ? computedActiveIndex + 1 : 0) % total
```
agubler, line 366: Can we do this outside the function call? It's a bit messy to understand at the moment.

src/list/index.tsx
```diff
@@ -362,15 +362,21 @@ export const List = factory(function List({
...
 				}
 				break;
 			case Keys.Up:
 				event.preventDefault();
 				if (event.metaKey || event.ctrlKey) {
 					setActiveIndex(0);
 				} else {
-					setActiveIndex((computedActiveIndex - 1 + total) % total);
+					setActiveIndex(
+						(computedActiveIndex !== undefined
```
agubler, line 376: Same comment re moving the logic out from the function call.

agubler (changes requested):
couple small comments

agubler (approved):

PR #1735, opened 2021-04-09
1565 - Add typeahead dropdown icon/button

### LINKED ISSUE #1565: Usability Typeahead widget: add select button
**Enhancement**

Typeahead widgets lack a visual clue that they are selection fields. When Typeahead is not used as an input helper for free text (e.g. input helper for mail recipients) but used in validation mode as a selection widget, the user wants a visual clue the field can be clicked. I suggest adding a button - exactly like in the Select widget - for this purpose. That might be configurable or an opinionated behavior (validating=show button, free text=no button).

Package Version: 7.0.3

**Code**

n/a

**Expected behavior:**

Show selection button like Select widget when Typeahead is configured as validating (or make it configurable)

**Actual behavior:**

No selection button is displayed, no visual cue.

KaneFreeman:
**Type:** feature

**Description:**
Add typeahead dropdown icon/button

Resolves #1565


src/typeahead/index.tsx
```diff
@@ -418,7 +431,28 @@ export const Typeahead = factory(function Typeahead({
 								}}
 								valid={valid}
 							>
-								{{ label, leading }}
+								{{
+									label,
+									leading,
+									trailing: showDropDownButton && (
+										<button
```
agubler, line 438: We couldn't use the dojo button here?

src/typeahead/index.tsx
```diff
@@ -53,6 +55,8 @@ export interface TypeaheadProperties {
 	itemDisabled?: ListProperties['disabled'];
 	/** Flag to indicate if values other than those in the resource can be entered, defaults to true */
 	strict?: boolean;
+	/** Flag to indicate if drop down arrow should be shown in trailing section of text input, defaults to false */
+	showDropDownButton?: boolean;
```
agubler, line 59: feels like a bit of a mouthful! I don't have a better idea other than we could potentially drop `show`

matt-gadd: I prefer dijit's `hasDownArrow` if that helps.

agubler: @matt-gadd That does help! Thanks, I'm sold.

agubler (changes requested):
A couple of comments

src/typeahead/index.tsx
KaneFreeman: This is following the same pattern that the `Select` widget is using. If we want to change that pattern both should be changed together.

src/theme/dojo/typeahead.m.css
```diff
@@ -5,3 +5,27 @@
...
+	line-height: var(--line-height-base);
+	padding: var(--spacing-regular);
+
+	cursor: pointer;
+	outline: none;
+
+	background: unset;
+
+	appearance: none;
+	-moz-appearance: none;
```
agubler, line 25: do we need these? doesn't the build generate these for us? cc @matt-gadd

agubler (approved):
One question but otherwise :+1:

src/theme/dojo/typeahead.m.css
KaneFreeman: It does not appear to. I just tested removing them and they are not in the end product.

Several other widgets have them in their css too.

PR #1736, opened 2021-04-12
1564 - Allow typeahead selection on tab

### LINKED ISSUE #1564: Usability Typeahead widget: completion
**Enhancement** 

I would prefer if Typeahead completed on tab as well, not just on enter. When you fill out a form (except on touch devices), you enter stuff, then tab to the next field. It does not feel natural that TypeAhead is an exception and requires Enter. As an alternative, you could sort of autocomplete once there is only one possible match in the dropdown. Then, tab will implicitly do what I want, because the field will be completed and I have tabbed to the next field.

Package Version: @dojo/widgets 7.0.1

**Code**
n/a

KaneFreeman:
**Type:** feature

**Description:**
Allow typeahead selection on tab in addition to enter.

Resolves #1564

### DIFF
--- a/src/typeahead/index.tsx
+++ b/src/typeahead/index.tsx
@@ -204,6 +204,7 @@ export const Typeahead = factory(function Typeahead({
 					});
 				}
 				break;
+			case Keys.Tab:
 			case Keys.Enter:
 				preventDefault();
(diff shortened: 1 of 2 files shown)


agubler (approved):

agubler:
Nice and simple one!

PR #1744, opened 2021-04-13
feat: easier list item theming

### LINKED ISSUE #1726: Add CSS Variables to be able to control specific list styles
**Enhancement**

The list currently uses the standard CSS variables that are shared across multiple components, it would be useful if the list had specific variables that use the current variables by default that could be overridden in a variant in order to independently customise styles for behaviours such as selected and hover. For example (using Material, would require complimentary vars in other theme variants):

```css
--mdc-list-item-selected-background: var(--mdc-theme-on-surface-hover);
--mdc-list-item-selected-text: var(--mdc-text-color);
--mdc-list-item-hover-background: var(--mdc-theme-on-surface);
```

bitpshr:
**Type:** feature

**Description:**

This pull request updates the `ListItem` by adding background and color variables to applicable states (disabled, active, selected) so that it can be themed independently of other widgets.

Resolves #1726

### DIFF
--- a/src/examples/src/config.tsx
+++ b/src/examples/src/config.tsx
@@ -110,6 +110,7 @@ import ListItemRenderer from './widgets/list/ListItemRenderer';
 import FetchedResource from './widgets/list/FetchedResource';
 import DisabledList from './widgets/list/Disabled';
 import DraggableList from './widgets/list/Draggable';
+import CustomThemeList from './widgets/list/CustomTheme';
 import Menu from './widgets/list/Menu';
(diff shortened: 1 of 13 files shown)


agubler:
What do you think to adding an example? We should be able to demonstrate this by rendering a list with the `variant="inherit"` property in a wrapping node that has a class that changes the definition of the new variables 😄 

I think you can define them inline over the class, which will be super clear in the example of what is going on.

agubler:
Sorry @bitpshr forgot to mention you in the comment above ^

agubler (approved):

PR #1711, opened 2021-03-26
Add disabled to CheckboxGroup and RadioGroup

samends:
**Type:** feature

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code matches the [style guide](https://github.com/dojo/framework/blob/master/STYLE.md)
* [x] Unit tests are included in the PR
* [x] For new widgets, an entry has been added to the `.dojorc`
* [x] For new widgets, `theme.variant()` is added to the root domnode
* [x] Any widget variant uses `theme.compose` like [this](https://github.com/dojo/widgets/issues/847)
* [x] WidgetProperties are exported

**Description:**
- Adds `disabled` field to `RadioGroup` `options` property
- - Adds `disabled` field to `CheckboxGroup` `options` property

Resolves #1698

PR #1745, opened 2021-04-13
feat: add disabled prop to radio and checkbox groups

### LINKED ISSUE #1698: Checkbox/Radiobox group missing disabled property
**Enhancement**

At present, checkbox and radiobox group widgets do not have a `disabled` property.
We should add a top level `disabled?: boolean` property that can be used to disable all inputs within the group.

bitpshr:
**Type:** bug / feature

**Description:**

This pull request adds a top-level `disabled` prop to both `CheckboxGroup` and `RadioGroup` and updates tests accordingly.

Resolves #1698
Replaces #1711

### DIFF
--- a/src/checkbox-group/index.tsx
+++ b/src/checkbox-group/index.tsx
@@ -18,6 +18,8 @@ export interface CheckboxGroupProperties {
 	initialValue?: string[];
 	/** A controlled value for the checkbox group */
 	value?: string[];
+	/** Disabled all inputs within this group */
+	disabled?: boolean;
 }
 
 export interface CheckboxGroupChildren {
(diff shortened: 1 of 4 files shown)


src/checkbox-group/index.tsx
```diff
@@ -18,6 +18,8 @@ export interface CheckboxGroupProperties {
 	initialValue?: string[];
 	/** A controlled value for the checkbox group */
 	value?: string[];
+	/** Disabled all inputs within this group */
+	disabled?: boolean;
```
agubler, line 22: Do you think we should pass the disabled value down to the custom renderer as well so that the user can decide to do something with that or nah?

agubler (approved):
A small question, but essentially this is good to go (depending on answer to said question!).

src/checkbox-group/index.tsx
bitpshr: Yes! Good catch.

agubler (approved):

PR #1747, opened 2021-04-14
1732 - non-fixed style tabs with scrolling

### LINKED ISSUE #1732: Enhance Tab Container to support non fixed tab widths
**Enhancement**

At the moment the tab container renders the tabs with equal widths fixed using up the full width. This is a "fixed" mode, it would be useful to support a mode that renders the tab buttons that don't stretch across the full width of the tab container.

We could investigate introducing a new property to control this, such as `fixed` when true applies the current behaviour and false would render the tabs contained to their size (with a sensible min and max width).

Material Specs: https://material.io/components/tabs#specs

KaneFreeman:
**Type:** feature

**Description:**
- Add variable width tab style
- Add new `fixed` property to toggle between the two styles
- Make tables scrollable when in variable width mode

Resolves #1732


src/tab-container/index.tsx
```diff
@@ -177,10 +182,46 @@ export const TabContainer = factory(function TabContainer({
...
 
-	const content = [
-		<div key="buttons" classes={themeCss.tabButtons}>
-			{tabs.map(renderTab)}
-		</div>,
+	/**
+	 * Computes the height of browser-rendered horizontal scrollbars using a self-created test element.
+	 * May return 0 (e.g. on OS X browsers under default configuration).
+	 */
+	const computeHorizontalScrollbarHeight = (): number => {
```
agubler, line 189: @KaneFreeman we have a middleware (https://github.com/dojo/widgets/blob/master/src/middleware/offscreen.ts) that can be used for offscreen  rendering in order to calculate dimensions call `offscreen`, also I think we can leverage the `getOrSet` icache api to make it a bit neater. I think something like this will work:

```tsx
icache.getOrSet('horizontalScrollbarHeight', () => {
    return offscreen(
        () => <div classes={[themeCss.scrollTest]}/>,
        (node) => node.offsetHeight - node.clientHeight
    );
});
```

Also I don't think the class `themeTest` should be a themeable class, probably should be a fixed class, which means we'll need to add a new css file for the tab-container.

agubler: Caveat I haven't tried it, but I think it should work.

agubler (approved):

PR #1748, opened 2021-05-03
fix: chip typeahead value placement

### LINKED ISSUE #1742: ChipTypeahead wraps input to new line
**Bug**

The `ChipTypeahead` wraps its input to a new line instead of keeping this inline with the actual chips. This worked prior to the changes introduces as party of #1697.

Package Version: latest

bitpshr:
**Type:** bug

**Description:**

This pull request fixes an issue where chip values were not wrapped inline with the input.

Resolves #1742

### DIFF
--- a/src/text-input/index.tsx
+++ b/src/text-input/index.tsx
@@ -8,7 +8,6 @@ import { formatAriaProperties } from '../common/util';
 import HelperText from '../helper-text/index';
 import Label from '../label/index';
 import * as css from '../theme/default/text-input.m.css';
-import { isArray } from 'util';
 
 export type TextInputType =
 	| 'text'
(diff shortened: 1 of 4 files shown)


src/text-input/index.tsx
```diff
@@ -8,6 +8,7 @@ import { formatAriaProperties } from '../common/util';
 import HelperText from '../helper-text/index';
 import Label from '../label/index';
 import * as css from '../theme/default/text-input.m.css';
+import { isArray } from 'util';
```
matt-gadd, line 11: not sure if this is a correct import

src/text-input/index.tsx
```diff
@@ -226,6 +227,11 @@ export const TextInput = factory(function TextInput({
 	const inputFocused = focus.isFocused('input');
 	const autofilled = Boolean(icache.get('autofilled'));
 
+	let leadingElements;
+	if (leading) {
+		leadingElements = isArray(leading) ? leading : [leading];
```
matt-gadd, line 232:
```
Array.isArray
```

bitpshr: TEMPORARY INSANITY

PR #1755, opened 2021-06-01
Widget: Text Widget

### LINKED ISSUE #1647: Text widget
It would be great to have a Text widget for creating consistent styled text without having to use CSS every time. This will make theming your application even easier.

Example usage,

```tsx
<Text size="h1">Heading</Text>
<Text style="success">Some lovely green text</Text>
```

You should be able to specify the size (i.e., header, subheader, normal body text, label text, etc) as well as a style like "success", "alert", etc.

agubler:
**Type:** feature

**Description:**

`Text` widget used to provide a consistent approach to typography for a theme. The `Text` widget renders a `p` node by default that can be changed using the `as` property, for example:

```tsx
<Text as="h1">Heading Text</Text>
```

The font size is controlled by the `size` property (default to `m`) which in turn is defined by CSS vars in the theme variants. 

For longer text `truncated` can be used to add an ellipsis and finally `inverse` to change the text color to work on the inverse theme background.

Resolves #1647


src/text/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+import theme from '../middleware/theme';
+
+import * as css from '../theme/default/text.m.css';
+
+export interface TextProperties {
+	size?: 'xs' | 's' | 'm' | 'l' | 'xl' | 'xxl';
+	inverse?: boolean;
+	truncated?: boolean;
+	as?: string;
+}
```
tomdye, line 11: Have you considered adding `weight` and `uppercase` ? I've found those useful on recent projects.

src/text/index.tsx
```diff
@@ -0,0 +1,54 @@
+import { v, create } from '@dojo/framework/core/vdom';
+import theme from '../middleware/theme';
+
+import * as css from '../theme/default/text.m.css';
+
+export interface TextProperties {
+	size?: 'xs' | 's' | 'm' | 'l' | 'xl' | 'xxl';
+	inverse?: boolean;
+	truncated?: boolean;
+	as?: string;
```
tomdye, line 10: is it worth typing this to standard text nodes such as `span | div | h1-h6 | p` ?

agubler: I thought about it, but it really could be any node, unless that's something we want to restrict...?

src/text/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+const factory = create({ theme }).properties<TextProperties>();
+
+const Text = factory(function Text({ middleware: { theme }, properties, children }) {
+	const themedCss = theme.classes(css);
+	const { as = 'p', truncated = false, size = 'm', inverse = false } = properties();
```
tomdye, line 17: Where did the choice to default this to `p` come from? I'd have though that a `span` would be best as it's an inline text node then. Would make it useable more easily within all other widgets for consistent text styling, ie. within a `Button` / `Label` / `HelperText` etc.

agubler: I took inspiration from the Text component in Chakra that defaulted to `p`, open for discussion :)

src/text/index.tsx
agubler: I think uppercase would be useful as a boolean that adds a text transform - Can you propose would you are thinking around weight?

tomdye: weight would follow the same as `size` ie. `s | m | l` etc and would apply `font-weight: 400 / 600 / 800` etc depending on the theme.

src/text/index.tsx
tomdye: I guess it's better to leave as is then, but might have been nice for auto-completion

tomdye (approved):

PR #1756, opened 2021-06-04
Feature/outlined input

### LINKED ISSUE #1752: Outlined Inputs
**Enhancement**

Add support for outlined inputs as per the material.io guidelines https://material.io/components/text-fields#input-types.

This should use a property `kind` to determine the type of input and supported by all the different types of input such as email-input, password-input, select etc

KaneFreeman:
**Type:** feature

**Description:**
Add outlined option "kind" to `text-input` and `text-area` widgets

**Text Input**
text-input

**Text Area**
text-area

Resolves #1752


src/text-area/index.tsx
```diff
@@ -196,6 +201,60 @@ export const TextArea = factory(function TextArea({
 	const [labelChild] = children();
 	const label = isRenderResult(labelChild) ? labelChild : labelChild.label;
 
+	let {
+		size: { width: labelWidth }
+	} = dimensions.get('label');
+
+	// When notched the label is transformed in the CSS to 75% size, then we need to add 10px of padding to get the desired look
+	labelWidth = labelWidth * 0.75 + 10;
```
tomdye, line 209: is this the same with dojo theme too?

tomdye (approved):

agubler:
@KaneFreeman Looks solid for these inputs, but we do need to add the same to Select and ChipTypeahead as well as they are inputs too.

The styling for the existing chip-typeahead seems to have been affected by the changes too https://widget-test-do-git-fork-kanefreeman-feature-outlined-inp-f7d95a.vercel.app/#widget/chip-typeahead/overview

src/text-area/index.tsx
KaneFreeman: Sadly no. Is there a way to detect the theme name? I did not see a way. I got around this by using CSS for override the width of the label one the dojo theme with an `important!` tag.

tomdye: Can't you just override the transform in the material theme CSS?

KaneFreeman: Are you suggesting removing the 75% text size transform from the material theme?

tomdye: not necessarily, I'm just saying that you should be doing this adjustment, whatever it needs to be, in the material theme css and not in this widget js.

